PR #505, opened 2020-03-23
test fix: wait on ingest to stop

mikesbrown:
Wait for the ingest progress indicator not to exist as a clue that ingest is complete. This helps reduce test races in cases where we want ingest to succeed.


mikesbrown:
There's a discrepancy in what `npm run format` does on my machine vs. what it expects in CI. 😠

mikesbrown:
Thanks @mason-fish !

PR #506, opened 2020-03-23
reduce itest platforms

alfred-landrum:


alfred-landrum:
Closing; we disabled the windows integration test temporarily in https://github.com/brimsec/brim/pull/508 while we're investigating what's different in the test environment.

PR #507, opened 2020-03-23
Initial changelog, including Windows limitations

philrz:
From a supportability perspective, it's great we have our Windows release, but I want to make sure we've written down that its "beta quality" limitations are known. In a discussion with @alfred-landrum we agreed on an approach like we've done for `zq` where we'll have a `CHANGELOG.md` that we'll update and commit with each release, then copy those same notes into the GitHub Releases tab alongside the artifacts.

Here I've backfilled the notes from our first public Releases into a new `CHANEGLOG.md`, along with my draft proposal for describing the Windows limitations. Once the wording is approved and this PR is merged, I'll backfill those Windows limitations into the notes for those two existing releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This is the same as v0.5.1, but addresses a CI issue that stopped the creation o
 Brim on Windows can currently be considered beta quality. There's two
 significant known limitations:
 
-* On Windows, generation of Zeek logs from pcaps is slower than we'd like it to be. This is due to lack of formal support for Zeek on Windows. We got it working with our own Zeek port, but the best we could do in our first take uses 32-bit [Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
+* On Windows, generation of Zeek logs from pcaps is slower than we'd like it to be. This is due to lack of formal support for Zeek on Windows. We got it working with [our own Zeek port](https://github.com/brimsec/zeek/tree/master/brim/windows), but the best we could do in our first take uses 32-bit [Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
 * We don't yet have the necessary Windows certifications in place to provide a warning-free install experience. Your anti-virus software is likely to jump in with "Are you sure you want to run this?" warnings and scans for viruses. Rest assured that we've not put viruses in our code, but we understand if you're concerned about installing under such conditions. You may want to install on a scratch Windows VM if that feels safer, or you can wait until we've secured the certifications in an upcoming release.
 
 ## v0.5.1
@@ -21,6 +21,6 @@ significant known limitations:
 Brim on Windows can currently be considered beta quality. There's two
 significant known limitations:
 
-* On Windows, generation of Zeek logs from pcaps is slower than we'd like it to be. This is due to lack of formal support for Zeek on Windows. We got it working with our own Zeek port, but the best we could do in our first take uses 32-bit [Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
+* On Windows, generation of Zeek logs from pcaps is slower than we'd like it to be. This is due to lack of formal support for Zeek on Windows. We got it working with [our own Zeek port](https://github.com/brimsec/zeek/tree/master/brim/windows), but the best we could do in our first take uses 32-bit [Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
 * We don't yet have the necessary Windows certifications in place to provide a warning-free install experience. Your anti-virus software is likely to jump in with "Are you sure you want to run this?" warnings and scans for viruses. Rest assured that we've not put viruses in our code, but we understand if you're concerned about installing under such conditions. You may want to install on a scratch Windows VM if that feels safer, or you can wait until we've secured the certifications in an upcoming release.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
...
+* Initial Windows release creation and support. Windows releases are currently unsigned (unlike our Mac releases), but will be soon.
+* Warn on close if there are still active pcap ingests.
+* Fix some issues saving search history.
+
+#### Windows Limitations (beta notice)
+
+Brim on Windows can currently be considered beta quality. There's two
+significant known limitations:
+
+* On Windows, generation of Zeek logs from pcaps is slower than we'd like it to be. This is due to lack of formal support for Zeek on Windows. We got it working with our own Zeek port, but the best we could do in our first take uses 32-bit [Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
```
alfred-landrum: For this sentence, you could add a link to the windows directory in our zeek repo:
https://github.com/brimsec/zeek/tree/master/brim/windows  
in case anyone wants to know how we built zeek.

CHANGELOG.md
```diff
@@ -0,0 +1,26 @@
+## v0.5.2
+
+This is the same as v0.5.1, but addresses a CI issue that stopped the creation of the Windows installer executable. 
+
+#### Windows Limitations (beta notice)
```
alfred-landrum: Did you intend to duplicate this section with the one in v0.5.1 below?

CHANGELOG.md
philrz: @alfred-landrum: Good idea. I've added that link.

CHANGELOG.md
philrz: @alfred-landrum: The duplication was intentional. This speaks to a style question you're certainly welcomed to weigh in on.

These limitations are technically in every Windows release until we address them. So if we look at these as Release Notes, they should be documented alongside every release. Now, due to the level of detail involved, I also respect that extending what I've done here means there's going to be bulky redundancy when looking at the changelog or scrolling down on the GitHub Releases tab.

I'm open to suggestion on ways to improve on that. One thought is that I could break out the detailed text into something like a `README-Windows.md` and just link to that from each block of release note bullets. That would significantly reduce the redundant visual crowding.

alfred-landrum: 👍 I like the idea of breaking the notes out to a separate windows-specific file.

philrz: @alfred-landrum: I've taken a shot at breaking it out into a separate file and linking to it from the `CHANGELOG.md`. Since the one note for v0.5.2 says it's effectively "the same as v0.5.1", I didn't repeat a bullet there about Windows support. But if the approach here is ok with you, in releases going forward, I'd be in favor of including a one-liner bullet:

* Windows releases are currently unsigned (unlike our Mac releases). See README-Windows for details.

It's still a bit of redundancy, but as long as we're coupling a main use case to Wireshark and 70-90% of Wireshark users are on Windows, I'd feel better doing all we can to try and set expectations.

README-Windows.md
```diff
@@ -0,0 +1,18 @@
...
+working with [our own Zeek port](https://github.com/brimsec/zeek/tree/master/brim/windows),
+but the best we could do in our first attempt uses 32-bit
+[Cygwin](https://www.cygwin.com/). We'll be working to improve this over time.
+2. We don't yet have the necessary Windows certifications in place to "sign"
+the release (like we do for macOS) and hence provide a warning-free
+install experience. Your anti-virus software is likely to jump in with "Are
+you sure you want to run this?" warnings and scans for viruses. Rest assured
+that we've not put viruses in our code, but we understand if you're concerned
+about installing under such conditions. You may want to install on a scratch
+Windows VM if that feels safer, or Watch the repo to be notified when we've
```
alfred-landrum: `Watch` -> `watch`

README-Windows.md
```diff
@@ -0,0 +1,18 @@
+# Windows Limitations (beta notice)
+
+Brim support on Windows is currently at beta quality. There's two
+significant known limitations:
+
+1. On Windows, generation of Zeek logs from pcaps is slower than we'd like it
```
alfred-landrum: You could remove the initial `On Windows,` phrase.

README-Windows.md
philrz: Thanks... changed to lowercase. I'm usually a stickler for arbitrary capitalization so perhaps that seemed out of character. It was actually intentional because I was trying to differentiate between "Watch" as a first-class GItHub verb (as in click the "Watch" button) vs. "keep an eye on our stuff and see if it's changed". But I'm happy to change it since that apparently didn't come across or it was still seen as an inappropriate use of caps. I know the proper fix is to link from the `README-Windows.md` to specific GitHub Issues that track those, so I've made a note-to-self to circle back and fix that when those issues exist.

README-Windows.md
philrz: Thanks. Done. Yeah, its previous wording was an artifact of when it was along with the other general release notes. I wanted to really drive home the point that it was a limitation that _only_ affected Windows. Now that it's broken out into a separate file with "Windows" in the name, I'm less concerned.

PR #514, opened 2020-03-26
PROD-1687 Surface download errors in the app

jameskerr:
Instead of reading the response from the server, we were just returning the StatusMessage. 

The pr reads the response and rejects the promise if its an error.

The reason for throwing the error again is so that we don't try to open anything when the promise resolves.

UPDATE:

Wrapped packets.get in a try catch and return a rejected promise if theres an error. This way we always see the error string in the app rather than the console.

The example below is the result of me hard coding a "null" value for the duration field.

image

Added some debug info: (space, destDir, log)

Screen Shot 2020-03-26 at 4 20 18 PM

### DIFF
--- a/src/js/state/Packets/flows.js
+++ b/src/js/state/Packets/flows.js
@@ -18,8 +18,6 @@ export default {
     const state = getState()
     const space = Tab.spaceName(state)
     const destDir = downloadsDir()
-
-    console.log("DEBUG: ", space, destDir, log.getFields())
     return boom.packets
       .get({
         ts_sec: log.getSec("ts"),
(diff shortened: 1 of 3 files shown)


src/js/state/Packets/flows.js
```diff
@@ -38,6 +38,7 @@ export default {
       })
       .catch((error) => {
         dispatch(Packets.error(log.getString("uid"), error))
+        throw error
```
jameskerr: This catch was swallowing the error, which resulted in trying to open an undefined file here: https://github.com/brimsec/brim/blob/surface-download-errors/src/js/flows/downloadPcap.js#L9

alfred-landrum:
I've made a build from this branch so we can try to gather more data about the error:
https://storage.googleapis.com/brimsec/scratch/Brim-Setup-PR514.exe

PR #515, opened 2020-03-27
PROD-1627 add autoUpdater

mason-fish:
The ultimate aim of this PR is to add an autoUpdater to the Brim app such that when a new release is out, the app will automatically update itself. There are a couple of details worth mentioning here which the code might not indicate:

* Using electron's `autoUpdater` along with https://github.com/electron/update.electronjs.org, an update server, which is maintained by GitHub and free for open-source projects.

* Since we are using the above server and framework, there is a dependency for macOS builds to include a `*{darwin}*.zip` asset for each release we would like to auto-update. So I have extended the `release` script to also produce that required zip asset. The autoUpdater will also fail to initialize if the running version is not code signed (i.e. you must run the release script with the `--sign` flag and have our dev ID installed in your mac key chain). 

* I have tested this by building a signed release zip where I adjusted the `package.json` to indicate that the version is `v0.5.2`, manually dropped that into the release assets for that version of release, then built another signed release where I had adjusted the `package.json` to `v0.5.1` and installed that .dmg locally. Here is a video of the result:

autoupdater

**NOTE** I discovered during this process that when we issue releases, we are not properly bumping the version in package.json (the `npm version` command can facilitate this for us) and we are now behind by a couple patches. We should fix this and also include pushing the included .zip as an additional asset for each release.

### DIFF
--- a/scripts/release/install.js
+++ b/scripts/release/install.js
@@ -3,9 +3,11 @@ const fs = require("fs")
 const os = require("os")
 const installer = require("electron-winstaller")
 const createDMG = require("electron-installer-dmg")
+const createZip = require("electron-installer-zip")
 const path = require("path")
 
 const out = "./dist/installers"
+const appPath = "dist/packages/Brim-darwin-x64/Brim.app"
 
 module.exports = {
   darwin: function() {
@@ -13,7 +15,7 @@ module.exports = {
     createDMG(
       {
         overwrite: true,
-        appPath: "dist/packages/Brim-darwin-x64/Brim.app",
+        appPath,
         name: "Brim",
         out
       },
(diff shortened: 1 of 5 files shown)


scripts/release/install.js
```diff
@@ -25,6 +27,18 @@ module.exports = {
         }
       }
     )
+    createZip(
```
alfred-landrum: Would you please chain this with createDMG, and make sure that calling the release script returns an exit error if either of them fail?

src/js/electron/main.js
```diff
@@ -9,13 +9,51 @@ import zqdMainHandler from "./ipc/zqd/mainHandler"
...
+  })
+
+  autoUpdater.on("error", (err) => {
+    console.error("There was a problem updating the application: " + err)
+  })
+
+  // check for updates immediately on startup
+  autoUpdater.checkForUpdates()
+
+  // then check for updates every 10 minutes
```
alfred-landrum: 10 minutes seems a little aggressive given our current release frequency :) . How about once a day?

alfred-landrum:
> Another question, is this expected to only work on a mac right now? Or is expected to work on windows too?

It can't work on Windows until we start signing the Windows releases.

mason-fish:
@jameskerr and @alfred-landrum theoretically, this autoUpdater **does** support Windows according to their docs [here](https://www.electronjs.org/docs/tutorial/updates#using-updateelectronjsorg) but as Al mentioned, since we cannot yet signing the code I didn't test that out yet.

PR #522, opened 2020-03-30
PROD-1691 Add ability to quit app from Menu bar

jameskerr:
Added the following menu items:

* File => Close Tab
* File => Close Window
* File => Exit (windows only)

Screen Shot 2020-03-30 at 4 19 43 PM
Screen Shot 2020-03-30 at 4 20 14 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "itest": "jest --config=itest/config.json --runInBand --verbose",
     "electron": "electron .",
     "build": "npm-run-all -p build:*",
-    "watch": "npm-run-all -p watch:*",
+    "watch": "npm-run-all -s clean build:* watch:*",
     "build:js": "babel src/js -d dist/js --verbose",
     "build:css": "node-sass-chokidar src/css --output dist/css",
     "build:static": "cpx \"src/static/**\" dist/static",
--- a/src/js/electron/menu/appMenu.js
+++ b/src/js/electron/menu/appMenu.js
@@ -37,6 +37,11 @@ export default function appMenu(
     click: () => manager.closeWindow()
   }
 
+  const closeTab = {
+    label: "Close Tab",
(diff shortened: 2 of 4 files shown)


jameskerr:
This is based off: https://github.com/brimsec/brim/pull/521 so that will need to go in first.

package.json
```diff
@@ -15,7 +15,7 @@
     "itest": "jest --config=itest/config.json --runInBand --verbose",
     "electron": "electron .",
     "build": "npm-run-all -p build:*",
-    "watch": "npm-run-all -p watch:*",
+    "watch": "npm-run-all -s clean build:* watch:*",
```
mason-fish: looks like the `watch:*` scripts also run corresponding builds and there is some duplication here

src/js/electron/menu/appMenu.js
```diff
@@ -12,20 +12,36 @@ export default function appMenu(
   platform: string = process.platform
 ) {
   const mac = platform === "darwin"
+  const __ = {type: "separator"}
```
mason-fish: nice variable name :)

package.json
jameskerr: Yes, I can see the reason for curiosity. The watch scripts are the same as the build scripts but just pass an extra flag to them. They also "skip the initial" build, so that only watch for changes.

```
    "watch:js": "npm run build:js -- --watch --skip-initial-build",
    "watch:css": "npm run build:css -- --watch --skip-initial",
    "watch:static": "npm run build:static -- --watch",
```

I basically, added that initial build back in in the diff here. This way we can debug the main process quickly by opening two terminals and running:

**Terminal 1**
`npm run watch`

**Terminal 2**
`npm run electron` we can start and stop this one as much as we want and not have to wait for that initial build.

PR #528, opened 2020-04-01
PROD-1694 Store pcap slices in temp dir

jameskerr:
Instead of the downloads directory, use `app.getPath("temp)` which gets a temp directory in a platform specific way.

I've verified this works and keeps the downloads directory clean. It is also possible to "Save As" in wireshark if a particular slice is interesting.


itest/lib/app.js
```diff
@@ -276,8 +275,6 @@ export const waitUntilDownloadFinished = async (app: Application) =>
     )
   })
 
-export const pcapsDir = () => downloadsDir()
```
mikesbrown, line 279: Yeah this is fine we'll come up with another way to handle this.

jameskerr:
Since the failing CI step relates to the node_modules cache service being unavailable, I'm gonna merge this.

PR #540, opened 2020-04-01
Use zqd waterfall logger

mattnibs:
Change logging config to use the waterfall logger added to zqd in
brimsec/zq#492. Logs named http.access will be logged in zqd-access.log,
every other log will go to zqd-core.log.


mattnibs:
For reviews- while this may pass tests, this will not work until the zq binary is updated.

mattnibs:
Ok, so this doesn't cause tests to fail, however it will cause zqd logging at it currently works to break. This pr should not be merged until brimsec/zq#492 is merged, a release is cut with it. This branch should include the reference to the new zq release.

PR #550, opened 2020-04-03
PROD-1538 use arrow keys to select logs

mason-fish:
Can use arrow keys to select logs from viewer if one has already been selected (and is currently in the search results).

**Note**: While playing around with this I discovered that the existing arrow behavior for scrolling has a bug where it will begin to scroll but then breaks, so this feature also acts as a bit of a bandaid for that since will now preventDafault() on either event. Ideally, IMO, we should track if the currently selected log is not only among the current search results, but is also within the current results view, and if it is not then we will adjust the scroll view to be in sight of it. Then if no result is selected then the up/down arrows will scroll the viewer properly (this is the part that is broken atm), though it looks like there is also some existing functionality in the search input box where that will go through search history (similar to the behavior of some terminals).

updownselect

### DIFF
--- a/src/js/components/useSearchShortcuts.js
+++ b/src/js/components/useSearchShortcuts.js
@@ -5,10 +5,13 @@ import {useEffect} from "react"
 import Mousetrap from "mousetrap"
 
 import Tabs from "../state/Tabs"
+import {adjustSelectedLogIndex} from "../flows/adjustSelectedLogIndex"
 
 export default function() {
   let dispatch = useDispatch()
   useEffect(() => {
+    Mousetrap.bind("down", () => dispatch(adjustSelectedLogIndex(1)))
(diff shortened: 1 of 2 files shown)


jameskerr:
This looks better than what we have now, so I see no reason to hold it back.

It looks like there is still a use case we need to still handle:

* Moving with the down arrow past the last visible log should scroll the page down to keep the selected log visible. Same with going up.

As a side note, I was hoping to soon refactor the viewer to use this libary: https://react-window.now.sh/#/examples/grid/variable-size. I think it will improve the amount of time it takes to mount the tab page.

mason-fish:
Sounds good @jameskerr, I put both of those ideas in https://github.com/brimsec/brim/issues/559 so we can track the work

PR #551, opened 2020-04-04
Prep for v0.6.0

philrz:


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## v0.6.0
+
+* Add auto-updater (#515)
```
alfred-landrum, line 3: This is only added in for MacOS at the moment, as the auto-updating required signed Windows images, and we weren't sure if we would have them ready or not.

philrz: Good catch. Fixed in my local copy and will push a change once I get your confirmation that all feedback is in.

PR #562, opened 2020-04-06
PROD-1606 PROD-1605 Wireshark button in toolbar

jameskerr:
This is an exciting update. In this pr, we've improved the look and feel of the toolbar that sits above the search input.

Screen Shot 2020-04-06 at 3 15 01 PM

Icons and labels have been added to each button to make it more clear what they do.

The Packets button is disabled if the selected log does not have a corresponding conn record, and enabled if it does. Tooltips have been added to make that clear.

image

image

In the log detail pane, the word "PCAPs" has been changed to a button with the shark icon. It is enabled and disabled based on the same criteria as the toolbar button. 

image

**Why have two buttons?**

I initially thought the button on the toolbar would be enough, but it felt awkward to click the Packets button on the toolbar after playing in the log detail pane. Also, users already expect the button to be at the top right of the detail pane. Since this is the core of our product, I'd like to not mess with what's already working. I think it logically makes sense for the toolbar button to go with the table, while the log detail button goes with log detail pane (even though they are exactly the same).

**Known edge cases**

If the "selected log" is from a previous search, the Packets button will be enabled, but no row in the table will be highlighted. It might not be intuitive for them to click the Packets button and have it pull up the packets from a connection in a previous search. 

The right click menu to "Download PCAPs" still only works on the conn logs. It is possible to make it behave like the buttons, but the code there is more complex and I'd like to handle that in another PR.


siskojr:
Looks very cool, nice work.  Two buttons is better for cases like working with two monitors where Log details pane is in another monitor, no need to move mouse back.

philrz:
My one piece of feedback is that the word "Space" feels a little odd where it's positioned here. Because it's sandwiched between the Space pull-down and the search box, it feels like it might not be as obvious to a new user which thing it's trying to label.

image

Some possibilities:

1. Would it be appropriate to have a horizontal line or subtle color change that visually separates the "menu section" from the "querying-and-working-with-events section"?

2. If not that, putting the word "Space" to the left of the pull-down would make it more clearer to me, but I don't know if others would be upset to not have all the words-as-labels oriented the same relative to what they're labeling. 😄

jameskerr:
Thanks @philrz, I see where you are coming from. I'll see if I can fix the ambiguity with better spacing. I'd prefer not to change the color because the space and the search bar are related. And having the label consistent with the other buttons is good. I wouldn't want to change that.

I could maybe put the space picker on the same line as the search input. Like amazon does here.

image

src/js/components/PacketsButton.js
```diff
@@ -0,0 +1,56 @@
...
+      data-tip={getTip()}
+      data-place="bottom"
+      data-effect="solid"
+      data-delay-show={300}
+    >
+      <ToolbarButton
+        icon={<Sharkfin />}
+        disabled={!enabled}
+        onClick={onClick}
+        {...reactElementProps("pcapsButton")}
```
mikesbrown, line 49: I was going to ask you to do this but you already did it. Thank you! 🥇

jameskerr: You got it!

src/js/components/ColumnChooser.js
```diff
@@ -3,22 +3,34 @@
...
-      <div className="column-chooser-wrapper">
+      <div className="columns-button">
         <DropMenu
           position="right-wall"
           value="columns"
           menu={XColumnChooserMenu}
           dim={true}
         >
-          <MenuBarButton>Columns</MenuBarButton>
+          <div
```
mason-fish, line 23: might be nice to refactor the BrimTooltip wrapper so that it can receive the `data-*` attributes, pass them into and return a div like this one, and then that div would wrap around the actual `<ReactTooltip {passing in the non-data-* related props} />` and `{this.props.children}` so that the new interface would be:
```
<BrimTooltip data-tip="helpful message">
            <ToolbarButton icon={<ColumnsIcon />} />
            <label>Columns</label>
</BrimTooltip>
```
wdyt?

src/js/components/hooks/useDebouncedEffect.js
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+import {useEffect, useRef} from "react"
+
+export default function useDebouncedEffect(
```
mason-fish, line 4: Oh awesome! I was just thinking about adding a debouncer for some of the keyboard events we are handling (up and down arrows). 👍

src/js/state/LogDetails/types.js
```diff
@@ -1,23 +1,34 @@
...
+import type {SearchStatus} from "../../types/searches"
 
 export type LogDetailsState = {
-  logs: Log[],
+  entries: LogDetails[],
   position: number,
   prevPosition: number
 }
 
+export type LogDetails = {
```
mason-fish, line 12: 👍

src/js/components/ColumnChooser.js
jameskerr: I see the appeal. But at the moment, I do think it is convenient to keep the target (the element with the mouseover events) and the tooltip component as two distinct pieces. But maybe more will be revealed and it will be more convenient to do as you suggest here.

src/js/components/hooks/useDebouncedEffect.js
jameskerr: Right on! Lodash has a debounce and throttle utility as well, in case you just need a function debounced.

jameskerr:
Thanks for the review of a long PR @mason-fish 🙏

PR #579, opened 2020-04-08
Initial troubleshooting & issue templates

philrz:
I've gone back through the initial set of user-reported issues after the Brim v0 launch and created this "troubleshooting guide" that may have helped the users understand their problems on their own or get the necessary debug info to us with fewer round-trips. We can keep adding to this over time.

I've also adjusted some of the other "open an issue" references in other docs to point to the relevant spot in the troubleshooting doc. I figure savvy/hurried GitHub users are likely to head straight to the Issues tab regardless, but for those that are taking their time, I'd like to guide them toward self-service first.

As part of this, I've also drafted some [issue templates](https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository) to further nudge users toward providing useful info up front. There's a bit of a chicken-&-egg situation with bringing these to life: We'd need to enable them for the repo to see what they look like in practice, but we'd want the templates to be the way we want before turning them on.

To help give you the idea, I've created a personal scratch repo where I've turned this feature on with GitHub's default issue templates. Click the link below to get the feel for what a user would now see when clicking "New Issue", then click a "Get Started" link to see examples of templates in use:

https://github.com/philrz/scratch/issues/new/choose

As you can see, the draft templates in _this_ PR are much shorter than those default ones. In my experience observing GitHub Issues for other projects, it seems like the verbosity in those default templates often backfires. For instance, do we really think saying "clear and concise" over & over again is going to make a user stop and think "Oh, shoot! Here I was going to be vague and rambling. Glad you mentioned this!" 😄

More often than not I see users just skip most/all of the template and write what they were going to write anyway, at which point all the user-opened issues just have tons of noise in them that needs to be deleted or ignored to get to the heart of the issue. When discussing this with @alfred-landrum, he seemed to agree that it would be ideal to keep it tight, e.g. not much more than asking for platform and version. That's what I've tried to do here, though I'm open to further tweaks. I figure it's easier to start tight and if there's important detail that people keep leaving out, we can add it to the templates.


TROUBLESHOOTING.md
```diff
@@ -0,0 +1,161 @@
...
+|[zeek/864](https://github.com/zeek/zeek/issues/864)|Multi-section pcapng with overlapping timestamps creates excess `conn` events|
+
+If you find yourself running into these issues or others of a similar nature,
+please reach out to us on our [public Slack](https://join.slack.com/t/brimsec/shared_invite/zt-cy34xoxg-hZiTKUT~1KdGjlaBIuUUdg)
+or email us at support@brimsecurity.com so we can chat further.
+
+# Gathering Info
+
+Before [opening an issue](#opening-an-issue), or during debug on a
+previously-opened issue, there's detail you can gather that may help us more
```
alfred-landrum, line 87:
```suggestion
Before [opening an issue](#opening-an-issue), or while debugging a
previously-opened issue, there's detail you can gather that may help us more
```

PR #581, opened 2020-04-08
Detect file type

### LINKED ISSUE #573: Detect File Type
We need a function to detect if a file is a pcap or a zeek log. I'm not sure how we're going to do this. I was thinking we could check the encoding and see if its binary or utf-8.

jameskerr:
Fixes #573 

This function reads the first 4 bytes to see if it matches the pcap signatures, then checks for zeek headers on the first four lines with a regex, then checks for json by reading the first 4 lines and seeing if they are json parsable.


src/js/lib/fileType.test.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+
+import path from "path"
+
+import lib from "./"
+
+const testFile = (name) => path.join(__dirname, "../../../itest/", name)
+const pcap = testFile("sample.pcap")
```
mikesbrown, line 8: @jameskerr Can I add a pcapng to this for you? I had planned to anyway and this is a good place to start testing it.

src/js/lib/fileType.js
```diff
@@ -0,0 +1,83 @@
...
+  let bytes = await firstBytes(file, 4)
+  for (let hex of PCAP_HEXES) {
+    if (bytes instanceof Buffer && bytes.equals(Buffer.from(hex, "hex")))
+      return true
+  }
+  return false
+}
+
+function isZeekHeader(line) {
+  return /^#\w+\s+\S+/.test(line)
```
mikesbrown, line 47: This will match a trivially-contrived sourcable sh file, like
```sh
#source me to do stuff
export A=b
```
Instead of `^#\w+\s+\S+`, check for `^#separator \S+` ?

src/js/lib/fileType.test.js
```diff
@@ -0,0 +1,22 @@
...
+const zeek = testFile("sample.zeek.log")
+const json = testFile("sample.json.zeek.log")
```
mikesbrown, line 10: Could you please make these come from sample.pcap already in the repo? You can do this by adding `itest/sample.pcap` to brim and then running the following on its bzng:
```sh
zq -f zeek itest/sample.pcap.brim/all.bzng > itest/sample.tsv
zq -f ndjson itest/sample.pcap/brim/all.bzng > itest/sample.ndjson
```
This makes it easier for testing and the resulting files will be smaller.

src/js/lib/fileType.test.js
jameskerr: Yeah, that's fine with me!

src/js/lib/fileType.js
jameskerr: I see what you mean. How does the proposed regex prevent that problem?

I'm also checking the first 4 lines of the file to see if they match that regex. Is it true that all zeek filees start with the "separator" header?

Lastly, is it a security concern if they spoof this with a non-zeek file? All we do with it pass along the file path to zqd. Whatever they do in zqd will probably throw an error. I don't think anything is able to be "sourced".

Let me ask a zqd dev if this needs to be more robust. @henridf will this zeek detector regex be good enough for ingest?

src/js/lib/fileType.test.js
jameskerr: Yes, I will do that.

src/js/lib/fileType.js
mikesbrown: > Is it true that all zeek filees start with the "separator" header?

I think they must since all lines after have to use the separator as defined.

> Lastly, is it a security concern if they spoof this with a non-zeek file? 

Not necessarily. My $0.02 was, if we are doing file detection we can at least block obviously invalid text files before they go all the way through to zqd.

jameskerr: Aww I see what you mean. If someone accidentally gives us a file with standard `#` comments at the top of the file (which many files do). Got it. Yeah I'll update that.

src/js/lib/fileType.test.js
mikesbrown: Added via:
```sh
mergecap -F pcapng -w sample.pcapng sample.pcap
```

src/js/lib/fileType.js
mikesbrown: Thanks. Still happy to hear anyone else's comments on this issue.

PR #587, opened 2020-04-09
Check the version of the state object

### LINKED ISSUE #584: master starts to error screen with existing space
If I try to run Brim off of current master (16e606a7) when I've previously created a space with the Brim 0.6.0 release, the application loads directly to an error screen, with this error:

```
TypeError: Invalid attempt to spread non-iterable instance
    at _nonIterableSpread (/Users/alfred/work/brim/dist/js/state/LogDetails/reducer.js:15:39)
    at _toConsumableArray (/Users/alfred/work/brim/dist/js/state/LogDetails/reducer.js:13:95)
    at toHistory (/Users/alfred/work/brim/dist/js/state/LogDetails/reducer.js:83:43)
    at /Users/alfred/work/brim/dist/js/state/LogDetails/selectors.js:32:33
    at /Users/alfred/work/brim/node_modules/reselect/lib/index.js:76:25
    at /Users/alfred/work/brim/node_modules/reselect/lib/index.js:36:25
    at /Users/alfred/work/brim/node_modules/reselect/lib/index.js:90:33
    at Object.getHistory (/Users/alfred/work/brim/node_modules/reselect/lib/index.js:36:25)
    at Function.stateToProps [as mapToProps] (/Users/alfred/work/brim/dist/js/components/RightPane.js:172:40)
    at mapToPropsProxy (/Users/alfred/work/brim/node_modules/react-redux/lib/connect/wrapMapToProps.js:53:92)
```

repro on windows & mac:
* run 0.6.0 release, ingest a pcap, exit brim
* run master (currently at 16e606a7)

jameskerr:
Drop the previously saved state if the version is different
This is a crude way to handle state migrations. I'd like to
think this through more, but at least there are no startup
errors when you upgrade versions now.

Going forward, we'll need to increment this VERSION string each time we mess with the shape of the state. 

I think it would be good to write a test that does some of the common operations on the redux store, the saves the resulting state in a snapshot. Then each time we make changes to the state shape, hopefully that test will fail with a message like "update the state version".

Depending on how critical it is for users to save state, like their history and in the future, their saved queries. We may want to write more code to migrate data and what not.

fixes #584

### DIFF
--- a/src/js/initializers/initPersistance.js
+++ b/src/js/initializers/initPersistance.js
@@ -5,7 +5,12 @@ import throttle from "lodash/throttle"
 
 import type {State} from "../state/types"
 
-export const VERSION = "5"
+export const VERSION = "6"
+
+export function isCurrentVersion(state: *) {
+  return state && state.version === VERSION
+}
+
 const KEY = "BRIM_STATE"
 const PERSIST = [
   "tabs",
--- a/src/js/state/createGlobalStore.js
+++ b/src/js/state/createGlobalStore.js
@@ -1,9 +1,12 @@
 /* @flow */
 import {createStore} from "redux"
 
+import {isCurrentVersion} from "../initializers/initPersistance"
 import globalReducer, {type GlobalState} from "./globalReducer"
 
-export default function(initialState: GlobalState | void) {
+export default function(prevState: GlobalState | void) {
+  let initState = isCurrentVersion(prevState) ? prevState : undefined
+
   // $FlowFixMe
-  return createStore(globalReducer, initialState)
+  return createStore(globalReducer, initState)
 }
--- a/src/js/initializers/index.js
+++ b/src/js/initializers/index.js
@@ -2,6 +2,7 @@
 
 import {ipcRenderer} from "electron"
 
+import {isCurrentVersion} from "./initPersistance"
 import closeWindow from "../flows/closeWindow"
 import initBoom from "./initBoom"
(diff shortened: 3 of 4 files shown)


src/js/initializers/initPersistance.js
```diff
@@ -5,7 +5,12 @@ import throttle from "lodash/throttle"
 
 import type {State} from "../state/types"
 
-export const VERSION = "5"
+export const VERSION = "6"
```
jameskerr, line 8: This version string is just an arbitrary number, not tied to our app versions in any way.

src/js/state/createGlobalStore.js
```diff
@@ -1,9 +1,12 @@
 /* @flow */
 import {createStore} from "redux"
 
+import {isCurrentVersion} from "../initializers/initPersistance"
 import globalReducer, {type GlobalState} from "./globalReducer"
 
-export default function(initialState: GlobalState | void) {
+export default function(prevState: GlobalState | void) {
+  let initState = isCurrentVersion(prevState) ? prevState : undefined
```
jameskerr, line 8: I figure it was best to let each store creation be responsible for dealing with its version.

src/js/initializers/initPersistance.js
jameskerr: Both the global reducer and the window reducer import this same constant.

jameskerr:
@mason-fish Yes, that is the plan for the time being.

PR #590, opened 2020-04-09
ci: improve npm caching (fixes #589)

### LINKED ISSUE #589: npm cache problems in ci
Ref: https://github.com/brimsec/brim/actions/runs/74573741

1. Windows cache isn't being found
1. MacOS cache is ~1154 MB. Can this be shrunk? See also https://github.com/actions/cache/issues/187 . For comparison the Linux and Windows caches are about 78M.

mikesbrown:
- Use `npm config get cache` to derive the npm cache directory and use it in the cache step. This fixes the cache misses on Windows.
- Clear the 1G of cache that is on the base macOS runner. Note that `npm cache clear --force` was only able to clear about 250M, so the more direct `rm -rf` approach was needed. This reduces the macOS cache  to about 78M, inline with Windows and Linux cache sizes.
- Change the cache name to prevent reuse of bad caches.

`npm install --no-audit` is still slower on macOS than other platforms, but the massive cache store/recall on mac goes away. It's difficult to compare cloud CI build times, but rough timing comparisons suggest this will generally speed up CI from ~30 to ~10 minutes during PDT mornings.

master: https://github.com/brimsec/brim/actions/runs/74573741
branch: https://github.com/brimsec/brim/actions/runs/74665070

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,9 @@ jobs:
       id: set-npm-cache-dir
       run: echo "::set-output name=npm_cache_dir::$(npm config get cache)"
     - name: Clear Extraneous Runner Cache
+      # The macos Github runner has 1G of NPM cache that we don't need.
+      # Clear it before we create our own cache to prevent slower build
+      # times. See https://github.com/brimsec/brim/pull/590
       run: |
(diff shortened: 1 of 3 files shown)


mikesbrown:
> Do we need add a comment referencing it so that we can remove some of this when that issue is resolved?

done

mikesbrown:
Didn't notice the approval before. Thanks!

PR #602, opened 2020-04-13
Spaces Reducer Bug

### LINKED ISSUE #599: Spaces reducer runs when it should not
We need to add a check and bail if the action is not related to spaces.

This is why we have this undefined key here.

jameskerr:
In redux, every dispatched action is sent to every reducer. There was a bug where the spaces reducer was performing a state update in response to any action passed to to. It now checks for the action type to make sure it starts with "SPACES_". If not it just returns the current state, as expected.

This fixes that error where we would see getState().spaces => {undefined: {}}

Fixes #599


src/js/state/Spaces/test.js
```diff
@@ -79,3 +79,8 @@ test("getting the spaces with details, others not", () => {
     {name: "space-b"}
   ])
 })
+
+test("only cares about spaces actions", () => {
+  store.dispatch({type: "NON_SPACE"})
+  expect(Spaces.raw(store.getState())).toEqual({})
```
jameskerr, line 85: Before the change above, this would fail because state was `{undefined: {}}`

src/js/state/View/reducer.js
```diff
@@ -8,20 +8,14 @@ const init: ViewState = {
   leftSidebarWidth: 350,
   resultsTab: null,
   timeZone: "UTC",
-  investigationView: "linear",
-  isIngesting: false
```
jameskerr, line 12: I have made changes to the state format here. I removed this `isIngesting` boolean that was not being read from anywhere in the application. I think we are safe here and do not need to migrate or increment the version number on the state. But I want reviewers to double check me here.

mattnibs:
> Fixes #602
s/#602/#599

src/js/state/View/reducer.js
mattnibs: I did a check w/ master vs. this branch and didn't run into any problems.

PR #604, opened 2020-04-13
Do not mutate the window state object

### LINKED ISSUE #600: We mutate state in onBeforeUnload
This results in an a bug where windows are unable to be closed. The problem is in flows/saveWindowState.js. We need to make a copy before deleting.

jameskerr:
Fixes #600 

This caused a strange bug where `state.handlers` was undefined. Then the app tried to do something with it, and it threw an error. I traced it back to this state mutation.

It's now locked in with a unit test

### DIFF
--- a/src/js/flows/saveWindowState.test.js
+++ b/src/js/flows/saveWindowState.test.js
@@ -1,0 +1,12 @@
+/* @flow */
+import initTestStore from "../test/initTestStore"
+import saveWindowState from "./saveWindowState"
+
+test("does not mutate state", () => {
+  let store = initTestStore()
+  let prevState = {...store.getState()}
+
+  store.dispatch(saveWindowState())
+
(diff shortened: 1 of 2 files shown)


src/js/flows/saveWindowState.test.js
```diff
@@ -0,0 +1,12 @@
...
+import initTestStore from "../test/initTestStore"
+import saveWindowState from "./saveWindowState"
+
+test("does not mutate state", () => {
+  let store = initTestStore()
+  let prevState = {...store.getState()}
+
+  store.dispatch(saveWindowState())
+
+  expect(store.getState()).toEqual(prevState)
```
jameskerr, line 11: Before the fix above, this would fail with the "errors" "notice" and "handlers" keys missing.

PR #624, opened 2020-04-16
test fix: increase default retryUntil to 60s

mikesbrown:
The tests still have a 5m timeout, which so far is enough padding.


mikesbrown:
Failure is known, not caused by this change.

PR #625, opened 2020-04-16
Load Files Input UI

### LINKED ISSUE #572: Update design of open file area
Add the zeek Icon
Add the zeek to the accepted formats

jameskerr:
fixes #572 

It now looks like so:

Screen Shot 2020-04-16 at 11 52 14 AM

With an animation that looks like this.

P3SfaZvFbQ

This PR is the html and style changes, but the PR #594 contains the functionality. That one must be merged first.


src/js/test/integration.js
```diff
@@ -109,9 +110,12 @@ export const selectors = {
     menu: dataAttrSelector("optionsMenu"),
     menuItem: genSelectorForTextUnderElement("optionsMenu")
   },
+  ingest: {
+    filesInput: dataAttrSelector("ingestFilesInput"),
+    filesButton: dataAttrSelector("ingestFilesButton")
+  },
```
mikesbrown, line 116: Thanks for being mindful of good naming!

bin/generator/icon.js
```diff
@@ -0,0 +1,27 @@
+/* @flow */
```
mason-fish, line 1: Oh fun! I was *just* thinking about how useful a generator like this would be as I added in the `expand` icon recently 👏

jameskerr:
@mikesbrown I'd be happy to write an integration test. Thanks for the suggestion.

PR #629, opened 2020-04-16
fix windows ci: dont rely on focus to destroy a window

### LINKED ISSUE #598: on-close handler prevents Spectron's app.stop() from functioning on Windows
@mikesbrown explains:

---
PR #499 prevents Spectron's `app.stop()` from working in Windows. This is a regression in the sense that reliable methods of driving the app in automated testing are broken for the windows platform. Being able to use `app.stop()` is important because it clears out chromedriver and electron and removes that responsibility from a test maintainer to do that themselves.

You can reproduce the problem by doing the following:
1. Getting a windows machine and installing Node 12 and GIt
2. Cloning Brim
3. Running:

```
$ npm install
$ npm run build
$ npm run itest -- -t tab itest/test/smoke
```

You may have to run the last command (that includes itest) a subsequent time to produce the problem.

Before #499 at hash `e0b01933e8bb564eca680f3d245870cdc28be3ac`, you could observe the test pass and the app stop at the end of the test (including all windows disappearing). At the next hash  `990eb2f0160d65748eeee1f6abcc8cc16e9edbc0` you can observe the app still up after the test passes.

Not only does Brim fail to exit using Spectron's `stop()`, but also Brim fails to close using the browser-window API `close()`. Tests using spectron can issue something like:

```
app.browserWindow.close()
```

Brim responds to this call on Mac; in Windows the window remains open. On both platforms, messages like the following are logged:

```
'[ec4b8f1316]:  Confirming Unload...',
      '[ec4b8f1316]:  Unloading',
      '[ec4b8f1316]:  Saving window state'
```

I can the problem go away by applying this patch:

```
diff --git a/src/js/electron/tron/window.js b/src/js/electron/tron/window.js
index 3dea96e7..732043cd 100644
--- a/src/js/electron/tron/window.js
+++ b/src/js/electron/tron/window.js
@@ -32,10 +32,6 @@ function searchWindow(params) {
       nodeIntegration: true,
       experimentalFeatures: true
     }
-  }).on("close", (e) => {
-    // Close handled by the search renderer
-    e.preventDefault()
-    e.sender.webContents.send("close")
   })

   if (size) {
```

The preventDefault is probably interfering in the Spectron-on-Windows case. Why this happens with Spectron-on-Windows only isn't clear.

alfred-landrum:
When Spectron launches Electron instances on Windows, empty command prompt windows are spawned along with the Electron instance window. (Spectron internally uses a `.bat` file to run chromedriver, and Windows apparently always shows a command prompt window when you run a batch file). 

These command prompt windows steal focus from the Electron window. So an individual test may succeed or fail, but Spectron can't shutdown the instance, because it can't close its test window, because it's not focused.

I verified this locally by running the integration tests w/o any changes, and verifying that despite test success or failure, the only way the Electron instances would go away is if I put in a brief sleep so that I could manually click on the Electron instances to make them focused.

@jameskerr @mason-fish : this is really beyond my redux knowledge, so I won't be surprised if there's a better way to address this. Also, I see that the 'closeWindow' operation relies on a focused window, and it may make sense to ensure it doesn't rely on focus so we don't hit this again.

Fixes #598 . cc @mikesbrown

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
     strategy:
       matrix:
         node-version: [12.x]
-        os: [macos-10.15, ubuntu-18.04]
+        os: [macos-10.15, ubuntu-18.04, windows-2019]
     needs: build
     steps:
     - uses: actions/checkout@v2
--- a/src/js/electron/ipc/windows/mainHandler.js
+++ b/src/js/electron/ipc/windows/mainHandler.js
@@ -35,8 +35,9 @@ export default function(manager: $WindowManager) {
     })
   })
 
-  ipcMain.handle("windows:destroy", () => {
-    manager.destroyWindow()
+  ipcMain.handle("windows:destroy", (e) => {
+    let win = BrowserWindow.fromWebContents(e.sender)
+    manager.destroyWindow(win)
   })
 
   ipcMain.handle("windows:log", (e, {id, args}) => {
--- a/src/js/electron/tron/windowManager.js
+++ b/src/js/electron/tron/windowManager.js
@@ -97,8 +97,7 @@ export default function windowManager() {
       if (win) win.close()
     },
 
-    destroyWindow() {
-      let win = BrowserWindow.getFocusedWindow()
+    destroyWindow(win: BrowserWindow) {
       if (win) win.destroy()
     }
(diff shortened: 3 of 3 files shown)


jameskerr:
LGTM

PR #632, opened 2020-04-17
avoid Path vs PATH errors on windows

### LINKED ISSUE #556: Path not being set properly for zqd to find zeek on Windows in dev/ci env
In a Windows development environment where Brim is checked out and built, Brim cannot ingest a pcap because zeek can't be found.

This is the reason for CI in Windows failing. Why zeek can't be found is unclear. There's nothing in zqd logs that indicate this one way or another. I confirmed when I run npm run electron that the running zqd is brim.git\zdeps\zqd.exe.

Brim derives the path of zqd here, and zeek just after.

From https://github.com/brimsec/brim/blob/3b912ea0f689e1a90cc9684cf7c5c858e0130c63/src/js/zqd/zqd.js#L13 :

```
const zqdPath = join(app.getAppPath(), "zdeps")
const zqdZeekPath = join(zqdPath, "zeek")
```

The rest of the code in zqd.js looks reasonable. We've already established zqdPath is working, so it's odd zqdZeekPath isn't.

To reproduce:

1. Get a windows machine
2. Download Git and Node 12
3. Clone brim.git; cd brim
4. npm install && npm run build && npm run electron
5. Attempt to ingest a pcap and note error

alfred-landrum:
There is apparently a difference on trying to use "Path" vs "PATH" depending on whether you use cmd.com or powershell. Avoid needing to clone process.env and worry about that difference, and just modify it directly. Without this change, I would see occasional failures if I ran the integration tests locally in powershell, cmd.com, or git bash. With this change, I don't see any errors.

Fixes #556 .

### DIFF
--- a/src/js/zqd/zqd.js
+++ b/src/js/zqd/zqd.js
@@ -81,18 +81,14 @@ export class ZQD {
   start() {
     mkdirSync(this.root, {recursive: true, mode: 0o755})
 
-    // PATH must include both the zqd and zeek bin directories, as zqd depends
-    // on having zeek available in its environment PATH.
+    // We saw errors on command.com vs powershell when we tried to clone
+    // process.env and then determine whether to use "PATH" or "Path".
+    // Directly altering process.env is safe and less error prone.
     const sep = process.platform == "win32" ? ";" : ":"
-    const pathKey = process.platform == "win32" ? "Path" : "PATH"
-    const zqdEnvironment = _merge({}, process.env)
-    zqdEnvironment[pathKey] = [zqdPath, zqdZeekPath, process.env[pathKey]].join(
-      sep
-    )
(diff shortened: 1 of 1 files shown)


src/js/zqd/zqd.js
```diff
@@ -81,18 +81,14 @@ export class ZQD {
   start() {
     mkdirSync(this.root, {recursive: true, mode: 0o755})
 
-    // PATH must include both the zqd and zeek bin directories, as zqd depends
-    // on having zeek available in its environment PATH.
+    // We saw errors on command.com vs powershell when we tried to clone
+    // process.env and then determine whether to use "PATH" or "Path".
+    // Directly altering process.env is safe and less error prone.
```
mikesbrown, line 86: I didn't understand why, but the note at https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_process_env helps. Mind linking to that or repeating the claim that Windows environment variables are case-insensitive?

PR #637, opened 2020-04-17
test fix: test download pcaps (fixes #543)

### LINKED ISSUE #543: Revive pcap tests
@mikesbrown says:
> This was a side feature before that now is more important. There were 3 tests, and I think we can bring them back. The tests will take some time to work on to find proper flows with different test data. Note that this is to test integration points, not exhaustively test pcap handling.

mikesbrown:
Add integration tests for downloading packet captures. The tests are
accomplished by uploading data, running zql to find a deterministic
connection (can't use UID), clicking a cell, and downloading the packet
capture. In CI the files are put into a directory, where we can examine
the file's MD5.

In older times, there was a bug where a null-duration connection's pcap
could not be downloaded. That has since been fixed, so that test is
enabled.

### DIFF
--- a/itest/lib/app.js
+++ b/itest/lib/app.js
@@ -275,6 +275,9 @@ export const waitUntilDownloadFinished = async (app: Application) =>
     )
   })
 
+export const pcapsDir = async (app: Application) =>
+  await app.electron.remote.app.getPath("temp")
+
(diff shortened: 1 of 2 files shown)


mikesbrown:
All the work that's now on master has allowed this to be stable.

Issue #575, opened 2020-04-08
Store ingest errors in redux


jameskerr:
Right now we store the ingest data in the space object. We could add an errors array there.

Or we can make a new reducer call the 'ingests' reducer. This could then store the progress, and any errors it encountered. We'd have to index it by cluster and space just like we're doing now with the 'spaces' reducer.

I might like the second idea more.

PR #643, opened 2020-04-17
Save ingest warnings in space state

jameskerr:
Fixes: #575 

Store the ingest warnings in the redux store under the space. This works fine since a space is only ever allowed to be ingested into once. We may have to refactor where we put these errors in the future.


jameskerr:
Merging this because the test that is failing is unrelated to this PR and is known to be flaky.

PR #646, opened 2020-04-18
ci fix: properly handle windows paths in bash

### LINKED ISSUE #641: windows node cache slowing down CI
Looking at a long running Windows build, I noticed this in the npm cache action:

Cache Size: ~1485 MB (1557289122 B)

We tried to address this with https://github.com/brimsec/brim/pull/590 , and that definitely helped the macos builds.

mikesbrown:
Fixes #641 

Adjust how the npm cache directory is expanded. We need to quote it in
bash so that "\\" is not quashed. But it's not possible to directly quote
the Github action expression, because the running shell will attempt to
expand it using normal shell expansion rules. That means this doesn't
work:
```
  - run: rm -rf "${{ foo.bar }}"
```
Also,
```
  - run: rm -rf ${{ foo.bar }}
```
will not work in some cases, like if foo.bar is a Windows path. If
foo.bar is "c:\foo", the shell will turn that into "c:foo".

The fix is to assign an environment variable from the expression
directly, then quote the environment variable in the shell.

To force using a new, smaller cache, change the cache name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,17 +19,19 @@ jobs:
       id: set-npm-cache-dir
       run: echo "::set-output name=npm_cache_dir::$(npm config get cache)"
     - name: Clear Extraneous Runner Cache
-      # The macos Github runner has 1G of NPM cache that we don't need.
-      # Clear it before we create our own cache to prevent slower build
-      # times. See https://github.com/brimsec/brim/pull/590
-      run: |
-          cachedir=${{ steps.set-npm-cache-dir.outputs.npm_cache_dir }}
-          rm -rf "${cachedir:?}"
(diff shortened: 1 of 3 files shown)


mikesbrown:
Please witness the following similar bits of text in all 6 checks:
```
Run actions/cache@v1
Cache Size: ~78 MB (82009008 B)
C:\windows\System32\tar.exe -xz -f D:\a\_temp\548ee660-35ba-46e4-acd3-ab18b323eaf5\cache.tgz -C C:\npm\cache
Cache restored from key: Windows-build-cache-node-modules-ci-v4-12.x-be6690a8d42b2b8d5700f716107b2c835243536b74e6fce38acf128392831359
```

PR #653, opened 2020-04-20
Use deb package installer in integration tests

### LINKED ISSUE #649: adjust linux integration tests to use the deb package
The Linux integration tests should build and install the deb package, then run tests against that deployed image. We were able to do this with for the mac with some adjustments to the test:

https://github.com/brimsec/brim/blob/master/.github/workflows/ci.yml#L85
https://github.com/brimsec/brim/blob/master/itest/lib/app.js#L23
https://github.com/brimsec/brim/blob/master/itest/lib/app.js#L52

mattnibs:
closes #649

### DIFF
--- a/itest/lib/app.js
+++ b/itest/lib/app.js
@@ -21,7 +21,7 @@ const appStep = async (stepMessage, f) => {
 
 export const newAppInstance = (name: string, idx: number): Application => {
   const macInstallPath = "/Applications/Brim.app/Contents/MacOS/Brim"
-  const debianInstallPath = "/usr/bin/brim"
+  const linuxInstallPath = "/usr/bin/brim"
   const userDataDir = path.resolve(path.join(itestDir(), name, idx.toString()))
   mkdirpSync(userDataDir)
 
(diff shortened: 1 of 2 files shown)


itest/lib/app.js
```diff
@@ -52,6 +53,9 @@ export const newAppInstance = (name: string, idx: number): Application => {
   if (isCI() && process.platform === "darwin") {
     appArgs = {...appArgs, path: macInstallPath}
     LOG.debug("Chose installed MacOS app location", macInstallPath)
+  } else if (isCI() && process.platform === "linux") {
+    appArgs = {...appArgs, path: debianInstallPath, args: [repoDir()]}
+    LOG.debug("Chose working copy app location", electronPath)
```
mikesbrown, line 58:
```suggestion
  } else if (isCI() && process.platform === "linux") {
    appArgs = {...appArgs, path: linuxInstallPath}
    LOG.debug("Chose installed Linux app location", linuxInstallPath)
```

mikesbrown: Explanation for the above: We want to run just `/usr/bin/brim`, not `/usr/bin/brim /home/runner/brim`. The latter might point to the working copy, not the installed app. Note how in the darwin branch above, `args` is not included either.

PR #670, opened 2020-04-22
Throttle key events

### LINKED ISSUE #591: repeated key navigation for log detail locks up UI
If I open up a log details pane, then hold down the 'down' or 'up' keys, the UI locks up, and eventually all of the records briefly to display themselves at once in the pane. 
I saw this on macos and windows, using the latest master.

mason-fish:
fixes #591 

I also had noticed that `cmd+w` was closing the whole window instead of just the tab, so fixed that too (prevent default)

### DIFF
--- a/src/js/components/useSearchShortcuts.js
+++ b/src/js/components/useSearchShortcuts.js
@@ -5,21 +5,31 @@ import {useEffect} from "react"
 import Mousetrap from "mousetrap"
 
 import Tabs from "../state/Tabs"
+import throttle from "lodash/throttle"
 import {adjustSelectedLogIndex} from "../flows/adjustSelectedLogIndex"
 
 export default function() {
   let dispatch = useDispatch()
   useEffect(() => {
-    Mousetrap.bind("down", (e) => {
-      e.preventDefault()
-      dispatch(adjustSelectedLogIndex(1))
-    })
-    Mousetrap.bind("up", (e) => {
-      e.preventDefault()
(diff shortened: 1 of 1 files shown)


src/js/components/useSearchShortcuts.js
```diff
@@ -5,21 +5,31 @@ import {useEffect} from "react"
...
-      e.preventDefault()
-      dispatch(adjustSelectedLogIndex(1))
-    })
-    Mousetrap.bind("up", (e) => {
+    Mousetrap.bind(
+      "down",
+      throttle((e) => {
+        e.preventDefault()
+        dispatch(adjustSelectedLogIndex(1))
+      }, 200)
```
mason-fish, line 19: I experimented w a few values here, this seemed to be the fastest that my computer could handle without any odd glitching, and it still feels plenty fast to me

src/js/components/useSearchShortcuts.js
```diff
@@ -5,21 +5,31 @@ import {useEffect} from "react"
...
+    )
+    Mousetrap.bind(
+      "up",
+      throttle((e) => {
+        e.preventDefault()
+        dispatch(adjustSelectedLogIndex(-1))
+      }, 200)
+    )
+    Mousetrap.bind("mod+t", () => dispatch(Tabs.new()))
+    Mousetrap.bind("mod+w", (e) => {
```
jameskerr, line 29: 🙇‍♂️ Thank you! That had been bothering me so much!

PR #683, opened 2020-04-24
harden clicks

### LINKED ISSUE #668: deflake menu clicks
Sometimes an element, especially one in a menu, cannot be clicked.
```
 Query tests › query004: "* | every 2s count() | sort ts"

    element click intercepted: Element <li>...</li> is not clickable at point (867, 98). Other element would receive the click: <ul data-test-locator="span_menu">...</ul>

      at new RuntimeError (../node_modules/webdriverio/build/lib/utils/ErrorHandler.js:143:12)
      at Request._callback (../node_modules/webdriverio/build/lib/utils/RequestHandler.js:318:39)
      at Request.self.callback (../node_modules/request/request.js:185:22)
      at Request.<anonymous> (../node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (../node_modules/request/request.js:1083:12)
```
I believe we are trying to click on the element while it is still being animated.

While we are [constrained to using WebdriverIO v4](https://github.com/electron-userland/spectron/issues/349) without any sort of notion of `isClickable()`, WebdriverIO v5 has an [`isClickable()`](https://v5.webdriver.io/docs/api/element/isClickable.html). We might be able to write our own. We could also do better about retrying clicks.

mikesbrown:
Fixes #668

### DIFF
--- a/itest/lib/app.js
+++ b/itest/lib/app.js
@@ -204,15 +204,24 @@ export const setSpan = async (app: Application, span: string) => {
   await clickSpan()
 }
 
+const waitForClickable = async (app: Application, selector: string) => {
+  // In testing, it's been shown than there is no need to scroll to
+  // elements to make them be visible, as long as the element's
+  // container allows scrolling.
+  await appStep(`wait for element to be visible: "${selector}"`, () =>
+    app.client.waitForVisible(selector)
+  )
+}
+
 export const click = (app: Application, selector: string) =>
   appStep(`click on selector "${selector}"`, () =>
-    app.client.waitForVisible(selector).then(() => app.client.click(selector))
+    waitForClickable(app, selector).then(() => app.client.click(selector))
   )
 
 export const rightClick = (app: Application, selector: string) =>
   appStep(`right-click on selector "${selector}"`, () =>
     app.client
-      .waitForVisible(selector)
+      .waitForClickable(app, selector)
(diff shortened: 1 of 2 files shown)


itest/tests/query.test.js
```diff
@@ -33,6 +33,7 @@ describe("Query tests", () => {
     app = newAppInstance(basename(__filename), ++testIdx)
     await startApp(app)
     await pcapIngestSample(app)
+    await setSpan(app, "Whole Space")
```
jameskerr, line 36: 👍

itest/lib/app.js
```diff
@@ -194,35 +191,66 @@ export const getSearchSpeed = (app: Application) =>
...
     appStep(`select span ${span}`, () =>
-      app.client
-        .waitForVisible(selectors.span.menuItem(span))
-        .then(() => app.client.click(selectors.span.menuItem(span)))
+      click(app, selectors.span.menuItem(span))
     )
 
-  return clickSpanButton().then(() => clickSpan())
+  await clickSpanButton()
+  await clickSpan()
```
jameskerr, line 204: I don't see the reason that these need to be in-line functions. Any reason not to run the bodies of these functions directly?

mikesbrown: 👍

PR #684, opened 2020-04-25
integration test for rpm package

### LINKED ISSUE #618: Initial Linux release
We want to provide installable Linux releases of Brim. Our CI already builds and executes Linux based tests (using Ubuntu & xfvb). We have user reports about desired distros in #586 (note the comments there about flatpak.org) and a few other in the public #brim slack channel.

We currently use the Electron update service for autoupdates:
https://www.electronjs.org/docs/tutorial/updates

which does not support autoupdates on Linux. It will be acceptable if we don't initially support auto-update on Linux, but the user shouldn't see any errors or warnings about auto-update not working.

We need to:
- decide on the packaging format(s) that we want to provide. 
- add support for them in script/release as we do for other platforms
- add a CI trigger that will generate the release artifact & upload it as for other platforms
- add an icon and links to the Linux package(s) on the brimsecurity.com/download page.

mattnibs:
Add the build ci phase to build the linux release artifacts (.deb and
at the .rpm) at the end of the job then upload artifacts.

Add integration_test_centos which downloads the created rpm package
and runs the standard suite of integration tests against it.

Also:
- Have ubuntu integration tests download and run against the uploaded
deb package.

closes #618

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,63 @@ jobs:
     - run: npm run lint
     - run: npm run flow
     - run: npm test -- --maxWorkers=2 --ci
+    - name: Build Linux Packages
+      if: runner.os == 'Linux'
+      run: |
+        set -x
+        node scripts/release --linux
+    - name: Persist rpm package
+      if: runner.os == 'Linux'
+      uses: actions/upload-artifact@v1
+      with:
+        name: linux-release
+        path: ./dist/installers
+
+  integration_test_centos:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        node-version: [12.x]
+    container:
+      image: centos:8
+      env:
+        # This environment variable causes chrome to run with no sandbox- needed
+        # whenever chrome is run as root.
+        APPVEYOR: true
+    needs: build
+    steps:
+    - run: dnf install -y xorg-x11-server-Xvfb git nss-tools pulseaudio
+    - run: git config --global core.autocrlf false
+    - uses: actions/checkout@v2
+    - name: setup node version ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Set NPM Cache Directory
+      id: set-npm-cache-dir
+      run: echo "::set-output name=npm_cache_dir::$(npm config get cache)"
+    - name: Clear Extraneous Runner Cache
+      run: rm -rf "${NPM_CACHE:?}"
+      env:
+        NPM_CACHE: ${{ steps.set-npm-cache-dir.outputs.npm_cache_dir }}
+    - name: Cache node modules
+      uses: actions/cache@v1
+      env:
+        cache-name: cache-node-modules-ci-v4
+      with:
+        path: ${{ steps.set-npm-cache-dir.outputs.npm_cache_dir }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +44,63 @@ jobs:
     - run: npm run lint
     - run: npm run flow
     - run: npm test -- --maxWorkers=2 --ci
+    - name: Build Linux Packages
+      if: runner.os == 'Linux'
+      run: |
+        set -x
+        node scripts/release --linux
+    - name: Persist rpm package
```
mikesbrown, line 52:
```suggestion
    - name: Persist Linux packages
```

.github/workflows/ci.yml
```diff
@@ -44,6 +44,63 @@ jobs:
...
+        node-version: [12.x]
+    container:
+      image: centos:8
+      env:
+        # This environment variable causes chrome to run with no sandbox- needed
+        # whenever chrome is run as root.
+        APPVEYOR: true
+    needs: build
+    steps:
+    - run: dnf install -y xorg-x11-server-Xvfb git nss-tools pulseaudio
```
mikesbrown, line 72: Please file an issue to have us create our own centos8 image that contains this software.

.github/workflows/ci.yml
```diff
@@ -44,6 +44,63 @@ jobs:
...
+          ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.node-version }}-
+    - run: npm install --no-audit
+    - name: Download rpm package
+      uses: actions/download-artifact@v1
+      with:
+        name: linux-release
+    - name: Integration Tests (Centos 8)
+      run: |
+        dnf install -y ./linux-release/brim_x86_64.rpm
+        xvfb-run -d -s "-screen 0 1280x1024x24" npm run itest -- --ci --forceExit
```
mikesbrown, line 103: What does `xvfb-run -d` do?

.github/workflows/ci.yml
```diff
@@ -86,15 +143,16 @@ jobs:
         hdiutil attach dist/installers/Brim.dmg
         cp -R /Volumes/Brim/Brim.app /Applications
         hdiutil detach /Volumes/Brim
-    - name: Build and Install Debian Pkg
+    - name: Download deb package
```
mikesbrown, line 146:
```suggestion
    - name: Download Linux packages
```

.github/workflows/ci.yml
mattnibs: Sure. Do we have a repo where these images are created? Are they attached to ci?

mikesbrown: When we had circleci, a dockerfile was put in the .circleci directory, and the image the dockerfile created was used in CI. So far we only use others' images. I don't believe a separate repository is needed.

.github/workflows/ci.yml
mattnibs: The newer version of `xvfb` which centos8 has but ubuntu 18 does not says that `--auto-servernum` is deprecated and to use `-d` instead.

.github/workflows/ci.yml
mikesbrown: Er, I was assuming you meant Git repository. We have a brimsec org on docker hub where a built image can be uploaded.

mattnibs: #701

.github/workflows/ci.yml
mikesbrown: TIL. Thanks!

PR #691, opened 2020-04-27
retry dimg detach

### LINKED ISSUE #690: macos CI occasionally fails when detaching installer volume
```
+ hdiutil detach /Volumes/Brim
hdiutil: couldn't eject "disk2" - Resource busy
```
https://github.com/brimsec/brim/runs/618623681?check_suite_focus=true

mikesbrown:
Fixes #690


lib/bash/common.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+# Source this file to take advantage of functions.
```
alfred-landrum, line 2: This script looks good to me, but how about putting it a different location, somewhere under `scripts`? Perhaps `scripts/lib/common.bash`?

mikesbrown: 👍

PR #692, opened 2020-04-27
add debounce to component props

### LINKED ISSUE #591: repeated key navigation for log detail locks up UI

mason-fish:
fixes #591 

Or attempts to...turns out that the crash fix is two-fold, one part came from throttling the key event so that we didn't make hundreds of network requests at once. I had thought that the component updating and re-rendering would be able to handle the same delay but in fact this was not enough. Adding a debounce for the corresponding component props was the only way I could calm down the rendering process without setting a much higher (and less user friendly) throttle on the key event. I played with the debounce delay quite a bit and it seems like only 50ms is enough to rate limit the rendering and prevent the crash.

debounceprops

NOTE: I did notice that holding down the button will still occasionally delay updating what log is selected and then it catches up within a second. We could further smooth this by tuning that throttle up a bit more but I thought we'd keep it as is since I anticipate most folks would only use the down button in smaller bursts anyway.


jameskerr:
I see what you're running up against. Let me propose an alternative solution.

Something I've done in the past  (when selecting a tab), is hold the "selected tab" state locally in the component (with useState) to update instantly when you select a log. Then add an debounced effect that updates the redux store once 100ms has gone by since the selected log has changed.

https://github.com/brimsec/brim/blob/master/src/js/components/TabBar/useTabController.js#L13

https://github.com/brimsec/brim/blob/master/src/js/components/TabBar/useTabController.js#L21-L23

philrz:
Verified in Brim commit `670d3b8` talking to `zqd` commit `a09d0d6`. Per the attached video, now I can hold down the arrow keys and it'll walk through the events without locking up the app.

[Verify.zip](https://github.com/brimsec/brim/files/4566769/Verify.zip)

Thanks @mason-fish!

PR #700, opened 2020-04-28
fix delete text

### LINKED ISSUE #686: Erasing Logs space includes message that will preserve original pcap file
Requesting a change of text message to one that is neutral for both Log and pcap scenarios. Example: "This will delete the created .brim folder, but will preserve the original data."

A user who is familiar with Brim might probably understand, however using a text message that applies both to Log case and pcap case is clearer and more user friendly. 

Figure: the window message appearing when erasing space

mason-fish:
fixes #686 

image


mason-fish:
**update**: removed the detail altogether. Once we start changing the name of the space to not include `*.brim` then there will be no more references to `.brim` to the user in this message as well.

image

PR #707, opened 2020-04-28
fix: use HTTPS to talk to zeek.org for "Link to Docs" in Log Details

mikesbrown:


mikesbrown:
The failure is an unrelated `hdiutil detach` problem.

PR #714, opened 2020-04-29
keep brim folders inside app data

### LINKED ISSUE #673: Stop creating the .brim folder
We're going to stop creating the .brim folder next to ingested data, By default, we should *not* specify a data_dir in the space create api call; so the zng data & other space information will be created under zqd's `data/spaces` directory, under Electron's per-platform `userData` directory.

mason-fish:
fixes #673 

Our spaces are still named `.brim` but the internal folder we use is now stored within the app's data. I tested this in dev and also release.

movebrimfolder

### DIFF
--- a/src/js/flows/ingestFiles.js
+++ b/src/js/flows/ingestFiles.js
@@ -57,22 +57,29 @@ const validateInput = (paths) => ({
 
 const createDir = () => ({
   async do({dataDir}) {
-    await fsExtra.ensureDir(dataDir)
+    dataDir && (await fsExtra.ensureDir(dataDir))
   },
   async undo({dataDir}) {
-    await fsExtra.remove(dataDir)
+    dataDir && (await fsExtra.remove(dataDir))
   }
 })
 
 const createSpace = (client, dispatch, clusterId) => ({
   async do(params) {
-    let {name} = await client.spaces.create({data_dir: params.dataDir})
+    let createParams
+    if (params.dataDir) {
+      createParams = {data_dir: params.dataDir}
+    } else {
+      createParams = {name: params.name}
+    }
+    let {name} = await client.spaces.create(createParams)
     dispatch(
       Spaces.setDetail(clusterId, {
         name,
         ingest: {progress: 0, snapshot: 0, warnings: []}
       })
     )
+
     return {...params, name}
   },
   async undo({name}) {
--- a/src/js/flows/initSpace.js
+++ b/src/js/flows/initSpace.js
@@ -56,6 +56,7 @@ function getCurrentSpaceName(spaces, desired) {
 function setSpace(dispatch, data, clusterId) {
   globalDispatch(Spaces.setDetail(clusterId, data))
   dispatch(Search.setSpace(data.name))
+  data = brim.interop.spacePayloadToSpace(data)
   return data
 }
 
--- a/src/js/brim/ingest/getParams.js
+++ b/src/js/brim/ingest/getParams.js
@@ -5,6 +5,7 @@ import path from "path"
 import type {IngestFileType} from "./detectFileType"
 import fileList, {type FileListData} from "./fileList"
 import time from "../time"
+import lib from "../../lib"
 
 export type IngestParams = {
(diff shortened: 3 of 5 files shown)


src/js/flows/ingestFiles.js
```diff
@@ -51,16 +51,23 @@ const validateInput = (paths) => ({
 
 const createDir = () => ({
   async do({dataDir}) {
-    await fsExtra.ensureDir(dataDir)
+    dataDir && (await fsExtra.ensureDir(dataDir))
```
jameskerr, line 54: It seems like this whole step will be removed once https://github.com/brimsec/brim/issues/676 is done. All we might do is check that the user specified directory exists.

src/js/flows/ingestFiles.js
```diff
@@ -51,16 +51,23 @@ const validateInput = (paths) => ({
...
   },
   async undo({dataDir}) {
-    await fsExtra.remove(dataDir)
+    dataDir && (await fsExtra.remove(dataDir))
```
jameskerr, line 57: And if they did provide a directory where all their spaces will go, it seems like we would not want to delete it for them.

src/js/flows/ingestFiles.js
jameskerr: Wait, I might understand the data_dir thing correctly. This might be fine.

src/js/flows/ingestFiles.js
```diff
@@ -51,16 +51,23 @@ const validateInput = (paths) => ({
...
 })
 
 const createSpace = (client) => ({
   async do(params) {
-    let {name} = await client.spaces.create({data_dir: params.dataDir})
+    let createParams
+    if (params.dataDir) {
+      createParams = {data_dir: params.dataDir}
+    } else {
+      createParams = {name: params.name}
```
jameskerr, line 67: Is there a reason why we don't send the name in all cases?

mason-fish: I opted to send one **or** the other bc zqd only handles for one (https://github.com/brimsec/zq/blob/master/zqd/space/space.go#L108-L123) (giving preference to `name` by handling it first) and I thought it might confuse anyone looking at this code to see that it sends both but only uses `name`.

jameskerr:
After testing this, I see some limitations here. 

After opening one pcap...

* When I try to load the pcap for the second time, I get the message from zqd saying "Space exists", and get put back on the New Tab Page.
* When I try load a different pcap (in a different dir) with the same name, I get the same error "Space exists".


image


In the first situation it seems we ought open that space for them. The second situation seems like a new bug. 

I think we'd need to do this ZQD issue first  https://github.com/brimsec/zq/issues/686 (generate space ids), before we'll be able to merge this PR without introducing bugs.

Does that sound right?

mason-fish:
@jameskerr I'm inclined to agree. If we make the space id change first in the backend then I'm p sure this will work as is, which includes continuing to be able to make duplicate spaces if a user wants (though it may be better to not allow that and instead detect the duplicate on the FE and take the user direct to the existing space, as you suggest)

src/js/flows/initSpace.js
```diff
@@ -56,6 +56,7 @@ function getCurrentSpaceName(spaces, desired) {
 function setSpace(dispatch, data, clusterId) {
   globalDispatch(Spaces.setDetail(clusterId, data))
   dispatch(Search.setSpace(data.name))
+  data = brim.interop.spacePayloadToSpace(data)
```
mason-fish, line 59: @jameskerr  The problem is that the span parsing we are doing rn is late in the stack, in the reducer (https://github.com/brimsec/brim/pull/715/files#diff-31bf947bfb654ba8ccfd57214bfb417dR21), and yet the original span which has not been put through that adapter/parser continues to get passed along by this function to the rest of the initSpace chain. So I've adjusted this to do the same operation on a different copy and return that. Not a great pattern, would be better if we can handle this mutation inside the client, or at least much earlier before it gets passed around in different formats.

Issue #703, opened 2020-04-28
Support zqd pcap endpoint api changes


mattnibs:
There's a couple of subtle changes coming down the pike for the zqd api.

See:
- brimsec/zq#668
- brimsec/zq#657

Once these get merged, the brim client will need to be updated to work with the latest version of zqd.

To enumerate the api changes:

Path updates:
- `POST /space/:space/packet` -> `POST /space/:space/pcap`
- `GET /space/:space/packet` -> `GET /space/:space/pcap`

SpaceInfo Response (returned in GET /space/:space):
- Fields `min_time` and `max_time` have been consolidated into a single `span` field. This is the same struct used in search for the `span` field.
- `packet_support` -> `pcap_support`
- `packet_size` -> `pcap_size`
- `packet_path` -> `pcap_path`

PcapPostStatus payload (stream in POST /space/:space/pcap endpoint)
- `min_time` and `max_time` fields have been consolidated into `span` field (see above)
- `packet_size` -> `pcap_size`
- `packet_read_size` -> `pcap_read_size`

LogPostStatus payload (stream in POST /space/:space/log endpoint)
- `min_time` and `max_time` fields have been consolidated into `span` field (see above)

PR #715, opened 2020-04-29
Support zqd api span changes

mattnibs:
Provide support for changes to the zqd api the changed min/max times
to spans in certain payloads. Changes were made in a way that the old
version will still work.

Also:
- Use big-integer package for adding together ts and duration to ensure
fidelity is not lost.

part of #703


src/js/state/Spaces/reducer.js
```diff
@@ -15,7 +16,20 @@ const spacesReducer = produce((draft, action: SpacesAction) => {
 
     case "SPACES_DETAIL":
       var {name} = action.space
-      draft[name] = {...draft[name], ...action.space}
+      var space = (action.space: any)
```
jameskerr, line 19: Makes sense to support both. Are you willing to move this logic to a new function in `src/js/brim/interop.js` then call it in here?

That's were functions go to convert between old and new types.

src/js/brim/time.js
```diff
@@ -41,12 +48,28 @@ function time(val: Ts | Date = new Date()) {
       return brim.time(ts)
     },
 
+    addDur(dur: Ts) {
```
jameskerr, line 51: Can we call this `addTs` to match `subTs` below.

src/js/state/Spaces/reducer.js
mattnibs: sure, make sense

src/js/brim/time.js
mattnibs: I can do that for now, but I would say this function is more for adding a duration to a timestamp- it would be kind of weird to add two actual timestamps togther. My use of the `Ts` type here is a little hacky- it was just easy because the payload for `ts` and `dur` in backend is the same (this is because the both are fundamentally int64s of nanoseconds and need to be split up because javascript numbers will lose precision) Probably the actual type here should be `bigint` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt). Excitingly this is now part of the builtin js api. Not-excitingly flow doesn't yet support this (thus why I had to import `big-integer`).

All that said, seeing as you and mason are the vanguards of this codebase, and I but a humble visitor from abroad, to your good judgment I will defer.

Issue #687, opened 2020-04-25
Attempted Zeek ingest of good+bad files halts and shows confusing error


philrz:
Found in Brim commit `1fae48f` talking to `zqd` commit `a2de891`.

In this case I'm adopting the persona of a Zeek user that's trying to drag the contents of a Zeek log directory into the app. Zeek is known to generate a handful of not-Zeek-format-logs in such a directory (e.g. the `conn-summary` log and the `stderr`/`stdout` logs). As there's so many logs in the directory, it would be an excessive burden on the user for them to delete/unclick each of the "bad" ones before dragging the contents into Brim. Therefore it's important that Brim ingest what's good, be tolerant of the "bad" ones, and present some helpful warnings so the user can at least review what was not ingested and confirm that they're ok with that.

I've simplified things here. Per the attached video, I have two files I'm attempting to ingest. One contains a single valid Zeek NDJSON event:

```
# cat bar.ndjson
{"_path":"pe","_write_ts":"2015-03-06T13:38:28.016893Z","ts":"2015-03-06T13:38:20.736865Z","id":"FxADxanpFc2GG166e","machine":"I386","compile_ts":"2012-02-17T14:55:26.000000Z","os":"Windows 2000","subsystem":"WINDOWS_GUI","is_exe":true,"is_64bit":false,"uses_aslr":false,"uses_dep":true,"uses_code_integrity":false,"uses_seh":false,"has_import_table":true,"has_export_table":true,"has_cert_table":true,"has_debug_data":true,"section_names":[".text",".rdata",".data",".CRT",".rsrc"]}
```

The other contains text that's neither pcap nor ingestible Zeek.

```
# cat not.ndjson 
not even json!
```

When each is ingested separately, the valid one loads in fine, and for the bad one an error message is popped up that says "Unknown file types" (side note: this should probably be singular "type", but that's not the primary issue here). Both of those are expected behaviors. However, if I try to drag both files into Brim as part of the same ingest attempt, now I get an error message "Only files of a single type (zeek or pcap) can be opened", which is confusing.

What I'd have expected is ingest of the "good" log and information about the "bad" one among the messages as shown in https://github.com/brimsec/brim/issues/577.

[Repro.zip](https://github.com/brimsec/brim/files/4533697/Repro.zip)

philrz:
Verified in Brim commit `6dac8af` talking to `zqd` commit `233e930`.

[...]

PR #716, opened 2020-04-29
If its not a pcap, its a log

jameskerr:
fixes: #687

Previously we were checking each file to see if was a pcap, zeek, json, or unknown type. Now we just check if any one of them is a pcap. If there is 1 pcap, we send it over the /pcaps/ endpoint, else /logs/.

This makes it simple and lets the backend do the rest of the auto detection. I've been even able to ingest a json and a zeek tsv at the same time now.

### DIFF
--- a/src/js/brim/ingest/detectFileType.js
+++ b/src/js/brim/ingest/detectFileType.js
@@ -1,38 +1,19 @@
 /* @flow */
 import fs from "fs"
-import readline from "readline"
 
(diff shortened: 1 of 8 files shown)


mikesbrown:
The symptoms in the console log looked like https://github.com/brimsec/zq/issues/613 but the artifacts for the failed run including any screen shots taken were removed on the rerun before I could get at them, so I can't be sure.

PR #719, opened 2020-04-30
Custom types config

### LINKED ISSUE #694: Allow user to provide a custom JSON typing schema
In its current state as of Brim commit `f837314`, JSON ingest is ready for the "happiest path" of a user that has Zeek JSON logs that conform to the schema we ship that's based on "stock" Zeek v3.1.2. However, as soon as their environment has an additional plugin or any other customization, they're now prone to seeing warnings that may confuse them and/or missing data, which will generate Support inquiries.

To deal with this, I'm planning to write a user-facing article that describes how to use `zq` to create a modified JSON typing schema. The only problem with this as things currently stand is that there's not a great way for them to put the schema to use. Two possibilities I see:

1. (As I think we've discussed in the past...) We could require them to use `zq` to apply typing to their Zeek JSON logs and turn them into ZNG before posting them into Brim. This would require us to finish https://github.com/brimsec/brim/issues/693. In an absolute pinch, I'd be prepared to point to this option in our next release, but it's not very user-friendly.

2. I could advise them to overwrite the built-in schema with their custom one, e.g. on MacOS, copy it to `/Applications/Brim.app/Contents/Resources/app/config/defaultTypes.json`. That would give the seamless experience of being able to drag Zeek JSON logs directly into the app, but creates a supportability hazard since now when they upgrade Brim it'll overwrite their custom schema.

Since we're already so close, I find myself wishing we could just have a place where the user could point to an alternate path for the JSON schema. Since JSON ingest is a bit of a "power user" use case, I'd be fine with _not_ investing significant cycles right now to make this a front & center part of the ingest UX. Instead, something simple in the Preferences menu would totally cover us here.

@alfred-landrum and I were discussing this and liked the sound of something like a checkbox that starts out checked that means the default built-in schema would be used. When unchecked, it would activate a way for the user to enter an alternate path to their custom schema. That way I could reference this option from the article so power users could find it, and all the non-JSON users would be blissfully ignorant.

jameskerr:
fixes #694 

Adds a global user preference for which JSON type config file the app should use when ingesting json files. If it is empty, then we use the default provided within the app. Otherwise, we read the contents of that file each time we ingest, and send it over in the payload of our POST /logs/ request.

Of interest:
* Created a useDropzone hook for making dropzones
* Created a new global state slice called Prefs to hold user preferences
* If there is no search window opened when the user selects "Preferences...", the app will now create new one and open the preferences in it

**Providing a custom types.json and ingest a custom ndjson file** 

happy_path

**A little bit of validation (check if the file exists and if its json)**

error_validation

**What happens when you select a garbage file as your types_config**

garbage_json_config


src/js/components/hooks/useDropzone.js
```diff
@@ -0,0 +1,39 @@
+/* @flow */
+import {useState} from "react"
+
+/*
+  When using this, all the children elements of the dropzone must have a
+  css rule of pointer-events: none.
+*/
+
+export default function useDropzone(dropCallback: Function) {
```
mason-fish, line 9: 🔥

Issue #698, opened 2020-04-28
Wireshark buttons inactive when re-ingesting right after deleting Space 


philrz:
I stumbled onto this while verifying https://github.com/brimsec/brim/issues/652. Not sure if that fix maybe was incomplete or introduced a new problem. Repro is using Brim commit `f837314` talking to `zqd` commit `0e6240b`. I happen to use the attached [hello.pcapng](https://app.zenhub.com/files/144212075/71ec6613-76f4-4ed1-b308-6b14162342cb/download) as test data.

1. Load the pcap file into the first tab and notice the Wireshark button is active when I click on events
2. Click to the second tab and delete the Space
3. Go back to the first tab and dismiss the message about the deleted Space
4. Immediately repeat the load of the same pcap file into the first tab
5. Click events and notice the Wireshark button is _not_ active when I click on events, no matter how long I wait

DevTools shows a ContentType-related error, but that doesn't seem catastrophic.

Repro video: [Repro.mov.zip](https://github.com/brimsec/brim/files/4546901/Repro.mov.zip)

philrz:
Verified in Brim commit `670d3b8` talking to `zqd` commit `a09d0d6`. Per the attached video, now when I repeat the original repro steps, in the re-used tab I'm able to re-import the same pcap and the Wireshark button is activated as expected.

[Verify.zip](https://github.com/brimsec/brim/files/4566791/Verify.zip)

Thanks @jameskerr!

PR #722, opened 2020-05-01
Wireshark button bug not active when re-opening a deleted space

jameskerr:
fixes: #698

**Problem:**
When a space is deleted, but a tab still references it, we left some stale state about that deleted space around, causing bugs like this.

**Solution:**
Add a new logic flow that runs when a tab is switched to. It checks to see if that space was deleted, and clears all the state out.

I also updated the error message to be more specific about what happened.

image


src/js/flows/initNewTab.js
```diff
@@ -0,0 +1,23 @@
...
+import refreshSpaceNames from "./refreshSpaceNames"
+
+export default (): Thunk => (dispatch, getState) => {
+  let state = getState()
+  let space = Tab.space(state)
+  let name = Tab.spaceName(state)
+  let id = Tab.clusterId(state)
+  let spaceIsDeleted = name && !space
+
+  if (spaceIsDeleted) {
```
mason-fish, line 17: nice!

Issue #702, opened 2020-04-28
remove 'make build' when installing zq at arbitrary hashes


mikesbrown:
Once zq's package.json contains the necessary build steps, `make build` on this side can be removed. This is needed for #634

PR #730, opened 2020-05-01
Install zq hashes

### LINKED ISSUE #699: add golang to CI host
To solve #634 it will be necessary to install golang on CI hosts so that zq at arbitrary commit hashes can be built.

mikesbrown:
It's possible to npm install zq from an arbitrary Github hash as long as you choose a zq hash that supports it. While npm's syntax for doing this is flexible, the only one that seems to use HTTPS is of this form:
`npm install https://github.com/brimsec/zq#086719e76d53ace254cc495b4cc21d2efe069e73`. The shorthand like `github:brimsec/zq#hash` use the `git://` URI scheme.

See https://github.com/brimsec/zq/pull/712 for the companion PR. It does the heavy lifting. Here, we need only copy into zdeps where needed and set up Go.

Fixes #699 #702

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "rimraf": "^3.0.2",
     "source-map-support": "^0.5.16",
     "valid-url": "^1.0.9",
-    "zq": "brimsec/zq#v0.12.0"
+    "zq": "git+https://github.com/brimsec/zq.git#335d284bfb0a5a4ced3e9cc3070c9270050f2376"
   },
   "optionalDependencies": {
     "electron-installer-debian": "^3.0.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -144,7 +144,7 @@
     "rimraf": "^3.0.2",
     "source-map-support": "^0.5.16",
     "valid-url": "^1.0.9",
-    "zq": "brimsec/zq#v0.12.0"
+    "zq": "git+https://github.com/brimsec/zq.git#335d284bfb0a5a4ced3e9cc3070c9270050f2376"
```
mikesbrown, line 147: Rebased, and this hash is at master.

PR #738, opened 2020-05-05
Brim wiki article about supported platforms

philrz:
This article provides details on the platforms on which we expect to deliver consistent quality (since we run on them in CI) vs. "best effort".

I could definitely use close review on the sections where I call out to the Electron and Go docs. I figure it's worth going to that level of detail since some in the open source community may be able to make use of this information. Since Brim depends on both Electron and Go, I basically took the approach of saying that anything _older_ _than_ the the earliest platform supported by _both_ should _not_ be expected to work. For instance, this is why I set expectations that anything older than macOS 10.11 would not work, since that was the oldest one in in Go's platform statement, though Electron claims to be ok to 10.10.

Closes #642

### DIFF
--- a/docs/Supported-Platforms.md
+++ b/docs/Supported-Platforms.md
@@ -23,8 +23,6 @@ all) for some platform versions.
 
 # Per-Platform Details
 
-The following details are current as of Brim `v0.9.0`.
-
 ## Windows
 
 Our Windows test automation runs on Windows Server 2019, and therefore this is
--- a/docs/Supported-Platforms.md
+++ b/docs/Supported-Platforms.md
@@ -19,7 +19,7 @@ continuously as well as occasional ad hoc testing. We welcome you to
 [open an issue](Troubleshooting.md#opening-an-issue) on any problem you
 experience with Brim regardless of platform version, but please understand
 that we may be limited in our ability to provide quick fixes (or any fix at
-all) for certain platforms, subject to availability.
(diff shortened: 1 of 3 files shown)


docs/Supported-Platforms.md
```diff
@@ -0,0 +1,82 @@
...
+versions/distributions of these platforms for which we can set expectations of
+quality are affected by end-of-life cycles and their availability in our
+automated test infrastructure.
+
+The sections below describe the details regarding the versions we test with
+continuously as well as occasional ad hoc testing. We welcome you to
+[open an issue](Troubleshooting.md#opening-an-issue) on any problem you
+experience with Brim regardless of platform version, but please understand
+that we may be limited in our ability to provide quick fixes (or any fix at
+all) for certain platforms, subject to availability.
```
alfred-landrum, line 22:
```suggestion
all) for some platform versions.
```

docs/Supported-Platforms.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Per-Platform Details
+
+The following details are current as of Brim `v0.9.0`.
```
alfred-landrum, line 26: Could we change this to say that it reflects the current latest Brim version, w/o specifying the version number? I don't want to us need to update this for every release.

philrz: Dropped the sentence.

PR #739, opened 2020-05-05
Use a better error messages for ingest interrupt

### LINKED ISSUE #533: Close/reopen laptop lid - Brim shows Network error
This can be considered low priority, as it does not always happen. It's hard to reproduce and may be related to the size of ingested pcaps. Here's a screenshot of one time the problem reproduced:

image

More detail from the user who originally reported it:

> When this happened, the ingest process stopped and I had to open file again and start over. I found the same problem when the screen saving/automatic lock of account happened.

jameskerr:
Screen Shot 2020-05-04 at 5 19 11 PM

This would happen during ingest if someone closes their laptop
then re-opens it, or if the computer falls asleep.

If the computer sleeps before the first IngestStatus is returned,
the error message is "Failed to fetch". If it happens after the
first status is returned, the error message is "network error".

We now look for both of these and present the importInterupt error
message.

Screen Shot 2020-05-04 at 5 19 53 PM
Screen Shot 2020-05-04 at 5 19 25 PM

fixes #533

### DIFF
--- a/src/js/components/TabWelcome.js
+++ b/src/js/components/TabWelcome.js
@@ -11,6 +11,7 @@ import Notice from "../state/Notice"
 import SavedSpacesList from "./SavedSpacesList"
 import Spaces from "../state/Spaces"
 import Tab from "../state/Tab"
+import errors from "../errors"
 import ingestFiles from "../flows/ingestFiles"
 import initNewTab from "../flows/initNewTab"
 import refreshSpaceNames from "../flows/refreshSpaceNames"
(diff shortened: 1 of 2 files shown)


jameskerr:
I should also note that we cannot catch the specific network errors (like ERR_NETWORK_IO_SUSPENDED that chrome emits. They are untouchable by javascript which I learned after seeing this stack overflow comment: https://stackoverflow.com/a/48182975/3499804

PR #744, opened 2020-05-05
fix windows autoupdate

### LINKED ISSUE #552: enable auto-update for windows
Now that we have code-signing support for our windows builds, we should be able to create the zip needed for the auto-update code.

EDIT: Our reading of the docs says that the exe file we produce should be enough for windows auto-update to work, but it doesn't appear to do so on the recent v0.7.0 release.

mason-fish:
fixes #552 

After digging through a fair amount of (un)documentation, the fix here has a few pieces but they are all small:

*  First off, there is a bizarre and inconsistent issue where if we try to run the updater immediately on startup then there can be an issue with multiple processes racing for resources. I only saw this once but since I added the 30s delay I never saw it again.

* Second, the `path.join` works great for cross platform filepaths, but is not designed for url construction.

* Third, in addition to the *.exe file, we also need to begin publishing the `RELEASES` and `*.nupkg` files which are already being generated by winstaller.

### DIFF
--- a/.github/workflows/win-release.yml
+++ b/.github/workflows/win-release.yml
@@ -54,7 +54,7 @@ jobs:
       env:
         WINDOWS_SIGNING_PASSPHRASE: ${{ secrets.WINDOWS_SIGNING_PASSPHRASE }}
         WINDOWS_SIGNING_PFX_BASE64: ${{ secrets.WINDOWS_SIGNING_PFX_BASE64 }}
-    - name: upload release assets
+    - name: upload release executable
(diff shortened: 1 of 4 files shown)


mason-fish:
@alfred-landrum Yes that's correct. So we can expect the first wave of windows autoupdates to trigger when we release `v.0.10.0`, assuming this PR goes into `v0.9.0`.

jameskerr:
If you rebase you won't run into that format error anymore.

PR #745, opened 2020-05-05
Dis-ambiguate the regex literal

jameskerr:
This is to get master back to green since prettier wants a semi-colon but eslint does not.

The future fix is to have prettier do all the re-formatting rules, while eslint only has qualitative rules.


jameskerr:
Thanks @mikesbrown

PR #749, opened 2020-05-05
Render html menus in integration tests

jameskerr:
We intercept the call to create the rightclick menu right before it gets sent to electron. Then we dispatch our own event to the dom, then build a simple html context menu that will only be used when `process.env.BRIM_ITEST === "true"`.

D6eZyhUISQ

To test this locally, start the app normally, then open the console and type in the following:

image

Now the app will respond to right clicks by building an html version of the context menus.

@mikesbrown this should make those integration tests much easier to write.

### DIFF
--- a/src/js/lib/System.js
+++ b/src/js/lib/System.js
@@ -2,10 +2,8 @@
 
 import {remote} from "electron"
 
-let test = true
-
 export function showContextMenu(template: Object) {
-  if (test) {
+  if (process.env.BRIM_ITEST === "true") {
     document.dispatchEvent(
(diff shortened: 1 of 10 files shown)


src/js/lib/System.js
```diff
@@ -2,6 +2,14 @@
 
 import {remote} from "electron"
 
+let test = true
+
 export function showContextMenu(template: Object) {
-  new remote.Menu.buildFromTemplate(template).popup()
+  if (test) {
```
jameskerr, line 8: @mikesbrown is there a way for the app to know that it's being run in an integration test?

PR #759, opened 2020-05-06
bare repository_dispatch handler

mikesbrown:
https://help.github.com/en/actions/reference/events-that-trigger-workflows#external-events-repository_dispatch says

> This event will only trigger a workflow run if the workflow file is on the master or default branch.

Start with this bare one in order to at least show zq can communicate to brim.


mikesbrown:
Thanks @mattnibs !

Issue #733, opened 2020-05-04
Date format - please support ISO 8601 


jedd:
Greetings,

Great tool, but the 05/04/20 format is at odds with most other security tooling, which lets you pick a date format - eg the naturally sortable and non-ambiguous ISO 8601  (yyyy-mm-dd hh:mm:ss.xxx)

The above time in Australia (and quite a few other places) would likely be assumed to mean 5th April 2020 -- but brim shows that for 4th of May 2020.

siskojr:
Hi @Jedd, 

Thanks for your suggestion. Makes sense, we are using the default format for a date in the US, and I know it is different in other countries, hence confusing. 

We will improve here. 
The time representation for the space is a bit more clear and I understand the format of the timestamp is the problem.


If you could join our slack channel would be more convenient to talk if you have more ideas for feature improvements, happy to consider them.

The link to join our slack: https://join.slack.com/t/brimsec/shared_invite/zt-cy34xoxg-hZiTKUT~1KdGjlaBIuUUdg

Appreciated.

Javier Rodriguez

Ref: ![image.png](https://images.zenhubusercontent.com/5e8b7dd61f02b3461f7761d3/d791f7f4-236e-4f3a-90a7-f8707b1089aa)

philrz:
Verified in Brim commit `09559ee`. ISO 8601 timestamps are the new default:

![image](https://user-images.githubusercontent.com/5934157/81846369-d428af80-9506-11ea-957b-aa982d9f5f79.png)

The **Preferences** menu now also offers an option to change this default.

![image](https://user-images.githubusercontent.com/5934157/81846493-089c6b80-9507-11ea-8c42-af213fc8526e.png)

Plugging in the example format string `dddd, MMMM Do YYYY, h:mm:ss a` from the [doc](https://momentjs.com/docs/#/displaying/format/) that's linked from that **Preferences** page, I can see the modified format.

![image](https://user-images.githubusercontent.com/5934157/81846558-210c8600-9507-11ea-938f-8615560a4e69.png)

This functionality should go out in the next Brim release, which is expected to be tagged `v0.10.0` and is likely to go out within the next week.

Thanks @jameskerr!

PR #766, opened 2020-05-11
Time Format for in results

jameskerr:
fixes: #733 

Adds a Time Format field to the preferences panel. By default it uses the standard ISO-8601 format. You can specify the format you want by using the [momentjs guidelines](https://momentjs.com/docs/#/displaying/format/)  (which are linked to in the preferences panel).

image

image

Screen Shot 2020-05-11 at 4 17 14 PM

**Notes about implementation:**

The form on the preferences page now has 3 inputs. They each had their own validation and submission. I looked into standardizing the way the form inputs work in this component, since its the only normal form we have in the app. 

I looked at [Formik](https://jaredpalmer.com/formik/) and [react-hook-forms](https://react-hook-form.com/). Their APIs consolidated all their logic (validation, submission) in the parent component, then all the other components were controlled by it. I liked this approach, but I didn’t like that there was only one submit and one validate function for all the inputs. I like when related code is co-located (the way `useEffect()` returns it's “cleanup” function). 

So I made my own object that stored all the data and behavior each form field needed.
```js
let formConfig = {
  timeZone: {
      name: "timeZone",
      label: "Timezone",
      defaultValue: useSelector(View.getTimeZone),
      submit: (value) => dispatch(View.setTimeZone(value)),
      check: (value) => [!isEmpty(value), "must not be blank"]
  }
}
```

Then another object `brim/form.js` that accepts all this config, along with the HTML form element. It has functions like `isValid()`, `submit()`, and `getErrors()`.

I read somewhere about very usable form error pattern that I’ve implemented here. It puts all the errors at the top of the form with links to focus the inputs that they are related to.

image


src/css/_settings-modal.scss
```diff
@@ -1,65 +1,80 @@
...
+      @include label-small;
+    }
 
-  label {
-    @include label-normal;
-  }
+    select {
+      border: none;
+      height: 24px;
+      border: 1px solid rgba(0, 0, 0, 0.1);
```
mason-fish, line 18: duplicate `border` 😉

src/js/components/useSearchShortcuts.js
```diff
@@ -20,6 +21,7 @@ export default function() {
       Mousetrap.bind(`mod+${i + 1}`, () => dispatch(Tabs.activateByIndex(i)))
     }
     Mousetrap.bind("mod+9", () => dispatch(Tabs.activateLast()))
+    Mousetrap.bind("mod+,", () => dispatch(Modal.show("settings")))
```
mason-fish, line 24: 👍

src/css/_settings-modal.scss
jameskerr: Nice, my next ticket will fix this.

PR #767, opened 2020-05-11
ui space ids + zealot itests

### LINKED ISSUE #539: Write jest client tests
@mason-fish says:
> per @jameskerr's research, using https://jestjs.io/docs/en/puppeteer would give jest access to the fetch api's that the app uses, and would enable us to write client testing for both boom and zealot.

> Ideally we would only have one client lib, so this is a good preliminary step to take before we refactor/consolidate the two clients.

### LINKED ISSUE #705: brim support for space ids
Brim should use the information that will be created in https://github.com/brimsec/zq/issues/687 to create its list of spaces, and to obtain the space id's it needs to issue api requests to zqd. Going forward, Brim should use a space's immutable id, not its name, when internally referring to any state about the space.

mason-fish:
fixes #705 
begins to fix #539 though for now it only includes the testing infrastructure and tests for the new spaces api

Despite all the changes here, the ui should literally function and look exactly the same as it had before.

Other notes:
* there is now a new test target for the zealot integration tests: `npm run itest:zealot`. These are not hooked up to the `npm run test` target for now, though we may choose to remove the ignored directory that I added to that config in package.json
* The GET /spaces api now returns a detailed array of all space objects, instead of just the names. So the `setSpaces` action is populating more of the state than it had previously


src/js/services/zealot/pcapsApi.js
```diff
@@ -1,19 +1,19 @@
...
+export type PcapsGetArgs = {spaceID: string}
+export type PcapsPostArgs = {spaceID: string, path: string}
 
 export default {
-  get({space}: PcapsGetArgs) {
+  get({spaceID}: PcapsGetArgs) {
     return {
       method: "GET",
-      path: `/space/${encodeURIComponent(space)}/pcap`
+      path: `/space/${encodeURIComponent(spaceID)}/pcap`
```
mattnibs, line 10: ugh, we should do a migration at some point so we can get rid of this encodeURIComponent business.

src/js/services/zealot/itest/tests/test.js
```diff
@@ -0,0 +1,61 @@
+/* @flow */
```
jameskerr, line 1: I think renaming this to spaces.test.js would be nice.

src/js/state/Search/reducer.js
```diff
@@ -8,7 +8,8 @@ export const initSearchState: SearchState = {
   ],
   spanArgs: ["now - 5m", "now"],
   spanFocus: null,
-  space: "",
+  spaceID: "",
+  spaceName: "",
```
jameskerr, line 12: I don't think we need both the id and the name here in the search reducer.

src/js/state/Search/selectors.js
```diff
@@ -15,7 +15,8 @@ export default {
       program: SearchBar.getSearchBar(state).previous,
       pins: SearchBar.getSearchBar(state).pinned,
       spanArgs: Tab.getSpanArgs(state),
-      space: Tab.spaceName(state)
+      spaceName: Tab.spaceName(state),
+      spaceID: Tab.spaceID(state)
```
jameskerr, line 19: I know I'm to blame for straying from the convention (trying to make javascript like ruby), but are you willing to prefix any new selectors we make with "get". So here we'd have `Tab.getSpaceID()`?

src/js/state/Spaces/types.js
```diff
@@ -49,31 +50,31 @@ export type SPACES_DETAIL = {
 export type SPACES_INGEST_PROGRESS = {
   type: "SPACES_INGEST_PROGRESS",
   clusterId: string,
-  name: string,
+  spaceID: string,
```
jameskerr, line 53: In the other places in the app we have, like right above this line here `clusterId` with a lower-case 'd'. It's probably a nuisance to change, but I think I prefer `spaceId` the lower case 'd' since it conforms to the rule (rather than being an exception like JSON.parse) and matches other variables like 'clusterId'.  It also matches the common DOM api "document.getElementById". 

I personally would love everything lower-cased with underscores, but javascript... or rather...JavaScript 🤦🏼‍♂️

src/js/types/index.js
```diff
@@ -89,7 +89,8 @@ export type SearchRecord = {
   program: string,
   pins: string[],
   spanArgs: SpanArgs,
-  space: string
+  spaceID: string,
+  spaceName: string
```
jameskerr, line 93: As mentioned in a comment earlier (near the search reducer). I think it would be difficult to maintain if we store the space name here along with the ID since the name can change at anytime.

src/js/state/Search/reducer.js
mason-fish: Ah yeah, my first pass was to just include it everywhere but I suspect you're right that we don't need it in this. Will confirm and if we don't then will remove 👍

src/js/state/Search/selectors.js
mason-fish: Ah yes, I prefer that pattern as well. Will adjust

src/js/state/Spaces/types.js
mason-fish: Haha, sure thing. I guess this is me trying to make js more like go, where acronyms and abbreviations like ID are capitalized within camelcase. But I recognize that's not appropriate here. Will adjust.

src/js/state/Search/reducer.js
mason-fish: Ah so we *do* need this. Or at least things are currently wired up to use that name for the tab title. which displays the space name and search terms

jameskerr: I see, for their we can get the search.clusterId and the space.spaceId in the tab state to then look up spaces[clusterId][spaceId].spaceName in the root state.

If we keep that space name around, it will go out of date as soon as we allow for renaming the space. But if you want to revisit this when we add the space rename feature, I'm cool with that.

Issue #741, opened 2020-05-05
UI preference for zeek runner


alfred-landrum:
To support the bring-your-own-zeek effort, Brim will run zqd with a `-zeekrunner <path>` option, where the specified path should, by default, point to `zdeps/zeek/zeekrunner` (or zeekrunner.exe on windows).

We should have a UI preference that allows a user to override the path provided to the `-zeekrunner` argument. The preference should also provide a link to a wiki page (to be written) that documents how they can use this setting to customize how we generate zeek logs from pcaps.

philrz:
Verified in Brim commit `718af8a` talking to `zqd` commit `8301413`.

In the attached video, I've already made the `zeekrunner` selection in **Preferences** and restarted Brim for the setting to take effect. I can see via `ps` that my `zeekrunner` selection is reflected in the `zqd` command line. In this case all I'd done was copy the "stock" `zeekrunner` script to the base directory of a customized Zeek v3.1.3 install to which I'd added the JA3 plugin. After importing a pcap, I click on an `ssl` event and confirm that the additional JA3 fields are present.

[Verify.zip](https://github.com/brimsec/brim/files/4629317/Verify.zip)

While working on this verification, I did stumble across new bug #777.

Thanks @jameskerr!

PR #771, opened 2020-05-12
Zeek Runner User Preference

jameskerr:
The ZQD class will now look in the global preferences for a user specified zeek runner. This means that if a user changes the value, it will not be in effect until the app is restarted. Because of this, whenever the value changes, we show a message "Restart required."

image

Screen Shot 2020-05-12 at 4 27 25 PM

fixes: #741

### DIFF
--- a/src/css/_settings-modal.scss
+++ b/src/css/_settings-modal.scss
@@ -77,4 +77,16 @@
       @include label-small;
     }
   }
+
+  .feedback {
+    position: absolute;
+    right: 24px;
+    top: -11px;
+    @include label-small;
+    background: $blue;
+    color: white;
+    padding: 0 6px;
+    border-radius: 3px 3px 0 0;
+    animation: fadein 300ms;
+  }
 }
--- a/src/css/_settings-modal.scss
+++ b/src/css/_settings-modal.scss
@@ -13,7 +13,6 @@
     }
 
     select {
-      border: none;
       height: 24px;
       border: 1px solid rgba(0, 0, 0, 0.1);
       width: 240px;
@@ -21,6 +20,7 @@
 
     label {
       @include label-normal;
+      user-select: none;
     }
 
     input {
--- a/src/js/components/IngestWarningsModal.js
+++ b/src/js/components/IngestWarningsModal.js
@@ -3,7 +3,9 @@
 import {useSelector} from "react-redux"
 import React from "react"
 
+import {JSON_TYPE_CONFIG_DOCS} from "./Preferences/JSONTypeConfig"
 import {globalDispatch} from "../state/GlobalContext"
+import Link from "./common/Link"
 import ModalBox from "./ModalBox/ModalBox"
 import Spaces from "../state/Spaces"
 import Tab from "../state/Tab"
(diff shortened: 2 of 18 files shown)


src/css/_settings-modal.scss
```diff
@@ -13,14 +13,14 @@
     }
 
     select {
-      border: none;
```
jameskerr, line 16: @mason-fish border gone!

src/js/components/Preferences/Preferences.js
```diff
@@ -46,6 +47,7 @@ export default function Preferences() {
         <form ref={setForm} className="settings-form">
           <Timezone config={prefsForm.timeZone} />
           <TimeFormat config={prefsForm.timeFormat} />
+          <ZeekRunner config={prefsForm.zeekRunner} />
```
jameskerr, line 50: Almost a direct copy and paste from JSONTypeConfig. Potential for abstracting soon.

src/js/zqd/zqd.js
```diff
@@ -78,33 +79,29 @@ function zqdCommand(): string {
...
-  let zeekRunner = process.env.BRIM_ZEEK_RUNNER
-  if (!zeekRunner) {
-    // TODO: https://github.com/brimsec/brim/issues/741
-    // If the environment varible isn't set, allow the user to set
-    // a preferred zeekrunner location via preferences.
-    zeekRunner = resolve(join(zdepsDirectory, "zeek", plat.zeekRunnerBin))
-  }
-
-  if (!pathExistsSync(zeekRunner)) {
-    throw new Error("zeek runner not present at " + zeekRunner)
```
jameskerr, line 96: I removed this thow since the value is now user specified. If they provided invalid input, the app would not start.

Instead, for now, if there is a bad zeekrunner value set, the app starts up fine, but we display an error when they try to import a pcap.

Here is what the error will look like if the zeek runner does not exist:

Screen Shot 2020-05-12 at 4 11 34 PM

And here is if the zeek runner file is present but invalid

Screen Shot 2020-05-12 at 4 14 25 PM

alfred-landrum: Thanks for the screenshots - I can improve those error messages, filed:
https://github.com/brimsec/zq/issues/767

jameskerr:
# Update

Added doc links to the zeek runner and json type config wiki pages.

image

Added a doc link and an explanation in the ingest errors modal.

Screen Shot 2020-05-13 at 12 26 45 PM

src/js/components/Preferences/Preferences.js
jameskerr: Update: Did abstract this to a FileInput component used for the preferences modal.

src/js/components/IngestWarningsModal.js
```diff
@@ -31,7 +33,15 @@ export default function IngestWarningsModal() {
     >
       <TextContent>
         {warnings.length ? (
-          <pre className="output">{warnings.join("\n")}</pre>
+          <>
```
mason-fish, line 36: ah cool didn't know about fragments

src/js/components/common/Link.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+import React from "react"
+
+import {shell} from "electron"
+
+type Props = {|href: string, children: *|}
+
+export default function Link({href, children}: Props) {
```
mason-fish, line 8: Nice, a new common component! I had also included these in the common/index.js so that we can import/export the same way we do with the `lib` package.

src/js/zqd/zqd.js
```diff
@@ -78,33 +79,29 @@ function zqdCommand(): string {
...
-  }
+  const precedence = [
```
mason-fish, line 88: `s/precedence/preference` ?

jameskerr: I meant precedence here. This is the lookup order for the zeek runner path. First check the ENV, then the userPref, then the default.

PR #773, opened 2020-05-13
705 space ids review

jameskerr:
Fixes two bugs I found in my testing of the 705-space-ids branch.

### DIFF
--- a/src/js/components/SpacePicker.js
+++ b/src/js/components/SpacePicker.js
@@ -18,7 +18,7 @@ export default function SpacePicker() {
   let dispatch = useDispatch()
   let template = spaces.map(({id, name}) => ({
     label: name,
-    click: () => onSpaceChange(id)
+    click: () => onSpaceChange(id, name)
   }))
   if (template.length === 0) {
     template = [{label: "No spaces in this cluster", disabled: true}]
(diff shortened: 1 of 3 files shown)


src/js/components/SpacePicker.js
```diff
@@ -34,9 +34,9 @@ export default function SpacePicker() {
     dispatch(refreshSpaceNames())
   }
 
-  function onSpaceChange(val) {
-    setSpace(val)
-    setTimeout(() => dispatch(initSpace(val)))
+  function onSpaceChange(id, name) {
+    setSpace(name)
```
jameskerr, line 38: `setSpace(name)` is like an optimistic update to the UI before that space actually gets initialized below. Before it was setting that to the id. The result was a brief display of the id, then back to the name once the space was fully initialized.

src/js/flows/ingestFiles.js
```diff
@@ -115,7 +115,7 @@ const postFiles = (client, jsonTypesPath) => ({
       let types = isEmpty(jsonTypesPath)
         ? "default"
         : await lib.file(jsonTypesPath).read()
-      stream = client.logs.post({space: spaceId, paths, types})
```
jameskerr, line 118: I tried to upload some zeek logs and found that his api call hadn't been fully converted.

PR #784, opened 2020-05-15
feat: auto-update zq after successful testing (fixes #734)

### LINKED ISSUE #734: autoupdate brim to point to zq masters
This can happen once part one of #634 is finished.

mikesbrown:
When a zq PR is merged, update zq locally, run unit and integration tests, and commit the change if the tests pass. If the tests fail, save artifacts and open a PR. The PR text will contain links to the failing auto-update attempt. Opening a PR lets us piggy-back on alert infrastructure already in place, and provides a branch on which further Brim adjustments could be made.

Testing

The zq-triggering portion was verified earlier and continues to work. See for example [this run](https://github.com/brimsec/brim/actions/runs/105690954), triggered after the merge of brimsec/zq#783 .

One difficulty is that the handling of repository_dispatch events only runs on the default branch, which is master in most cases. To test, I copied the JSON from events into files I temporarily checked in to a Git branch. [This is an example of a passing run](https://github.com/brimsec/brim/runs/665156542). [This is an example of a failing run](https://github.com/brimsec/brim/actions/runs/101955289). The PR opened after the fail is #768.

This has one known shortcoming, in that if a different brim PR is merged while this is running, this job will fail, because a fast-forward is no longer possible. I'd rather get this in, see how often that happens, and adjust as needed.

### DIFF
--- a/.github/workflows/advance-zq.yml
+++ b/.github/workflows/advance-zq.yml
@@ -14,9 +14,14 @@ jobs:
       matrix:
         node-version: [12.x]
         os: [ubuntu-18.04]
+    timeout-minutes: 30
     steps:
       # Only one of these should run at a time, and the checkout of brim
-      # has to be in the "protected section".
+      # has to be in the "protected section". This will poll every 60s
+      # forever. It will be timed out based on any change to
+      # https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes
+      # It is not possible to time out this step and fail. It's only
+      # possible to time out this step and continue.
       - name: Turnstyle
         uses: softprops/turnstyle@v0.1.2
         env:
@@ -94,7 +99,6 @@ jobs:
 
       - run: npm install --no-audit
       - run: npm run build
-      # Unit tests touch zq-produced *.js
       - run: npm test -- --maxWorkers=2 --ci
       - run: xvfb-run --auto-servernum -s "-screen 0 1280x1024x24" npm run itest -- --ci --forceExit
         env:
--- a/.github/workflows/advance-zq.yml
+++ b/.github/workflows/advance-zq.yml
@@ -1,12 +1,143 @@
 name: Advance ZQ
 # This type must match the event type from zq.
 # https://help.github.com/en/actions/reference/events-that-trigger-workflows#external-events-repository_dispatch
+# These events only trigger on the Github default branch (usually
+# master).
 on:
   repository_dispatch:
     types: [zq-pr-merged]
 jobs:
   advance-zq:
     name: Advance ZQ
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node-version: [12.x]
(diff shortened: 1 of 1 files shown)


mikesbrown:
Once again thanks to go @henridf for providing an earlier version of similar work on which I could build.

.github/workflows/advance-zq.yml
```diff
@@ -1,12 +1,143 @@
...
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-${{ matrix.node-version }}-
+
+      - name: update zq
+        run: npm install https://github.com/brimsec/zq#${{ steps.zq_pr.outputs.sha }}
+
+      - run: npm install --no-audit
+      - run: npm run build
+      # Unit tests touch zq-produced *.js
```
alfred-landrum, line 97: I'm not sure what this means?

.github/workflows/advance-zq.yml
```diff
@@ -1,12 +1,143 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node-version: [12.x]
+        os: [ubuntu-18.04]
     steps:
+      # Only one of these should run at a time, and the checkout of brim
+      # has to be in the "protected section".
+      - name: Turnstyle
+        uses: softprops/turnstyle@v0.1.2
```
alfred-landrum, line 21: I couldn't tell from the github site for this action, is there a default timeout value?

.github/workflows/advance-zq.yml
```diff
@@ -1,12 +1,143 @@
...
+
+      # This section gets event information.
       - run: jq '.' "${GITHUB_EVENT_PATH}"
+      - name: process pull request event
+        id: zq_pr
+        # $GITHUB_EVENT_PATH is the JSON we posted from zq.
+        # Variables for other steps get set as described here:
+        # https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter
+        # body can be multiline and must be escaped as described here:
+        # https://github.community/t5/GitHub-Actions/set-output-Truncates-Multiline-Strings/m-p/37870
```
alfred-landrum, line 43: 👍

.github/workflows/advance-zq.yml
mikesbrown: removed

.github/workflows/advance-zq.yml
mikesbrown: Sort of. The step allows "timeout and continue". Timeout and fail is configurable via https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes . While I gave this action some basic testing, it's true it's not rigorously been tested in our environment. To avoid the chance of a big queue building up due to some bug, I've set `timeout-minutes` to 30 from its default of 360.

mikesbrown:
Thanks for the review @alfred-landrum !

PR #787, opened 2020-05-16
Migration session data

### LINKED ISSUE #751: App state migration
In #587 we introduced versioning of the app state object. A comment from @jameskerr on follow-on work we'd now like to take care of:

> Depending on how critical it is for users to save state, like their history and in the future, their saved queries. We may want to write more code to migrate data and what not.

Indeed, we've warned users through two upgrades that they're going to lose their History, and no doubt they'd appreciate being able to start relying on this data going forward. So let's make migration a thing.

Another comment from there that might be worth taking care of at the same time, if it isn't already:

> I think it would be good to write a test that does some of the common operations on the redux store, the saves the resulting state in a snapshot. Then each time we make changes to the state shape, hopefully that test will fail with a message like "update the state version".

jameskerr:
fixes #751 

Each commit here has good comments so I won't repeat them here. But basically, this PR is the meat of the feature.

The `migrations` class will run all pending migrations.

The `session` became a singleton and automatically runs the migrations when it loads previous data.

Finally, I used the new framework to do our first migration which adds a default value for our new ZeekRunner preference (`state.prefs.zeekRunner = ""`).

**Future PRs:**

I did a diff between master and our previous release v0.9.1 looking for changes to the state. Here's what changed. I'll open another PR that migrates the state to handle these state changes.

```diff
src/js/state/Prefs/types.js

-export type PrefsState = {
-  jsonTypeConfig: string
-}
+export type PrefsState = {|
+  jsonTypeConfig: string,
+  timeFormat: string,
+  zeekRunner: string
+|}
```

```diff
src/js/state/Search/types.js

@@ -7,7 +7,8 @@ export type SearchState = {
   span: Span,
   spanArgs: SpanArgs,
   spanFocus: ?Span,
-  space: string,
+  spaceName: string,
+  spaceId: string,
   clusterId: string,
   ts: number
 }
```

```diff
src/js/state/Spaces/types.js

 export type SpacesState = {
   // clusterId
   [string]: {
-    // spaceName
+    // spaceId
     [string]: Space
   }
 }

 export type Space = {
   name: string,
+  id: string,
   min_time: Ts,
   max_time: Ts,
-  packet_support: boolean,
+  pcap_support: boolean,
   ingest: SpaceIngest
 }
```

**Final thought**

I think that we should not save so much stuff in the state. For example, all the records from the viewer (500 zeek records) should not be persisted. We can easily re-run the query to populate those again.


src/js/electron/main.js
```diff
@@ -55,21 +54,26 @@ async function main() {
     }
   }
 
-  app.on("ready", () => {
+  function onReady() {
     installExtensions()
```
mason-fish, line 58: Not exactly related, but I noticed we call this but the function is all commented out. Any reason why we don't use those tools in dev? (the Redux devtools are particularly useful)

src/js/electron/main.js
```diff
@@ -55,21 +54,26 @@ async function main() {
...
     } else {
-      winMan.init(sessionState)
+      winMan.init(data)
     }
-  })
+  }
+
+  // The app might be ready by the time we get here due to async stuff above
+  if (app.isReady()) onReady()
+  else app.on("ready", onReady)
```
mason-fish, line 68: why do we need more than the event listener (`app.on("ready",...`) here?

src/js/electron/main.js
jameskerr: Oh right. I think the redux dev tools was getting really slow. I'm open to trying them out again though.

src/js/electron/main.js
jameskerr: See the comment right above it there. The async stuff above (loading and migrating the state stuff from the file) was taking longer than it took for `app` to be ready. So here we check if it's ready, then call the callback immediately, else setup the callback.

jameskerr:
That's a good question. I think if there is an error thrown in the migrations, the app won't start. Maybe that's a good idea for dev, then in production we just give them a blank slate?

PR #794, opened 2020-05-18
User dataDir preference

### LINKED ISSUE #676: Provide preference for default space directory
With https://github.com/brimsec/brim/issues/673 , we're going to default to creating space directories in `<userData>/data/spaces`. We can add a Brim preference that will allow a user to specify the default data directory location for space creation. After this preference is set, any spaces created should use a `data_dir` parameter that includes the specified location.

mason-fish:
fixes #676 

Turns out html `input` forms **really** prefer files over directories. There is apparently a `webkitdirectory` option which I experimented with first. The result was that the selector dialog would behave correctly but the `onChange` handler was no longer being called internally inside react. That property also is not recommended in the [docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory) (though I believe the primary reason for this is browser compatibility, which we are in control of in electron).

To fix this then, I added an ipc call so that the dialog that appears comes from electron and the main process. What's nice about this is that it will actually not allow users to select anything BUT a directory. Still, the form has validation in case someone manually types in a bad path.

datadirpref

### DIFF
--- a/src/js/components/Preferences/FileInput.js
+++ b/src/js/components/Preferences/FileInput.js
@@ -5,18 +5,36 @@ import classNames from "classnames"
 import ToolbarButton from "../ToolbarButton"
 import useCallbackRef from "../hooks/useCallbackRef"
 import useDropzone from "../hooks/useDropzone"
+import invoke from "../../electron/ipc/invoke"
+import ipc from "../../electron/ipc"
+import {isEmpty} from "../../lib/Array"
 
 type Props = {|
   defaultValue: string,
   name: string,
   placeholder?: string,
-  onChange?: Function
+  onChange?: Function,
+  isDirInput?: boolean
 |}
 
 export default function FileInput(props: Props) {
   let [picker, ref] = useCallbackRef()
   let [bindDropzone, dragging] = useDropzone(onDrop)
   let [value, setValue] = useState(props.defaultValue)
+
+  function onClick() {
+    const {isDirInput} = props
+    isDirInput
+      ? invoke(ipc.windows.openDirectorySelect()).then(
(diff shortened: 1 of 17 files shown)


src/js/lib/file.js
```diff
@@ -18,21 +18,23 @@ export default function file(p: string) {
...
+                Promise.all(files.map((f) => file(path.join(p, f)).allFiles()))
+              )
+              .then((results) =>
+                results.reduce<string[]>((all, one) => all.concat(one), [])
+              )
+          : [p]
       })
     },
 
+    isDirectory() {
```
jameskerr, line 34: 👍

src/js/components/Preferences/Preferences.js
```diff
@@ -25,6 +26,7 @@ export default function Preferences() {
       let form = brim.form(f, prefsForm)
 
       if (await form.isValid()) {
+        setErrors([])
```
jameskerr, line 29: Makes sense.

src/js/components/Preferences/FileInput.js
```diff
@@ -5,19 +5,37 @@ import classNames from "classnames"
...
 
 export default function FileInput(props: Props) {
   let [picker, ref] = useCallbackRef()
   let [bindDropzone, dragging] = useDropzone(onDrop)
   let [value, setValue] = useState(props.defaultValue)
 
+  function onClick() {
+    const {isDirInput} = props
+    isDirInput
+      ? invoke(ipc.windows.openDirectorySelect()).then(
```
jameskerr, line 28: 👍

PR #796, opened 2020-05-19
Ingest test improvements

mikesbrown:
This change refactors and improves ingest tests. Ingest tests are pulled out of the smoke test, and the old initial smoke test is retired. The ingest test code is simplified, and new integration tests for streaming JSON and pcapng are introduced. sample.ndjson has been regenerated from Zeek, not zq (an earlier decision of convenience). Application start/stop churn is reduced. Before ingesting a new file, a new tab is opened. This is intended to cut down instances of #785.

PR #802, opened 2020-05-20
Export Results to ZNG

### LINKED ISSUE #677: UI for zng file export
**Request**
As a User I want to be able to export the results of a query so that I can share or move the data from my analysis.

**Description**
Other tools like Wireshark or Caploader allow export or a selection of packets to disk.  In our case we want to provide similar functionality that will enable future transfer of information between desktop and server deployments.

This functionality will also allow cross-platform re-use of data, ZNG data, for example, linux to Windows.

Detailed requests:

1. Provide an export icon as part of the existing toolbar. This is a user friendly option.
2. Provide a menu option to export the query.
3. Provide a screen where the User can specify the filename and location to export.

Future functionality will allow Log tagging. Tagged logs could then be exported  to ZNG. Re-use of design is desired.

User flow:

1. The user performs an analysis that renders a query.
2. The user clicks the export icon (or query export option in menu) to export the results to disk.
3. The user specifies the location and name of the file. The type is predefined as ZNG.

Figure: Export icon as part of existing toolbar

Figure: Query section export option

Reference information, Wireshark export functionality.

1. Select packets

2. Export

3. Specify export parameters

Reference, Caploader implementation:

jameskerr:
closes #677 

Allows a user to export the current list of results to a zng file. This can be done from the App Menu via `File => Export Results to ZNG...` or from the search menu toolbar. 

image

image

sVxaQIfVW4


siskojr:
Looks great. Also the icon. Thanks!

package.json
```diff
@@ -150,6 +150,7 @@
     "reselect": "^4.0.0",
     "rimraf": "^3.0.2",
     "source-map-support": "^0.5.16",
+    "styled-components": "^5.1.0",
```
jameskerr, line 154: 💅🏼

I think this is such a handy, well-made library, let's just start using it.

jameskerr: https://styled-components.com/

src/js/flows/exportResults.js
```diff
@@ -1,9 +1,46 @@
...
+    data.pipe(file)
+  })
+}
 
-export default (filePath: string): Thunk => (dispatch, getState) => {
-  let args = Search.getArgs(getState())
-  console.log(args)
-  console.log("exporting", filePath)
+function toNodeReadable(reader) {
+  return new stream.Readable({
```
jameskerr, line 39: https://nodejs.org/api/stream.html#stream_new_stream_readable_options

jameskerr: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream

src/js/components/hooks/useSetTimeout.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+import {useEffect, useRef} from "react"
+
+export default function useSetTimeout() {
```
jameskerr, line 4: This is for when I set a timeout in a click handler. If the component is unmounted, the timeout will still fire and might ask the component to "re-render". But since the component is unmounted, react throws this error:

```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method
```

src/js/components/ExportButton.js
```diff
@@ -0,0 +1,80 @@
...
+    properties: ["createDirectory"]
+  })
+}
+
+export default function ExportButton() {
+  let dispatch = useDispatch()
+  let [status, setStatus] = useState("INIT")
+  let setTimeout = useSetTimeout()
+
+  useIpcListener("exportResults", onClick, [])
```
jameskerr, line 52: It seems more maintainable to put the ipc listeners in the components that care about them, rather than in the initializer `initIpcListeners.js` file.

src/js/components/ColumnChooser.js
```diff
@@ -13,24 +12,15 @@ type Props = {}
 export default class ColumnChooser extends React.Component<Props> {
   render() {
     return (
-      <div className="columns-button">
+      <div className="columns-button" title="Show or hide columns in the table">
```
jameskerr, line 15: In the spirit of "as native as possible", I think we should use the standard html title attributes. They look great, they always get positioned properly, and it appears they have a global timer for all of them, making the user not have to wait the full delay if they jump from one button to another.

Screen Shot 2020-05-20 at 5 18 34 PM

src/js/components/ExportButton.js
```diff
@@ -72,7 +72,11 @@ export default function ExportButton() {
 
   return (
     <div title="Export search results to ZNG file">
-      <ToolbarButton icon={<ExportIcon />} onClick={onClick} />
+      <ToolbarButton
+        icon={<ExportIcon />}
+        onClick={onClick}
+        data-test-locator="export-results"
```
jameskerr, line 78: @mikesbrown what do you think about simply having a string like this in the data attribute in react. Then in the spectron test we do:

```js
// in the lib file
export const locator = (name: string) =>`[data-test-locator]="${name}"`

// in the test
await click(app, locator("export-results"))

```

mikesbrown: Was this because you find the `reactElementProps()` and `integration.js` dance tedious? I understand. Let me outline some reasonings for why that's used:

1. We can use static analysis to find places where we make mistakes. Two examples:
   1. Such attribute names need to be unique in the product, and a convenient way to ensure uniqueness is to make them all keys in the same object so that `eslint` can warn you if there are duplicates.
   1. Mistyping `data-test-locator` as a string literal is more prone to errors and typos and won't be caught in linters, flow, etc., whereas calling `reactElementProps()` is going to have more chance for a mistake to be caught in static analysis. For example, if you mistype `reactElemntProps()` then you'll get an import error from `eslint`. If you typo the string literal `data-tets-locator` on the other hand, the app will happily pass all checks, and you'll wonder why your integration test is hanging trying to find the element.
1. Ease-of-maintenance: Tests shouldn't directly interact with locators. Locators are abstracted away so that if they move, change names, etc., fewer test changes have to be made. For example, you'd likely only need change `src/js/test/integration.js` and possibly `itest/lib/app.js`.

With these considerations, I'm happy to have a conversation about how to rework this to try to make this less tedious but also allow errors to be caught earlier rather than later. Let me know what you think.

itest/tests/exportResults.test.js
```diff
@@ -0,0 +1,57 @@
...
+    dialog.mock([
+      {
+        method: "showSaveDialog",
+        value: {
+          canceled: false,
+          filePath
+        }
+      }
+    ])
+    await writeSearch(app, "")
```
mikesbrown, line 46: It would be good to have a few tests that actually perform zql searches and verify the results come from that search, and not something else or the whole space. There are a few ways to do this and some considerations. First, the zql should be deterministic: you should apply some sort of `sort` so that the data always comes back in the same order. Second, the output zng file should be inspected. With pcaps, I did this using md5 checksums, after first locally verifying the pcap contained the correct output. Or you could do something more complex like call `zq`.

itest/tests/exportResults.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+describe("test exporting results", () => {
+  let app
+  let testIdx = 0
+
+  beforeAll(async () => {
+    app = newAppInstance(path.basename(__filename), ++testIdx)
+    dialog.apply(app)
+    await startApp(app)
+    await pcapIngestSample(app)
```
mikesbrown, line 26: FYI I'm trying to get rid of `pcapIngestSample()` in favor of `ingestFile()` everywhere as part of #796. If you could use `ingestFile()` here that would be helpful.

itest/tests/exportResults.test.js
```diff
@@ -0,0 +1,57 @@
...
+} from "../lib/app"
+import {locator} from "../../src/js/test/integration"
+import {retryUntil} from "../lib/control"
+
+describe("test exporting results", () => {
+  let app
+  let testIdx = 0
+
+  beforeAll(async () => {
+    app = newAppInstance(path.basename(__filename), ++testIdx)
```
mikesbrown, line 23:
```suggestion
    await fsExtra.remove("tmp")
    app = newAppInstance(path.basename(__filename), ++testIdx)
```
It's good habit to do "cleaning up" as part of "set up".

itest/tests/exportResults.test.js
```diff
@@ -0,0 +1,57 @@
...
+    let filePath = path.normalize("tmp/test-export-results.zng")
+    dialog.mock([
+      {
+        method: "showSaveDialog",
+        value: {
+          canceled: false,
+          filePath
+        }
+      }
+    ])
```
mikesbrown, line 45: Neat!

package-lock.json
```diff
@@ -2258,7 +2267,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
```
mikesbrown, line 2270: Here and elsehwere in many places, there is a https -> http change to package-lock.json. Could you please change these back to https?

itest/tests/exportResults.test.js
mikesbrown: Now that #796 is in, if you rebase on master, you'll need to use `ingestFile()`.

itest/tests/exportResults.test.js
```diff
@@ -0,0 +1,57 @@
...
+    await pcapIngestSample(app)
+    await fsExtra.ensureDir("tmp")
+  })
+
+  afterAll(async () => {
+    if (app && app.isRunning()) await app.stop()
+    await fsExtra.remove("tmp")
+  })
+
+  test("clicking the button", async () => {
```
mikesbrown, line 35: I missed this before. This should be a `stdTest` so that we get logging when the test starts, and it should call the standard handler if there's an exception.

src/js/components/ColumnChooser.js
mason-fish: Agreed. Simple and functional is very good.

src/js/components/ExportButton.js
mason-fish: Makes sense to me

src/js/components/ExportButton.js
mikesbrown: This got brought up again in #822

PR #806, opened 2020-05-21
add rename right click

### LINKED ISSUE #674: brim: UI to rename a space
**Request**
As a user I want to change the space name so that I can have a naming structure similar to my preferences of file name organization.

**Description**
If a user ingests files in the following sequence:
…/day1/errors.pcap
…/day2/errors.pcap
the spaces that are created will have the following names:
errors.pcap.brim
errors.pcap.brim_01

The users may prefer to change the name to something more meaningful for them.

Provide the rename functionality with a right-click action in the Recent spaces pane.

Desired: provide also the rename action directly in the tab name. Either right-click and rename or hit return key and allow rename like macOS functionality of file.

mason-fish:
fixes #674 

Uses a simple modal for the rename mechanism, after right clicking on a space.

brim

### DIFF
--- a/src/js/components/SpaceModal.js
+++ b/src/js/components/SpaceModal.js
@@ -1,0 +1,52 @@
+/* @flow */
+import React, {useState} from "react"
+
+import ModalBox from "./ModalBox/ModalBox"
+import {useDispatch, useSelector} from "react-redux"
+import {reactElementProps} from "../test/integration"
+import Modal from "../state/Modal"
+import Spaces from "../state/Spaces"
+import {Input, InputSubmit} from "./form/Inputs"
+import Form from "./form/Form"
+import renameSpace from "../flows/renameSpace"
+
+export default function SpaceModal() {
+  return (
+    <ModalBox
+      name="space"
+      title=""
+      buttons={[]}
+      {...reactElementProps("spaceModal")}
+    >
+      <SpaceModalContents />
+    </ModalBox>
+  )
+}
+
+const SpaceModalContents = () => {
+  const {spaceId, clusterId} = useSelector(Modal.getArgs)
+  const space = useSelector(Spaces.get(clusterId, spaceId))
+  const {name} = space || ""
+  const [nameInput, setNameInput] = useState(name)
+  const dispatch = useDispatch()
+
+  const onSubmit = () => {
+    dispatch(renameSpace(clusterId, spaceId, nameInput))
+    dispatch(Modal.hide())
+  }
+
+  return (
+    <div>
+      <Form onSubmit={onSubmit}>
+        <Input
+          label="Space Name"
+          value={nameInput}
+          onChange={(e) => {
+            setNameInput(e.target.value)
+          }}
+        />
+        <InputSubmit value={"Rename"} />
+      </Form>
+    </div>
+  )
+}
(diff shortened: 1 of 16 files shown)


siskojr:
Looks great. Thanks.

src/js/components/TabWelcome.js
```diff
@@ -15,12 +15,14 @@ import errors from "../errors"
...
 import initNewTab from "../flows/initNewTab"
 import refreshSpaceNames from "../flows/refreshSpaceNames"
+import menu from "../electron/menu"
 
 export default function TabWelcome() {
   let dispatch = useDispatch()
   let id = useSelector(Tab.clusterId)
   let spaces = useSelector(Spaces.getSpaces(id))
   let spacesPresent = spaces.length !== 0
+  const spaceContextMenu = menu.spaceContextMenu(id)
```
jameskerr, line 25: I know this might be silly, but do you have any objection to using `let` for most variables like this? My reasons are simply that it's two less characters and makes code a little easier to read in my opinion. `const` will throw an error if it's re-assigned, which is a safeguard, but I've not had much benefit from that safety in my javascript time. Maybe I'm being superficial and if you feel strongly about `const` we can come to some consensus about where and when to use each.

src/js/components/SpaceModal.js
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <SpaceModalContents />
+    </ModalBox>
+  )
+}
+
+const SpaceModalContents = () => {
+  const {spaceId, clusterId} = useSelector(Modal.getArgs)
+  const space = useSelector(Spaces.get(clusterId, spaceId))
+  const {name} = space || ""
```
jameskerr, line 29: An undefined space will cause an error here because it will try to destructure the empty string.

You might want `const {name} = space || {name: ""}`

src/js/flows/renameSpace.js
```diff
@@ -0,0 +1,23 @@
...
+  dispatch,
+  getState
+) => {
+  const state = getState()
+  const zClient = Cluster.getZealot(clusterId)(state)
+  const tabs = Tabs.getData(state)
+
+  zClient.spaces.update(spaceId, {name}).then(() => {
+    dispatch(Spaces.rename(clusterId, spaceId, name))
+    tabs.forEach((t) => {
```
jameskerr, line 18: Maybe this can be removed one day in the future? And we can refactor Search to only need the space id?

src/js/state/Clusters/selectors.js
```diff
@@ -1,9 +1,15 @@
 /* @flow */
 
 import type {State} from "../types"
+import zealot from "../../services/zealot"
 
 export default {
   id: (id: string) => (state: State) => state.clusters[id],
+  getZealot: (id: string) => (state: State) => {
```
jameskerr, line 8: Something like this already exists in the `Tab/selectors.js`

src/js/components/TabWelcome.js
mason-fish: After chatting offline, we've decided to go with `const` for all things unless you need to use `let` for reassignment.

src/js/state/Clusters/selectors.js
mason-fish: Switched to use that one instead 👍, thanks!

PR #821, opened 2020-05-26
migrate investigation spaceId

### LINKED ISSUE #813: Clicking History entries in upgraded app causes "space does not exist" errors
In the attached video, I start out running GA Brim tagged `v0.9.1`. I perform a couple queries to populate my History panel. Then I close the app and upgrade to a release artifact based on Brim commit `dc488b1`. My History entries now show the yellow exclamation point warning, and when I click on any of them, I get a "space does not exist" error.

[newbug.mp4.zip](https://github.com/brimsec/brim/files/4683170/newbug.mp4.zip)

I sniffed around in Wireshark to see if it's likely a frontend or backend bug. Here's the initial exchange in `v0.9.1`:

Then in the newer version, I see a lot more info was returned in the initial `GET /space`, but when the Search is issued, there's no `space` field like there was in `v0.9.1`

mason-fish:
fixes #813 

There was still another piece of state migration left after the spaceId work. In this case we needed to update the `investigation` branch so that appState history could remain in sync with the id refactor.


philrz:
At @mason-fish's request, I tested out a release artifact based on this branch. By following the repro steps in #813 I confirmed that the issue no longer occurs after upgrade from `v0.9.1` to that new artifact. I can't speak to the code, so leaving that for other reviews. But I'm 👍 from a functional perspective.

mason-fish:
@jameskerr nice catch with the entries in `History`. I pushed another migration to fix that. I couldn't find a way to trigger any ui bugs from that, so no pics to share, but the tests capture the coverage.

src/js/state/migrations/202005271240_setHistorySpaceToNameAndId.test.js
```diff
@@ -0,0 +1,31 @@
...
+import migrate from "./202005271240_setHistorySpaceToNameAndId"
+import {cloneDeep} from "lodash"
+
+const getTabHistoryEntries = (state) => {
+  return state.tabs.data.map((t) => t.history.entries)
+}
+
+test("migrating 202005271240_setHistorySpaceToNameAndId", () => {
+  const prev = getTestState("v0.9.1")
+  const next = migrate(cloneDeep(prev))
```
mason-fish, line 13: @jameskerr side note, how do you feel about making `migrate()` more "pure" by always returning a new copy? I personally prefer that sort of functional paradigm which, as an example, would make our tests more idiomatic since we often will compare prev to next.

jameskerr: In general, I like as little mutation as possible. However, I'd like to be able to mutate the object in the migration file. I think (in javascript), mutative code is easier to read and understand. So if we were to implement that, I think doing something like you're doing here (cloning it right before you pass it to the migrate function) is ideal. 

Related to the need to use deepClone, I'll share something I've heard about writing tests. They said it's helpful if tests don't repeat the implementation of the code they test. Instead, they assert against hardcoded, expected values. This makes them more likely to catch errors, and easier to read, and won't change if the undelying code changes. There may have been other good reason's but I can't think of them.

FWIW, your test here is totally fine with me, but I'll throw out this alternative.

```js
  const prev = getTestState("v0.9.1")
  const next = migrate(prev)

  let windows = Object.values(next.windows)
  let tab1 = getTabHistoryEntries(windows[0].state)[0]
  let tab2 = getTabHistoryEntries(windows[0].state)[1]

  expect(tab1.every((e) => e.spaceId === "corelight.pcap.brim")).toBe(true)
  expect(tab1.every((e) => e.spaceName === "corelight.pcap.brim")).toBe(true)

  expect(tab2.every((e) => e.spaceId === "pcaps.brim")).toBe(true)
  expect(tab2.every((e) => e.spaceName === "pcaps.brim")).toBe(true)
```

I see that it's a little more difficult here because there's an array of windows and for each of those and array of tabs to test. Anyway, in the case of testing against hardcoded values, you wouldn't need to use the previous value in the assertions.

Anyway, this PR looks good. I just wanted to share that thought.

PR #822, opened 2020-05-27
JSON Type Config Preferences iTest

### LINKED ISSUE #788: json ingest integration test
Write an integration test that verified the json ingest: so setting a non-default location in the preference UI, doing an ingest, and verifying that the resulting data shows up with the correct types & data.

jameskerr:
An integration test for selecting a custom json type config, then ingesting some json logs with that custom schema.

Thanks @mikesbrown for the test data.

fixes  #788


itest/tests/preferences.test.js
```diff
@@ -0,0 +1,28 @@
...
+  stdTest("setting a custom json type config", async (done) => {
+    brim.clickAppMenuItem("preferences")
+    await brim.setValue(jsonTypeConfigInput, config)
+    await brim.click(defaultModalButton)
+    await brim.ingest("custom-sample.ndjson")
+    await brim.waitUntil(() => brim.isNotVisible(ingestProgressBar))
+
+    expect(await brim.getText(viewerResults)).toMatchSnapshot()
+    done()
+  })
```
jameskerr, line 27: This function is the focus point.

src/js/electron/itest.js
```diff
@@ -1,10 +1,17 @@
...
   ipcMain.removeHandler(name)
   ipcMain.handle(name, () => value)
 })
+
+global.process.on("spectron:clickAppMenuItem", (id) => {
+  log.info("Specton is clicking app menu item id: ", id)
+  let menu = Menu.getApplicationMenu()
+  let item = menu.getMenuItemById(id)
+  item.click()
+})
```
jameskerr, line 17: This is a way we can click on items in the app menu from spectron.

src/js/test/locators.js
```diff
@@ -0,0 +1,8 @@
+/* @flow */
+import createLocator from "./createLocator"
+
+export const jsonTypeConfigInput = createLocator("json_type_config_input")
+export const defaultModalButton = createLocator("default_modal_button")
+export const viewerResults = createLocator("viewer_results")
+export const ingestProgressBar = createLocator("ingest_progress_bar")
+export const toolbarExportButton = createLocator("toolbar_export_button")
```
jameskerr, line 8: @mikesbrown here is another proposal for a new way to do integration test "locators". 

```js
// locators.js
export const myButton = createLocator("my_button")

// in react
<SomeButton {...myButton.props} />

// in an integration test
app.client.click(myButton.css)
```

What do you think of this pattern? I think it accounts for some of those good points you made in my previous PR, and has less friction that our current pattern.

itest/lib/createTestBrim.js
```diff
@@ -36,6 +38,25 @@ export default (name: string) => {
       )
     },
 
+    setValue(locator: Locator, value: string) {
+      return appStep(`set input ${locator.css} to ${value}`, () =>
+        app.client
+          .waitForVisible(locator.css)
+          .then(() => app.client.setValue(locator.css, value))
+      )
+    },
```
mikesbrown, line 47: This is an example of a great function for a test to use because the wait-until-I-can-do-something-then-do-it dance is handled here, not requiring the test to do anything.

itest/lib/createTestBrim.js
```diff
@@ -36,6 +38,25 @@ export default (name: string) => {
...
+    getText(locator: Locator) {
+      return appStep(`get text from ${locator.css}`, async () => {
+        await app.client.waitForVisible(locator.css)
+        return await app.client.getText(locator.css)
+      })
+    },
+
+    clickAppMenuItem(id: string) {
+      return app.mainProcess.emit("spectron:clickAppMenuItem", id)
+    },
```
mikesbrown, line 58: Contrast this with `setValue()` above. Here, this operation is asynchronous. A test is going to have to do its own work to ensure the click has actually performed the expected action before the test moves onto the next action. We just need to be clear and set expectations about what these do. I greatly prefer that the layer used by tests is responsible for ensuring the element is ready to be acted on. I understand in some cases that's hard.

itest/tests/preferences.test.js
```diff
@@ -0,0 +1,28 @@
...
+import {
+  defaultModalButton,
+  ingestProgressBar,
+  jsonTypeConfigInput,
+  viewerResults
+} from "../../src/js/test/locators"
+import {stdTest} from "../lib/jest"
+import createTestBrim from "../lib/createTestBrim"
+
+const config = path.join(__dirname, "../testdata/custom-schema.json")
```
mikesbrown, line 13:
```suggestion
import testDataDir from "../lib/env"

const config = path.join(testDataDir(), "custom-schema.json")
```

itest/tests/preferences.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+describe("Preferences Modal Tests", () => {
+  const brim = createTestBrim("preferences.test")
+
+  stdTest("setting a custom json type config", async (done) => {
+    brim.clickAppMenuItem("preferences")
+    await brim.setValue(jsonTypeConfigInput, config)
+    await brim.click(defaultModalButton)
+    await brim.ingest("custom-sample.ndjson")
+    await brim.waitUntil(() => brim.isNotVisible(ingestProgressBar))
```
mikesbrown, line 23: In general, tests shouldn't be asked to do waiting. That is better served with application steps. This particular pattern is already done with `ingestFile()`. You could add another method to createTestBrim for that, right?

itest/tests/preferences.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+    expect(await brim.getText(viewerResults)).toMatchSnapshot()
```
mikesbrown, line 25: In general, bare, zql-less log results are prone to giving nondeterminstic results. That could be why this test failed on Windows. Can I suggest what I used in #817? It's very equivalent to the search used in the ingest tests:
```
_path=conn proto=tcp | cut ts, src_ip, src_port, dst_ip, dst_port, proto | sort ts
```
Using this gets more bang for your buck as a test: not only do you get (or are supposed to get) determinstic results, but also you exercise the custom schema (note the use of `src_ip` et al, which aren't in normal Zeek logs).

src/js/test/locators.js
jameskerr: I think it's nice to split them up because one will rarely need to look at the createLocator function (after looking at it once). But I'll probably want to reference the locators more often. In general, it's helpful for me if a file only contains code that's relevant to my task at hand. I'm in favor of any structure that lowers the amount of focus I need to have when I open up a file. Hence, splitting them up.

src/js/components/StatusBar.js
```diff
@@ -42,7 +43,9 @@ export default function StatusBar() {
       >
         <label>{getMessage()}</label>
         <div className="group">
-          {s.ingesting() && <ProgressIndicator percent={value} />}
+          {s.ingesting() && (
+            <ProgressIndicator {...ingestProgressBar.props} percent={value} />
+          )}
```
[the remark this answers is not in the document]
jameskerr: I looks like the progress bar always disappears when ingest is done, but it seems that the parent div (with the existing selector) might not? Like in the case of there being an ingest error.

itest/lib/createTestBrim.js
jameskerr: Aww yes, I see what you mean. Yeah, it's nice to have that type of code in here.

itest/lib/createTestBrim.js
jameskerr: Yeah, the test will have to ensure that the expected thing takes place after clicking the application menu.

However, the next line of the test will usually do the waiting I think. For example:
```js
brim.clickAppMenuItem("preferences")
brim.getText(preferencesModal) // in here it could wait for the preferencesModal to appear
```

itest/tests/preferences.test.js
jameskerr: Oh, ingest file already handles waiting for the ingest to finish. `brim.ingest` internally uses `ingestFile`, so this line is not needed.

itest/tests/preferences.test.js
jameskerr: Yes! Thanks for that.

jameskerr:
@mikesbrown I appreciate your comments. I agree that the waters have muddied with my introducing the createTestBrim function. I hesitated adding it, but decided to go for it and at least have something tangible to discuss. I'd love to talk live about the best way forward.

I personally think the app step functions are still really nice. Maybe createTestBrim is just a layer on top of those. My hope would be that createTestBrim doesn't get too big since I'll most likely want to refer to it often to see what methods are available.

But anyway, I'm looking forward to discussing. Writing integration tests is kinda fun :)

src/js/components/StatusBar.js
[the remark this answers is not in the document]
mikesbrown: I looked into this. `ingestProgress` is fine for when we expect ingest to succeed. It might even be preferable, because if there is an unexpected error, the test will fail at that point (`ingestProgress` never disappears, even though we are waiting for it to do so). Test authors don't have to worry about "checking that ingest succeeded" because this is taken care of in the app step of `ingestFile()`.

On the other hand for a negative test where we want to test integration of failures, waiting for `ingestProgressBar` is the way to signal the end and then scrape for the error.

itest/tests/preferences.test.js
```diff
@@ -0,0 +1,30 @@
...
+  stdTest("setting a custom json type config", async (done) => {
+    brim.clickAppMenuItem("preferences")
+    await brim.setValue(jsonTypeConfigInput, config)
+    await brim.click(defaultModalButton)
+    await brim.ingest("custom-sample.ndjson")
+    await brim.search(
+      "_path=conn proto=tcp | cut ts, src_ip, src_port, dst_ip, dst_port, proto | sort ts"
+    )
+
+    expect(await brim.getText(viewerResults)).toMatchSnapshot()
```
mikesbrown, line 27: I wonder if you're getting "End of results" as shown above because you're not using the app step `searchDisplay()`?

mikesbrown: I looked at the `searchDisplay()` code and can't see any reason why it would exclude "End of results"...

mikesbrown: (Answer: different locators)

PR #823, opened 2020-05-27
Reset state if less than v0.9.1

jameskerr:
If a user has a version prior to v0.9.1, we don't try to migrate their state.

If you open the app in v0.8.0, ingest some data, perform some searches, then quit, then open the app at master, you'll see a message in the logs that says:

```
[2020-05-27 14:45:59.759] [info] migrations: unsupported version, using fresh state
```

It will indeed clear the state and start fresh. However, the spaces that you ingested will remain. 

**Known Bug With Space Names**

Testing this revealed a bug in zqd. The spaces created in v0.8.0 (before space ids), are returned without a name after upgrading to the latest brim version. It results in the app looking like this when you load up the newest version. 

image

In this example, I ingested a file called `a.pcap` in v0.8.0, then quit, upgraded, and re-opened brim.

Work on the zqd migration is being tracked in https://github.com/brimsec/zq/issues/791

### DIFF
--- a/src/js/electron/tron/migrations.js
+++ b/src/js/electron/tron/migrations.js
@@ -21,7 +21,8 @@ export type Migrations = {
   run: (VersionedData, Migration[]) => VersionedData,
   getLatestVersion: () => number,
   getPending: () => Migration[],
-  getAll: () => Migration[]
+  getAll: () => Migration[],
+  setCurrentVersion: (number) => void
 }
 
 export default async function migrations(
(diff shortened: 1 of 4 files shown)


mikesbrown:
The failures here look like https://github.com/brimsec/zq/issues/613

alfred-landrum:
I believe this zq PR: https://github.com/brimsec/zq/pull/826 will address the blank space names.

Issue #814, opened 2020-05-26
Eport Only Selected Columns


jameskerr:
Use the cut proc to cut out only the fields that the user has selected in the columns picker.

philrz:
Verified in Brim commit `16a82e1` talking to `zqd` commit `51553cf`.

In the attached video, I select three columns in the picker, export the results, then re-import them into a new Space. This all worked correctly relative to how the `cut` processor works, though using Export the way I did reminded me of some of behaviors of `cut` that we might want to reconsider or add additional options, as users less familiar with `cut` might be thrown by the current behavior.

[Verify.zip](https://github.com/brimsec/brim/files/4703689/Verify.zip)

In the video, you can see how I originally had 6 Zeek events on my screen after I'd narrowed down the set of columns, but then when I exported them, only 2 events made it into the exported ZNG. This seems to be due to the `cut` behavior that only produces an event if there's values for all the listed field names. Here's a repro of that using the same data in `zq` outside the app:

```
$ zq -t all.zng 
#0:record[_path:string,ts:time,ts_delta:duration,peer:bstring,gaps:uint64,acks:uint64,percent_lost:float64]
0:[capture_loss;1585245809.994157;0.487586;zeek;0;2;0;]
#1:record[_path:string,ts:time,peer:bstring,mem:uint64,pkts_proc:uint64,bytes_recv:uint64,pkts_dropped:uint64,pkts_link:uint64,pkt_lag:duration,events_proc:uint64,events_queued:uint64,active_tcp_conns:uint64,active_udp_conns:uint64,active_icmp_conns:uint64,tcp_conns:uint64,udp_conns:uint64,icmp_conns:uint64,timers:uint64,active_timers:uint64,files:uint64,active_files:uint64,dns_requests:uint64,active_dns_requests:uint64,reassem_tcp_size:uint64,reassem_file_size:uint64,reassem_frag_size:uint64,reassem_unknown_size:uint64]
1:[stats;1585245809.994157;zeek;63;9;989;-;-;-;41;40;1;0;0;0;0;0;11;0;1;0;0;0;0;0;0;0;]
[...]

PR #827, opened 2020-05-28
Export Only Selected Columns

jameskerr:
If the user has chosen to only view a subset of the columns in the log viewer, then only export those columns if they click the export button.

fixes: #814


jameskerr:
Merging as it appears the windows test failure is the flaky one.

PR #828, opened 2020-05-28
start comprehensive context menu tests (fixes #804)

### LINKED ISSUE #804: comprehensive context menu tests centered around types
Add tests that exercise the context menu (right-clicking a cell) with a focus on types and special values in those types. This is technically an integration point, because we want to ensure well-formed ZQL continues to be generated. An example of this is ensuring a string field with a quote character is properly escaped.

mikesbrown:
Context menu usage builds ZQL and should be tested as an integration point. For brevity, I've only introduced tests for the "Filter =" option and only 2 types (`string`, `addr`).

We start by introducing a specially-crafted Zeek-style TSV. It is crafted with the following in mind:
- coverage of scalars in records vs. plain scalars
- coverage of containers, both plain and in a record

This coverage exists because ZQ had a history of bugs in this area. The Brim and ZQ unit tests are in good shape, but it's important to make sure Brim and ZQ are in agreement. That hadn't happened before in any automated fashion.

The `string` type is the most interesting type to test for context menus, because there are a variety of considerations:
- How hex sequences in Zeek TSV are rendered in Brim
- The unset representation of ZNG in Brim ('⦻' as of this writing)
- Special Zeek markers, like both literal and hex-sequence tabs and commas, and `(empty)`.
- How Brim treats the above when building ZQL with them.

Here, we can test these strings' Zeek TSV representation through to how they are displayed in Brim, and also how queries are built around them.

`addr`, a much simpler type to test, is introduced simply to show that the code that generates tests handles multiple types.

There were a few unexpected bugs encountered here which prevented some testing.
- #832 
- #833


itest/tests/contextMenu.test.js
```diff
@@ -34,24 +61,63 @@ describe("Test search mods via right-clicks", () => {
     }
   })
 
-  beforeEach(async () => {
-    await writeSearch(app, "")
-    await startSearch(app)
-    await waitForResults(app)
+  stdTest("FilterEq ensure ingest", (done) => {
+    writeSearch(app, `* | count() by _path | sort _path`)
+      .then(async () => {
```
jameskerr, line 66: I've observed this pattern in the stdTests. The first promise used uses the `.then` syntax, and the following promises use the `await` syntax within the first `then`. Finally, the first promise `catch`es the any error and calls our common error handler.

I'd guess that pattern is there so that we can catch all the errors without writing a try catch block. Is that correct?

I thought about a potential solution that would reduce some of the handleError boilerplate and would not need to switch between the different promise syntax.

```js
// Add the app to the stdTest args (needed to catch the errors)
function stdTest(desc, app, func, timeout) {
  test(desc, async (done) => {
    LOG.info(`Running test: ${descr}`)
    try {
      await func(done)
    } catch (err) {
      handleError(app, err, done)
    }
  }, timeout)
}
```

Now the code in the test can always use the await syntax and won't need to handle the error.

```js
stdTest("my test", app, async (done) => {
  await writeSearch(app, "*")
  await startSearch(app)
  await searchDisplay(app)
  ...
  done()
})
```

For your consideration, not a blocker.

itest/tests/contextMenu.test.js
```diff
@@ -34,24 +61,63 @@ describe("Test search mods via right-clicks", () => {
...
+      .catch((err) => {
+        handleError(app, err, done)
+      })
   })
 
-  stdTest("count by _path; sort asc", (done) => {
-    rightClick(app, selectors.viewer.resultCellContaining("capture_loss"))
+  const run = (path: string, values: string[]) => {
+    let testIdx = 0
+    let testId: string
```
jameskerr, line 78: I think this declaration can be moved to line 81. I don't a see a need to declare testId here, then assign it down there. I may be missing something though.

itest/tests/contextMenu.test.js
```diff
@@ -17,14 +18,40 @@ import {
...
+  ";",
+  "∫£œßü™",
+  '"mystr"',
+  "'mystr'",
+  "*",
+  '"*"',
+  "1.1.1.1",
+  " ",
+  "null",
+  "⦻" // The actual character, not unset.
```
jameskerr, line 44: It's reassuring seeing how comprehensive this is with all these edge cases.

itest/tests/contextMenu.test.js
mikesbrown: Appreciate the feedback. It's good you're starting to look about ways to improve integration tests. The feedback helps me be a better Javascript writer.

`stdTest` needs work for sure. I wrote it in early days before I even started using `async` / `await`.

`handleError` really should be used inside `stdTest`; I agree. The flow signature on `stdTest` also somewhat limits its usage.

Something else `stdTest` should do: get the prototype of a jest `test`. This doesn't work for example:
```javascript
stdTest.skip("my test", ....)
```

itest/tests/contextMenu.test.js
mikesbrown: Good spot. Leftover from an internal refactor. Fixed.

mikesbrown:
The failure is brimsec/zq#613

Issue #805, opened 2020-05-21
complaints about missing appState.json on Brim start with no appPath


mikesbrown:
I removed my `run/` directory and ran the app. It complained about not finding appState.json. It seems like if no `run/` directory existed, it shouldn't complain.
```
$ npm run electron

> Brim@0.9.1 electron /Users/mikeb/git/brim
> electron .

13:57:42.391 › app paths: getAppPath=/Users/mikeb/git/brim userData=/Users/mikeb/git/brim/run logs=/Users/mikeb/git/brim/run/logs
13:57:42.409 › Unable to load session state
13:57:42.411 › {
  constructor: 'Error',
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/mikeb/git/brim/run/appState.json',
  stack: "Error: ENOENT: no such file or directory, open '/Users/mikeb/git/brim/run/appState.json'"
}
```

PR #835, opened 2020-05-29
No error if appState.json does not exist

jameskerr:
If appState does not exist, don't throw a crazy error in the main process logs about NOENT. 

```
 {
  constructor: 'Error',
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/mikeb/git/brim/run/appState.json',
  stack: "Error: ENOENT: no such file or directory, open '/Users/mikeb/git/brim/run/appState.json'"
}
```

fixes: #805

### DIFF
--- a/src/js/electron/tron/session.js
+++ b/src/js/electron/tron/session.js
@@ -25,20 +25,21 @@ export default function session(path: string = sessionStateFile()) {
 
     load: async function(): Promise<?SessionState> {
       const migrator = await tron.migrations()
-      const saved = await lib
-        .file(path)
-        .read()
-        .then(JSON.parse)
-        .then((state) => migrate(state, migrator))
-        .catch((e) => {
-          log.error("Unable to load session state")
-          log.error(e)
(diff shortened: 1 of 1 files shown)


src/js/electron/tron/session.js
```diff
@@ -25,20 +25,21 @@ export default function session(path: string = sessionStateFile()) {
...
-          log.error(e)
-          return freshState(migrator.getLatestVersion())
-        })
-
-      if (saved) {
-        version = saved.version
-        return saved.data
+      const file = lib.file(path)
+
+      version = migrator.getLatestVersion()
```
jameskerr, line 30: I realized that in all cases, we are going to set the version to the latest version. So we just do that right here instead of spreading it all around.

If migrations succeed, version will be the latest.
If migrations fail, we return undefined and set the version to the latest.
If the file does not exist, we return undefined and set version to the latest.

mason-fish: Nice, make sense

PR #836, opened 2020-05-29
fix: use the run/ dir if BRIM_ITEST=true

mikesbrown:
There is a case where `BRIM_ITEST=true` but `--user-data-dir` is not specified: manual runs of the app, via command like:
```
BRIM_ITEST=true npm run electron
```
Such cases are useful in order to observe app behavior with the mock context menu.

Before, this would use the default system user data directory. Now, this uses `run/`.

### DIFF
--- a/src/js/electron/appPathSetup.js
+++ b/src/js/electron/appPathSetup.js
@@ -11,14 +11,23 @@ import path from "path"
 function appPathSetup() {
   if (electronIsDev) {
     // isDev is true for general dev execution and integration tests.
-    if (process.env.BRIM_ITEST != "true") {
-      // For general developer execution, put state and logs under a
-      // "run" directory in the git directory.
-      app.setPath("userData", path.join(app.getAppPath(), "run"))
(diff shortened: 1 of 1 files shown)


src/js/electron/appPathSetup.js
```diff
@@ -11,14 +11,23 @@ import path from "path"
...
+    for (const arg of process.argv) {
+      if (arg.startsWith("--user-data-dir")) {
+        hasDataDirArg = true
+        break
+      }
     }
     // We don't override for integration tests as they set userData
     // via the user-data-dir chromeDriver command line argument,
     // so don't override it.
+    if (!hasDataDirArg) {
```
jameskerr, line 26: This electron api might do exactly what you need: https://www.electronjs.org/docs/api/command-line#commandlinehasswitchswitch

mikesbrown: 👍

mikesbrown: done

PR #837, opened 2020-05-29
Update zq to v0.14.0

philrz:


philrz:
Went with a single, larger PR. Closing without merging.

PR #857, opened 2020-06-03
Sidebar makeover

### LINKED ISSUE #844: Move Side Bars Under Tab Bar
To prepare for this design, the left sidebar (history) and the right side bar (log detail) need to be moved in the layout of the UI to beneath the tab bar. 

This ticket is only for moving them in the design, not for re-styling. 

As for the state, the right sidebar's state is already specific to a tab, and the left side bar contents should remain global. 

I think the state will need to be updated so that the left sidebar can be closed or opened in different tabs. Along with resized.

mason-fish:
fixes #844

Refactors the search page to use styled components and places both left and right sidebars under the tab bar and above the status bar.

**left sidebar**
image

**right sidebar**
image

**with status bar visible**
image


src/js/components/SearchPage.js
```diff
@@ -11,6 +11,36 @@ import TabBar from "./TabBar/TabBar"
 import TabContent from "./TabContent"
 import Tabs from "../state/Tabs"
 import useSearchShortcuts from "./useSearchShortcuts"
+import styled from "styled-components"
+
+const ColumnLayout: React$ComponentType<any> = styled.div`
```
mason-fish, line 16: I hit some flow errors here which baffled me, and adding this type annotation/assertion was the only solution that worked. Happy to take different suggestions!

src/js/components/SearchPage.js
```diff
@@ -20,16 +50,20 @@ export default function SearchPage() {
   useEffect(() => () => dispatch(Handlers.abortAll()), [])
 
   return (
-    <div className="search-page-wrapper">
-      <div className="search-page">
-        <LeftPane />
-        <div className="search-page-main">
+    <SearchPageWrapper>
+      <SearchPageMain>
+        <ColumnLayout>
```
jameskerr, line 55: Very cool.

src/js/state/Layout/reducer.js
```diff
@@ -4,7 +4,10 @@ import type {LayoutState, LayoutAction} from "./types"
 
 const init: LayoutState = {
   rightSidebarIsOpen: false,
-  rightSidebarWidth: 450
+  rightSidebarWidth: 450,
+  leftSidebarIsOpen: false,
+  leftSidebarWidth: 350,
+  investigationView: "linear"
```
jameskerr, line 10: We'll need to add a migration for this state change. Maybe all the layouts will be initialized with the old "view" global setting.

src/js/state/View/types.js
```diff
@@ -1,55 +1,21 @@
 /* @flow */
 
-export type InvestigationView = "tree" | "linear"
 export type ViewState = {
```
jameskerr, line 3: We don't need a migration for this since we are deleting and never reading from them again. The next time the state is saved, they will be removed.

src/js/state/migrations/202006091248_moveSidebarViewToTabLayout.test.js
```diff
@@ -0,0 +1,17 @@
+/* @flow */
+
+import getTestState from "../../test/helpers/getTestState"
+import migrate from "./202006091248_moveSidebarViewToTabLayout"
+
+test("migrating 202006091248_moveSidebarViewToTabLayout", () => {
+  let prev = getTestState("v0.9.1")
```
jameskerr, line 7: I think this works. But ideally we'd be testing against the previous released version. So we'd make one called v0.10.0. This can be done in the "Developer" menu when the app is running.

src/js/state/migrations/202006091248_moveSidebarViewToTabLayout.js
```diff
@@ -0,0 +1,23 @@
...
+      tab.layout.leftSidebarWidth = leftSidebarWidth
+    }
+
+    delete s.view.investigationView
+    delete s.view.leftSidebarIsOpen
+    delete s.view.leftSidebarWidth
+  }
+
+  // Migrate state here
+  return state
```
jameskerr, line 22: 🎉

src/js/components/SavedSpacesList.js
```diff
@@ -52,8 +52,8 @@ export default function SavedSpacesList({spaces, spaceContextMenu}: Props) {
                 !s.ingesting() &&
                   showContextMenu(spaceContextMenu(s.id, s.name))
               }}
-              className={`space-link ${
```
jameskerr: Oh yeah! I've been bit by this before. If you feel like using it, the `classNames` library is nice for this. The code you wrote would be equivalent to: 
```js
const className = classNames("space-link", {
  "current-space-link": s.id === currentSpaceId
})
```
 
The function accepts any number of aguments. If it's given a string, that gets added to the className. If it's given an object, all of the keys that have a truthy value to them will be added to the className. If it's given anything falsey, those are excluded.

Pretty handy, and avoids the "extra-space in the className" issue.
```

PR #858, opened 2020-06-04
add a test that creates a zar space under brim

mikesbrown:
Brim isn't fully ready to handle zar spaces yet, but this test
establishes a temporary workflow for getting zar data into a space and
making sure Brim starts. As Brim's functionality to read a Zar space
improves, this test can be modified. The purpose of this is to put
something in place that changes alongside the functionality as opposed
to having nothing until all the functionality is ready. As such, there's
not as much abstraction as there otherwise would be: much of this should
get thrown away once Brim supports Zar.

The test uses Brim as the driver for starting zqd so that the test don't
have to do zqd orchestration itself. It ingests sample.tsv as usual.
Then it uses zealot to create a second, empty space called sample.zar
and also get space metadata from space sample.tsv.brim. The test then
imports samle.tsv.brim's all.zng into sample.zar, with a small enough
chop size to ensure multiple files. The test creates a number of
indexes. Finally, it reloads Brim and ensures Brim sees the second space.

### DIFF
--- a/itest/tests/zar.test.js
+++ b/itest/tests/zar.test.js
@@ -56,8 +56,18 @@ describe("Zar tests", () => {
         // interesting stuff.
         execSync(`"${ZAR}" import -s 1024B -R "${zarRoot}" "${zngFile}"`)
         await Promise.all(
-          [":ip", ":port", "uid", "_path"].map((index) =>
-            exec(`"${ZAR}" index -R "${zarRoot}" ${index}`)
+          [":ip", ":port", "uid", "_path"].map(
+            (index) =>
+              new Promise((resolve, reject) => {
+                const cmd = `"${ZAR}" index -R "${zarRoot}" ${index}`
+                exec(cmd, (error, stdout, stderr) => {
+                  if (error) {
+                    reject(`${cmd}: ${stderr}`)
+                  } else {
+                    resolve(0)
+                  }
+                })
+              })
           )
         )
 
--- a/itest/tests/zar.test.js
+++ b/itest/tests/zar.test.js
@@ -1,0 +1,85 @@
+/* @flow */
+
+// This is needed to use zealot outside a browser.
+global.fetch = require("node-fetch")
+
+import {exec, execSync} from "child_process"
+import path from "path"
+
+import zealot from "../../src/js/services/zealot"
+
+import {nodeZqDistDir} from "../lib/env"
+import {handleError, stdTest} from "../lib/jest.js"
+import {
+  ingestFile,
+  newAppInstance,
+  startApp,
+  waitForNewTab,
+  waitForResults
+} from "../lib/app"
+
+describe("Zar tests", () => {
+  let app
+  let testIdx = 0
+  const ZAR = path.join(nodeZqDistDir(), "zar")
+  beforeEach(() => {
+    app = newAppInstance(path.basename(__filename), ++testIdx)
+    return startApp(app)
+  })
+
+  afterEach(async () => {
+    if (app && app.isRunning()) {
+      return await app.stop()
+    }
+  })
+
+  stdTest(`Brim starts when a Zar space is present`, (done) => {
+    ingestFile(app, "sample.tsv")
+      .then(async () => {
+        await waitForResults(app)
+
+        // Figure out underlying space directories. One will be a
+        // zarRoot. The other will be used to read zng into zar.
+        const client = zealot.client("localhost:9867")
+
+        const sampleSpace = (await client.spaces.list())[0]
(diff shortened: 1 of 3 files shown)


itest/tests/zar.test.js
```diff
@@ -0,0 +1,85 @@
...
+        // sample.tsv.brim. Other times, it's the new tab page. Instead
+        // of forking off the flow further, I'm stopping any testing at
+        // this point.
+        // TODO: Understand why different pages appear when the app is
+        // restarted.
+        try {
+          await waitForResults(app)
+        } catch {
+          await waitForNewTab(app)
+        }
```
mikesbrown, line 78: Try a browser refresh instead.

mikesbrown: Considered and rejected due to other flakiness: in some instances, after issuing a https://www.electronjs.org/docs/api/browser-window#winreload, the Brim app would report the initial ingesting space as deleted. See for example here: https://github.com/brimsec/brim/runs/740557239?check_suite_focus=true

Logs show the Brim app sending the DELETE to zqd for reasons to me that are unclear.
```
[2020-06-05 00:18:56.130] [info] [6663dce9b9]:  refreshing window
[2020-06-05 00:18:56.132] [info] [6663dce9b9]:  starting delete for sp_1csTWovm9caRhn3KdnVzF2o4nDV
[2020-06-05 00:18:56.200] [info] [6663dce9b9]:  Deleted sp_1csTWovm9caRhn3KdnVzF2o4nDV
```
```
{"level":"info","ts":1591316336.133745,"logger":"http.access","msg":"Request completed","request_id":"10","host":"localhost:9867","method":"DELETE","proto":"HTTP/1.1","remote_addr":"127.0.0.1:33708","request_content_length":0,"url":"/space/sp_1csTWovm9caRhn3KdnVzF2o4nDV","elapsed":0.0001414,"response_content_length":0,"status_code":204}
```
Note that nothing in the test at  7bd212b was doing any deleting.

I don't think it's worth it to spend much more time honing this test when the zqd API is going to change again later.

jameskerr: Hey Mike, when you reload the page, brim will delete all spaces that are currently ingesting. It is almost the same logic as closing the window while ingest is happening. Since reloading the page is possible but not common for users to perform, we opted for not presenting a confirmed dialogue when reloading.

mikesbrown: > Hey Mike, when you reload the page, brim will delete all spaces that are currently ingesting.

Thanks for pointing this out. `ingestFile()` is supposed not to return until ingest has completed, and for that, we rely on the app. Maybe something is happening in a CI environment that isn't observable locally such that that logic isn't working.

mikesbrown: @jameskerr I fixed this problem by waiting 10 seconds for the ingest progress indicator to appear. In some cases, the test was racing ahead of the app and moving on because the ingest progress indicator hadn't yet appeared, so the test assumed ingest was finished. 911d308 does this. I was able to reproduce the problem on a linux vm within a few iterations. After 911d308, the test past more than 2000 iterations.

mikesbrown:
This is ready again. The failure is https://github.com/brimsec/zq/issues/613 .

itest/tests/zar.test.js
jameskerr: 2000 iterations! Thorough

itest/tests/zar.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+        const zngFile = path.join(sampleSpace.data_path, "all.zng")
+        const zarRoot = zarSpace.data_path
+
+        // Create a zar archive inside the space and index some
+        // interesting stuff.
+        execSync(`"${ZAR}" import -s 1024B -R "${zarRoot}" "${zngFile}"`)
+        await Promise.all(
+          [":ip", ":port", "uid", "_path"].map((index) =>
+            exec(`"${ZAR}" index -R "${zarRoot}" ${index}`)
```
jameskerr, line 60: Exec doesn't return a promise by default. It returns a child process object. (which Promise.all will think is a resolved promise. 

You can "promisify it" though. There is an example in the docs here: https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

```js
const util = require('util');
const exec = util.promisify(require('child_process').exec);

async function lsExample() {
  const { stdout, stderr } = await exec('ls');
  console.log('stdout:', stdout);
  console.error('stderr:', stderr);
}
lsExample();
```

mikesbrown: I appreciate this feedback. I fixed it in a slightly different way.

jameskerr:
👍

PR #863, opened 2020-06-05
Update for GroupbyProc ast format change

henridf:
Companion to brimsec/zq/pull/860

### DIFF
--- a/src/js/brim/ast.js
+++ b/src/js/brim/ast.js
@@ -11,7 +11,8 @@ export default function ast(tree: Object) {
     },
     groupByKeys() {
       let g = this.proc("GroupByProc")
-      return g ? g.keys : []
+      let keys = g ? g.keys : []
+      return keys.map<string>((k) => k.target)
     },
     proc(name: string) {
       return getProcs(tree).find((p) => p.op === name)


henridf:
Ok, this isn't quite working yet. With this change, after having done `npm install https://github.com/brimsec/zq#groupby-computed-keys && npm install`, I can run the test suite (via `npm test` without errors), and also open an existing space in the app.

However, importing a new space fails with 
image

I'd assume this is related to how the app builds the query for the histogram (although I'm not sure why that would be an issue the first time around and not subsequently when reopening a space). @jameskerr @mason-fish any tips on where the app builds ASTs?

henridf:
I've verified this against the companion branch by:

- running `npm install https://github.com/brimsec/zq#groupby-computed-keys && npm install
`
- verifying that `npm test` passes
- verifying that upon starting the app, I could import a new space and the histogram would show up without errors. also ran a few queries without issue, such as `head 10 | count() by p=String.toUpper(_path)`.

henridf:
(if there's any other manual validation i haven't thought of, please let me know)

mikesbrown:
@henridf I smoke-tested `count() by` with as many types as I could find. I'm not claiming I exhausted all types when doing this, but I found some defects:
1. `_path=conn | count() by id.orig_p` and `id.resp_h`show: "Unknown Expression op FieldCall". I think this must be related to records, because `_path=dhcp | count() by client_addr` seem to work. Confirmed: `_path=x509 | count() by certificate.version` fails in the same way.
1. `_path=conn | count() by proto` has duplicate `proto` keys.

henridf:
Updated brimsec/zq/pull/860 to address the first issue; the second appears to be a pre-existing bug that we weren't aware of (it repros on master).

PR #864, opened 2020-06-08
Use -sorted flag in histogram query

henridf:
Companion to brimsec/zq#869 that tweaks the zql query used to populate the histogram so that it continues to render progressively with results streamed back.


henridf:
`npm test` now passes after doing `npm install https://github.com/brimsec/zq#groupby-every-to-key && npm install` on this branch

PR #866, opened 2020-06-08
Restyle history

### LINKED ISSUE #846: Re-style History Section
It will now have a lighter theme. Here is a snapshot of the design.

Contact me and I can get you the actual design file with the correct pixel values for all the elements.

This PR is simply for re-styling the existing history section. It will not add any more information to the sidebar.

mason-fish:
fixes #846 

depends on and builds off of https://github.com/brimsec/brim/pull/857 (once that goes in, the diff here will be a little smaller and easier to review)

**NOTES**:
* I started a styled-components theme and transferred some of the newer brim styles (specifically the new color palette and a few of our typography mixins) into it. Currently I'm just using the theme provider at the top level, but I imagine we may want to be less global about that in the future, where we only provide the appropriate sub-theme to the appropriate component to properly enforce style options.
* I also wrote these newer restyled components in the paradigm of "styled-components". It makes for a very different pattern in our code. I'm noticing that with all of these smaller styled components our js files want to get much larger.
* There were a few pieces which the .sketch files did not provide a design for or omitted in a way that I did not think was intentional (i.e. the color coating that we do for history searches, as well as the tree view), so I just used my best guess of what would look nice.
* I tried to write the `DropdownHeader` component in a way that I can easily reuse for the `Spaces` section I will be adding to this sidebar next, hence making it somewhat generic despite its current single use case. 

image

image

image


src/js/components/LeftPane.js
```diff
@@ -1,38 +1,123 @@
...
+        <IconWrapper>{icon}</IconWrapper>
+        <Title>{title}</Title>
+        <ViewSelectWrapper>{viewSelect}</ViewSelectWrapper>
+      </SectionHeader>
+      {show && children}
+    </section>
+  )
+}
+
+export function LeftPane() {
```
jameskerr, line 110: I was just thinking last week that this last thing in the file is the thing that we export and all the private functions are above it. That way I always know I can scroll down to find the component that this file exports. Having the exported function end up in the middle always makes it hard to find. What do you think about that convention going forward. 

I often put private functions at the bottom of files since that's what I did in C++ and in ruby classes. But the convention above makes more sense to me in javascript. What do you think?

src/js/components/LeftPane.js
```diff
@@ -47,60 +132,43 @@ export function LeftPane() {
...
       onMouseLeave={() => setShowCollapse(false)}
     >
-      <InvestigationTitleBar onClearAll={onClearAll} />
-      <div className="investigation-pane-body">
-        <InvestigationHeader
-          view={view}
-          onViewChange={onViewChange}
-          onClearAll={onClearAll}
-        />
+      <DropdownHeader
```
jameskerr, line 135: It looks like you want to make a re-usable sidebar header component. I think that's a good idea, but it might be better to wait until we have a few instances first. For example, the spaces header will not have a viewSelect dropdown. And maybe a better name for this is "Sidebar header" or "Collapsible Sidebar Header" or CollabsibleSection". Dropdown to me implies a pop menu.

jameskerr: The colors and placement look great, but some of the styles didn't make it through.

image

In this image above, the triangle is 8x8, the folder icon is 11x11, the text is:
```css
font-size: 9px; 
font-weight: 700;
letter-spacing: 0.8px;
```

If that's not what was in the design let me know. Maybe you had an old version.

src/js/components/LeftPane.js
```diff
@@ -47,60 +132,43 @@ export function LeftPane() {
...
-        <LinkButton
-          className={classNames({selected: view === "linear"})}
-          onClick={linearView}
-        >
-          Linear
-        </LinkButton>
-      </nav>
-      <LinkButton onClick={onClearAll}>Clear</LinkButton>
-    </header>
+    <StyledViewSelect onClick={onClick}>
```
jameskerr, line 168: Let's make this font-size: 11px; and make the svg 8x8 px.

src/js/components/LeftPane.js
```diff
@@ -1,38 +1,123 @@
...
 
-  function onDrag(e: MouseEvent) {
-    const width = e.clientX
-    const max = global.innerWidth
-    dispatch(View.setLeftSidebarWidth(Math.min(width, max)))
+const SectionHeader = styled.div`
+  display: flex;
+  flex-direction: row;
+  background-color: ${(props) => props.theme.colors.coconut};
+  height: 32px;
```
jameskerr, line 28: The height for this is 24px. (since the box-sizing is "border-box", the two 1px borders will be subtracted from this height in actuality. So when you inspect the element in chrome, you should see:

image

jameskerr:
And thank you for filling in the parts of the UI that I forget to make a design for. I think your decisions look great.

mason-fish:
@jameskerr thanks for your feedback, I think I caught all your suggestions. Here are some updated screen shots:
image

image

image

src/js/components/LeftPane.js
```diff
@@ -186,3 +105,45 @@ function InvestigationView({view}) {
...
+          />
+          <Title>History</Title>
+          <ViewSelect />
+        </SectionHeader>
+        {showHistory && <InvestigationView view={view} />}
+      </section>
+      <XLeftPaneCollapser show={showCollapse} />
+    </Pane>
+  )
+}
```
jameskerr, line 149: ♥️ This looks nice.

PR #870, opened 2020-06-10
linux autoupdate notification

### LINKED ISSUE #855: New Brim version notification for Linux
We don't currently support auto-update for Linux releases of Brim, as the Electron auto-updater mechanism ( https://www.electronjs.org/docs/api/auto-updater#platform-notices ) doesn't support it. We'll eventually shift to an auto-update mechanism that does support Linux; in the meantime, we should at least let Linux users know when there's a relevant update for them, and point them to the download page.

mason-fish:
fixes #855 

See ^^ ticket for screenshot of current dialog box as well as implementation notes


mason-fish:
Just confirmed by setting package version to `v0.9.0`, building and running a release on linux:
image

package.json
```diff
@@ -125,6 +124,7 @@
     "electron-log": "^4.1.1",
     "extract-zip": "^1.6.7",
     "fs-extra": "^7.0.0",
+    "got": "^10.7.0",
```
mason-fish, line 127: This is the last minor version before [11.0.0](https://github.com/sindresorhus/got/releases/tag/v11.0.0) which, generally, has breaking changes (though I didn't investigate if any applied to our pretty minimal use of this lib). My own use within this PR will also work for `10.6.0`, I only bumped it bc it was "free" (no extra work needed) and keeps us a bit more up to date.

src/js/electron/autoUpdater.js
```diff
@@ -1,11 +1,59 @@
...
+  return `https://update.electronjs.org/brimsec/brim/${platform}/${app.getVersion()}`
+}
+
+const autoUpdateLinux = async () => {
+  // Check for updates for MacOS and if there are then we assume there is also one for linux
+  const url = getFeedURLForPlatform("darwin-x64")
+  try {
+    const currentVersion = app.getVersion()
+    if (!semver.valid(currentVersion))
+      throw "Invalid current version format: " + currentVersion
```
jameskerr, line 19: Not a blocker.  When you throw just a string, the stacktrace and file info are not saved in the error. If you want any of those things, then you can `throw new Error("message"). Some debug tooling takes advantace of the extra meta information provided by the Error instance (chrome devtools, jest tests).

Here is a little more detail.
https://stackoverflow.com/a/57077203/3499804

mason-fish: Oh this is good to know, thanks!

PR #875, opened 2020-06-15
Remove space name

### LINKED ISSUE #856: Local Storage: only store Space IDs
SpaceIDs are immutable. Everything else in a space (names included) can change. Along these lines zq#853 includes a migration that may mutate some of the names of stored spaces. If space names are stored in local storage they could become out-of-date.

As such the local storage of the brim app should store the space id when it references a space. No doubt this change will have to include functionality to refetch space information on app init.

jameskerr:
fixes #856 

Get the name of a space from the spaces state slice rather than from the cached spaceName property on the SearchRecord.

This allows zqd to migrate the space names, and have the app be up to date with those changes.

An side effect of no longer storing the names is that we cannot display the name in the "deleted" messages. So I've changed them to be more generic. Previously this message included the space name in it like: "Default" space has been deleted.
image

### DIFF
--- a/src/js/components/App.js
+++ b/src/js/components/App.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import {useDispatch, useSelector} from "react-redux"
+import {useSelector} from "react-redux"
 import React, {useEffect} from "react"
 
 import {ipcRenderer} from "electron"
@@ -12,18 +12,15 @@ import ErrorNotice from "./ErrorNotice"
 import HTMLContextMenu from "./HTMLContextMenu"
 import Preferences from "./Preferences/Preferences"
 import Prefs from "../state/Prefs"
+import SpaceModal from "./SpaceModal"
 import View from "../state/View"
 import brim from "../brim"
-import refreshSpaceNames from "../flows/refreshSpaceNames"
-import SpaceModal from "./SpaceModal"
 
 export default function App() {
   brim.time.setZone(useSelector(View.getTimeZone))
   brim.time.setDefaultFormat(useSelector(Prefs.getTimeFormat))
-  let dispatch = useDispatch()
 
   useEffect(() => {
-    dispatch(refreshSpaceNames())
     ipcRenderer.invoke("windows:ready")
   }, [])
 
--- a/src/js/flows/openNewSearchWindow.js
+++ b/src/js/flows/openNewSearchWindow.js
@@ -1,20 +1,20 @@
 /* @flow */
 
+import type {Thunk} from "redux-thunk"
+
+import SearchBar from "../state/SearchBar"
+import Tab from "../state/Tab"
 import invoke from "../electron/ipc/invoke"
 import ipc from "../electron/ipc"
-import type {Thunk} from "redux-thunk"
-import Tab from "../state/Tab"
-import SearchBar from "../state/SearchBar"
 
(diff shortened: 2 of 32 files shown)


src/js/flows/openNewSearchWindow.js
```diff
@@ -1,20 +1,20 @@
...
   const state = getState()
 
   const params = {
-    spaceName: Tab.getSpaceName(state),
     spaceId: Tab.getSpaceId(state),
     span: Tab.getSpan(state),
     program: SearchBar.getSearchBarInputValue(state)
   }
-
+  // $FlowFixMe
```
jameskerr, line 18: Flow is having a hard time understanding that a Span is a subset of a SpanArgs.

src/js/components/App.js
```diff
@@ -12,18 +12,15 @@ import ErrorNotice from "./ErrorNotice"
...
-import refreshSpaceNames from "../flows/refreshSpaceNames"
-import SpaceModal from "./SpaceModal"
 
 export default function App() {
   brim.time.setZone(useSelector(View.getTimeZone))
   brim.time.setDefaultFormat(useSelector(Prefs.getTimeFormat))
-  let dispatch = useDispatch()
 
   useEffect(() => {
-    dispatch(refreshSpaceNames())
```
jameskerr, line 26: This happens in the initializer now before react boots up: https://github.com/brimsec/brim/pull/875/files#diff-0b427a5466e0c7a032be98e7a3e28501R50

src/js/state/Search/types.js
```diff
@@ -3,15 +3,14 @@ import type {DateTuple} from "../../lib/TimeWindow"
 import type {SEARCH_BAR_SUBMIT} from "../SearchBar/types"
 import type {Span, Ts} from "../../brim"
 
-export type SearchState = {
+export type SearchState = {|
```
jameskerr, line 6: In flow the {||} syntax means an "exact" object with only and all these properties. The default {} syntax means "at least" these properties.

src/js/state/Tab/activeTabSelect.js
```diff
@@ -5,6 +5,12 @@ import type {State} from "../types"
...
 import Tabs from "../Tabs"
 
-export default function activeTabSelect<T>(fn: (TabState) => T): (State) => T {
-  return createSelector<State, void, T, TabState>(Tabs.getActiveTab, fn)
+export default function activeTabSelect<T>(
+  fn: (TabState, State) => T
+): (State) => T {
+  return createSelector<State, void, T, TabState, State>(
+    Tabs.getActiveTab,
+    (state) => state,
```
jameskerr, line 13: We now pass in all of the state as the second parameter to the activeTabSelect function.

src/js/state/globalReducer.js
```diff
@@ -2,18 +2,16 @@
...
-import type {SpacesState} from "./Spaces/types"
+import type {PrefsState} from "./Prefs/types"
 import Investigation from "./Investigation"
 import Prefs from "./Prefs"
-import Spaces from "./Spaces"
 
 export type GlobalState = {
   investigation: InvestigationState,
-  spaces: SpacesState
+  prefs: PrefsState
```
jameskerr, line 11: We forgot to add the prefs: key to the GlobalState flow type.

PR #879, opened 2020-06-16
Add spaces to sidebar

### LINKED ISSUE #848: Add Spaces List to Left Pane
Contact James for the real design files.

Also, please hide the "i" icon for now since we don't have a page to see the Space details yet.

mason-fish:
fixes #848 

Here are some open questions that i've observed throughout implementation, mostly relating to history:
* Clicking on history items will execute the "program"/search that is presented but will run in the historical space as well but we don't show what space is associated with each history item. So this is a bit confusing.
* We also record empty searches every time we switch spaces. This pollutes the history list with a bunch of `*`s that aren't particularly useful to the user.
* We have dropped the inline `clear all` history link. We can fit it back inside another context menu for the history section, or if we take more of a browser-based approach we can put that operation within the preferences modal. I think I prefer the former.
* Finally, the empty state/first-time load of the app hits you with a somewhat bland look now that the spaces selection has been moved entirely to the left sidebar. Ideas are to either add some empty state text to the sidebar to fill in those dropdowns a bit more, or we adjust the default view state for the left sidebar such that it always defaults to open *unless* there is a new tab open and there are no existing spaces.

empty state, first time launch of the app looks like this now:
image

with data:
image

image

image

`Delete` has been moved to the spaces context menu. We also do not render this menu mid-ingest 
image


mason-fish:
@jameskerr thanks for the fast designs on empty state!
image

siskojr:
Great work. Looks amazing.

jameskerr:
> Clicking on history items will execute the "program"/search that is presented but will run in the historical space as well but we don't show what space is associated with each history item. So this is a bit confusing.

This is just the way it is for now, until we update the design. As a quick fix, we could add a title attribute that shows the space so that when a user hovers over the row they can see the space and maybe time span for that search.

> We also record empty searches every time we switch spaces. This pollutes the history list with a bunch of *s that aren't particularly useful to the user.

Agreed, let's not do that. In the "initSpace" flow, I think we execute a search. Let's set the "saveToHistory" flag to false on that.

> We have dropped the inline clear all history link. We can fit it back inside another context menu for the history section, or if we take more of a browser-based approach we can put that operation within the preferences modal. I think I prefer the former.

Let's add it as a right click option to all history entries. So I'll be able to right click any history entry and select "Clear All".

# Additional Comments

- [ ] I changed my mind on the default width of the sidebar. Would you be willing to make it `270px`? After pulling it up in real life it looked a little small.

- [ ] Rather than having the whole left sidebar scroll, let's make it so that each section is it's own scrollable container. That means that the spaces scrolls independently of the history. 

- [ ] I've updated label-small to now has a line-height of 17px. It will be apparent here:

image

src/js/components/LeftPane.js
```diff
@@ -122,53 +135,82 @@ function InvestigationTree() {
...
+
+  const paneHeight = useRef(0)
+
+  function onDragPane(e: MouseEvent) {
     const width = e.clientX
     const max = global.innerWidth
     dispatch(Layout.setLeftSidebarWidth(Math.min(width, max)))
   }
 
+  function onDragSpaces({dy, type}) {
```
jameskerr, line 163: While we're dragging, well want to change the cursor of the mouse to remain in the "row-resize" position.

Right now as I drag, the cursor flickers between "row-resize" and "default" as I enter and leave the DragAnchor. 

To do this with javascript, you can either add class to the body called like "dragging-row" with the proper css.

Or you can use inline styles and set `document.body.style.cursor = "row-resize" in the "down" event, then switch back to `document.body.style.cursor = ""` in the "up" event.

src/css/_finding-card.scss
```diff
@@ -1,12 +1,13 @@
...
-  padding-left: $space-l;
+  display: flex;
+  align-items: center;
+  height: 25px;
   color: white;
+  padding-left: 22px;
+  margin: 0 2px;
   border-radius: 5px;
-  margin: 1px 3px;
+  cursor: pointer;
```
jameskerr, line 9: The only think I use cursor: pointer for is links (underlined, blue links that point to webpages like docs). Most everything in the App is `cursor: default`.

This is different from most web apps out there which make everything that is possibly clickable `pointer: cursor`. However, native apps tend not to. Since we're all into being native, let's mimic them.

https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

src/js/components/LeftPane.js
```diff
@@ -20,6 +25,18 @@ const Arrow = (props) => {
...
+`
+
+const SectionContents = styled.div`
+  height: 100%;
+  display: ${(props) => (props.show ? "block" : "none")};
+  overflow-y: scroll;
+`
+
 const SectionHeader = styled.div`
   display: flex;
```
jameskerr, line 41: If you would add a `user-select: none` here, that would make it feel more native as well. That when when a user is interacting with it, it won't make a selection:

image

src/css/_finding-card.scss
jameskerr: This also applies to the Section Header UI elements. All those can be cursor: default as well.

Issue #878, opened 2020-06-16
reload is flaky


mikesbrown:
[This test run](https://github.com/brimsec/brim/runs/777908057?check_suite_focus=true) failed in Spectron `app.browserWindow.reload()`. We know this is so because it's written inside an `appStep`, and that `appStep` never completed.

```
[2020-06-16 19:44:14.955 debug]: Starting step "app reload"
[2020-06-16 19:44:16.769 error]: handleError: Test hit exception: unknown error: cannot determine loading status
from unknown error: unhandled inspector error: {"code":-32000,"message":"Inspected target navigated or closed"}
```
Looks related to or actually is https://github.com/electron-userland/spectron/issues/493 .

wolfoo2931:
for me this workaround solved the problem:

`this.app.client.execute(function() { location.reload() })`

Best Regards,
Oliver, working on [www.monsterwriter.app](https://www.monsterwriter.app/)

PR #881, opened 2020-06-17
separate out appStep api

mikesbrown:
- the notion of an appStep is now a logStep
- app steps call logStep
- separate consumable appSteps into api/ from things app steps use into
  util/
- prefer split files
- prefer export default where possible
- create an index.js that supplies the most common steps; the rest of
  the API can still be imported as needed
- remove various holdover steps that are redundant or not used
- update tests

The two-hierarchy approach as seen in Zealot isn't implemented here. I
figure createTestBrim can have that, if we ever want it.


itest/lib/appStep/api/reload.js
```diff
@@ -0,0 +1,8 @@
+/* @flow */
+import {Application} from "spectron"
+
+import logStep from "../util/logStep"
+
+export default async (app: Application) => {
+  await logStep("app reload", () => app.browserWindow.reload())
+}
```
jameskerr, line 8: When the return value of a function is a promise, and you don't need to capture a promise fulfillment in a variable, the async/await keywords are redundant.

This could be written like this:

```js
export default (app) => {
  return logStep("app reload", () => app.browserWindow.reload())
}
```

Whether you keep in the way it is, or change it to the above, the caller will probably "await" in the test file.

```js
await reload()
```

itest/lib/appStep/api/savePcap.js
```diff
@@ -0,0 +1,22 @@
...
+    await app.client.waitForVisible(selectors.downloadMessage)
+    return await retryUntil(
+      () => app.client.getText(selectors.downloadMessage),
+      (text) => text == "Download Complete" || text.includes("Download error")
+    )
+  })
+}
+
+export const pcapsDir = async (app: Application) =>
+  await app.electron.remote.app.getPath("temp")
```
jameskerr, line 22: Building off the comment above, I guess a general rule could be that if you are returning a promise, there is no need to await it right before you return it.  Await should be used only as an alternative to a `.then()` call.

This line above would function identically to:

```js
export const pcapsDir = (app) => app.electron.remote.app.getPath("temp")
```

itest/lib/appStep/api/search.js
```diff
@@ -0,0 +1,74 @@
+/* @flow */
+
+import {Application} from "spectron"
+
+import logStep from "../util/logStep"
+import waitForResults from "../util//waitForResults"
+import {selectors} from "../../../../src/js/test/integration"
+
+export default async (app: Application, searchText: string) => {
```
jameskerr, line 9: Mason and I recently agreed upon a convention where the default export is always at the bottom of the file, and private functions are on top. Since javascript has all these imports, it's sometimes hard to find the default export when it's in the middle of the file. It's not what I usually do in other languages, but I think it works well for javascript. Would you be willing to adopt that convention here too?

itest/lib/createTestBrim.js
```diff
@@ -1,14 +1,9 @@
...
-import {
-  appStep,
-  click,
-  ingestFile,
-  newAppInstance,
-  startApp,
-  startSearch,
-  writeSearch
-} from "./app"
+import appStep from "./appStep/api"
```
jameskerr, line 3: Very cool!

itest/tests/preferences.test.js
```diff
@@ -5,7 +5,7 @@ import {
   defaultModalButton,
   jsonTypeConfigInput
 } from "../../src/js/test/locators"
-import {searchDisplay} from "../lib/app"
+import {getResults} from "../lib/appStep/api/search"
```
jameskerr, line 8: I like this name more too.

itest/tests/query.test.js
```diff
@@ -43,20 +35,12 @@ describe("Query tests", () => {
...
     let zql = simpleQueries[i]
     let testId = sprintf("%03d", i)
     stdTest(`query${testId}: "${zql}"`, (done) => {
-      writeSearch(app, zql)
-        .then(async () => {
-          await startSearch(app)
-          return searchDisplay(app)
-        })
+      appStep
+        .search(app, zql)
```
jameskerr, line 43: How cool, the search function returns the results. That's very nice.

itest/lib/appStep/api/reload.js
mikesbrown: This is going to get attention as part of #878

itest/lib/appStep/api/search.js
mikesbrown: Absolutely. I had thought to put it at the top as the first useful thing to be observed, but your import point makes sense.

itest/lib/appStep/api/savePcap.js
mikesbrown: This is a helpful comment. Fixed.

jameskerr:
👍

PR #884, opened 2020-06-18
Add cluster picker to sidebar

### LINKED ISSUE #847: Add Cluster Info to the Left Pane
Contact James for the exact design files.

There does not need to be a menu dropdown yet, so you can hide that down carrot before merging this pr.

mason-fish:
fixes #847 

Drops the `<ClusterPicker />` back in with some small restyling. The "Add new cluster" functionality takes us to the dormant sign-in page, which is next up for being redesigned (though we may want to remove that functionality until after it's new design is implemented).

image

image

image


mason-fish:
Latest simplified appearance
image

src/css/_saved-spaces-list.scss
```diff
@@ -57,6 +57,8 @@
     height: 25px;
     margin: 0 2px;
     width: 95%;
+    padding-left: 22px;
```
mason-fish, line 60: I meant for this to be in the spaces PR, I think while I was doing some rebasing it may have gotten dropped

jameskerr: Right on.

src/js/components/LeftPane.js
```diff
@@ -31,6 +32,7 @@ const StyledSection = styled.section`
   min-height: 24px;
   flex-basis: 0%;
   flex-shrink: 1;
+  padding-bottom: 24px;
```
mason-fish, line 35: I noticed this in this PR as well. This padding is needed for each sectionContents to be able to scroll through all of it's contents properly since we are setting the height dynamically on the containing section element now and using flex instead of setting the height on the actual sectionContents.

PR #888, opened 2020-06-22
adjust advance zq ci timeouts

mikesbrown:
Webdriver-style tests wait for elements to appear or to be set to a
specific state. This means a lot of test failures are timeout-driven.
An overall job timeout is already set here in case turnstyle hangs.
However, the current value is too short in cases where Zq and Brim are
so broken as to cause every integration test to fail. Locally under
these conditions, the integration tests took 63 minutes to fail. In CI,
assume that could take up to 90 minutes. Set that as the integration
test step timeout. Increase the job timeout to 120.

Other PRs will make other adjustments and hopefully allow for these
numbers to be reduced or removed.

### DIFF
--- a/.github/workflows/advance-zq.yml
+++ b/.github/workflows/advance-zq.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         node-version: [12.x]
         os: [ubuntu-18.04]
-    timeout-minutes: 30
+    timeout-minutes: 120
     steps:
       # Only one of these should run at a time, and the checkout of brim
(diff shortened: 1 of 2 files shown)


jameskerr:
Hey Mike, we could use this option too when running the tests in CI.

> **--bail**
>
>Alias: -b. Exit the test suite immediately upon n number of failing test suite. Defaults to 1.

https://jestjs.io/docs/en/cli.html#--bail

mikesbrown:
> Hey Mike, we could use this option too when running the tests in CI.

Thanks for the suggestion! Done.

Issue #849, opened 2020-06-02
Create the New Connection Page


jameskerr:
This page already exists in the app somewhere. Find it and update it to the new design. This will appear when the user does not have a "current cluster" selected for that tab.

philrz:
Verified in Brim commit `d8b8f1e` talking to `zqd` commit `4444632`.

In addition to my local `zqd` launched by the local Brim app on my Mac, I also started a remote `zqd` in an Ubuntu VM running in VirtualBox:

```
$ /home/phil/brim/zdeps/zqd listen -l :9867 -data /home/phil/brim/run/data/spaces -config /home/phil/brim/run/zqd-config.yaml -zeekrunner /home/phil/brim/zdeps/zeek/zeekrunner
```

I then set up a Port Forwarding rule in VirtualBox to redirect `localhost:9999` on my Mac to port `9867` inside the Ubuntu VM.

As shown in the attached video, I was able to use the "New Connection" modal to connect to `localhost:9999` at which point I was connected to the remote `zqd` and able to browse data that's stored in the remote `zqd`.

[Verify.zip](https://github.com/brimsec/brim/files/5139654/Verify.zip)

Thanks @mason-fish!

PR #889, opened 2020-06-22
Standardize Form Input Across the App

jameskerr:
Five new common components have been added:

* TextInput
* SelectInput
* FileInput
* InputLabel
* InputField

They all look something like this: 

image

These have replaced all the form components across the app.

This is in preparation for implementing issue #849


src/js/components/DebugModal.js
```diff
@@ -35,15 +36,15 @@ function DebugModalContents() {
...
+      <InputField>
+        <InputLabel>Query:</InputLabel>
+        <TextInput
+          autoFocus
           value={program}
           onChange={(e) => setProgram(e.target.value)}
           {...reactElementProps("debugProgram")}
         />
-      </Form>
+      </InputField>
```
jameskerr, line 47: Screen Shot 2020-06-22 at 12 50 08 PM

src/js/components/SpaceModal.js
```diff
@@ -42,14 +44,17 @@ export default function SpaceModal() {
...
+        <InputLabel>New Name:</InputLabel>
+        <TextInput
+          autoFocus
+          label="Space Name"
+          value={value}
+          onChange={(e) => {
+            onChange(e.target.value)
+          }}
+        />
+      </InputField>
```
jameskerr, line 57: Screen Shot 2020-06-22 at 12 49 37 PM

src/css/_time-span-pickers.scss
```diff
@@ -24,6 +24,7 @@
 
   input {
     @include label-small;
+    @include input-focus;
```
jameskerr, line 27: Makes editing the time span look like this: 
Screen Shot 2020-06-22 at 12 33 16 PM

src/js/components/SpaceModal.js
mason-fish: I think that `label="Space Name"` prop/attribute in the `<TextInput />` will not be of use anymore.

mason-fish: Also, I'm aware that putting colons after labels is helpful for older screen readers to parse forms but from a design perspective I find I prefer these labels only with the emphasis we are already giving them (boldness, placement, followed by a clearly editable box) and no colon. Just one person's opinion though :)

src/css/settings/_colors.scss
```diff
@@ -43,8 +43,20 @@ $pink: #ad3f95;
 
 $purple: #544aa6;
 $orange: #eeb457;
-$aqua: #252525;
-$slate: #737373;
+
+$azure: #2f629c;
```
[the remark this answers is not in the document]
jameskerr: Thanks for bringing this up. My first inclination is to hold onto my scss files. I enjoy writing scss, it's super stable (12 years old) and simple, and the feedback loop is super short (due to livereload). 

However, I was skeptical of when html and javascript were blended together with react, but that turned out to be a great idea. Blending the css into the component is probably going to be a good idea too. The fact that you've already used it and really like it is a plus too.

Ok, I'm willing do an experiment for a while where every new component I make will use StyledComponents.


Some articles I read that influenced me: 

https://medium.com/building-crowdriff/styled-components-to-use-or-not-to-use-a6bb4a7ffc21

https://blog.logrocket.com/moving-from-scss-to-styled-components-advantages-and-caveats/

PR #893, opened 2020-06-25
add view toolbar button

### LINKED ISSUE #845: Add "View" Tool Bar Button
It should go right above the two left and right arrows as in the design above.

The svg icon is to be used is here. 

The menu items should be:
* {checkmark} Left Pane
* {checkmark} Right Pane

Ideally we use the native menu popup.

The checkmarks indicate whether is visible or not.

mason-fish:
fixes #845 

brim


src/js/components/LeftPane.js
```diff
@@ -1,8 +1,7 @@
 /* @flow */
 
 import {useDispatch, useSelector} from "react-redux"
-import React, {useState, useRef} from "react"
-import {XLeftPaneCollapser} from "./LeftPaneCollapser"
```
jameskerr, line 5: 🔥

PR #894, opened 2020-06-25
Add a context option to "zoom out" to unfiltered data

r-franke:
I've been using Brim extensively over the past week and I'm loving it. I was just missing a small feature that I implemented.

Often when I'm investigating data I will work from a filter, say DNS to locate something of interest. What I would like to do then is to "zoom out" of my filter back to the unfiltered data to see what's happening in the direct timespan surrounding my point of interest.

Of course we already had "Use as start time"  and "Use as end time". This pull requests adds a option to "View in full context"  in the context menu.

**Image example**
image

**What it does**
- It set the start and end time of the timespan to + and - 1 minute respectively.
- It clears the searchbar
- It submits the new search
- After a timeout of 1 seconds it scrolls the View to the middle of the `view.scrollHeight`.

**Possible improvements**
The way I scroll to the middle is a bit rough. I've tried figuring out whether I could scroll to the specific line that I clicked on, however I think chunking is standing in the way of that. So jumping to the middle of `scrollHeight` is an approximation. 

The timeout before the scroll is necessary to wait for the rendering to finish. There are probably better ways to fix this, but I'm coming from a Vue background so React was quite unfamiliar to me.

**Test results**
I've ran the tests as described in the contribution guidelines. Everything is passing except for the itest phase:

```
[2020-06-25 14:12:09.769 info]: Running test: histogram deep inspection
[2020-06-25 14:13:12.272 error]: handleError: Test hit exception: element ("[data-test-locator='span_button']") still not visible after 60000ms
[2020-06-25 14:13:12.272 info]: handleError: Looking for any desktop app notifications
[2020-06-25 14:13:12.437 info]: wrote out screen shot to "run/itest/failure-435313d5.png"
[2020-06-25 14:13:12.444 info]: handleError: desktop app notification not found
 FAIL  itest/tests/histogram.test.js (65.868s)
  Histogram tests
    ✕ histogram deep inspection (65711ms)

  ● Histogram tests › histogram deep inspection

    element ("[data-test-locator='span_button']") still not visible after 60000ms

      at new WaitUntilTimeoutError (../node_modules/webdriverio/build/lib/utils/ErrorHandler.js:149:12)
      at ../node_modules/webdriverio/build/lib/commands/waitUntil.js:29:19
```

I think it's related to my environment (Manjaro with i3wm)

_This is my first pull request. Feedback is welcome._


alfred-landrum:
hi @r-franke - just a note that we're happy to hear you've been using Brim, and thrilled that you want to contribute! The UI folks will review this soon & get back with comments.

src/js/electron/menu/actions/searchActions.js
```diff
@@ -101,6 +101,26 @@ function buildSearchActions() {
...
+      label: "View in full context",
+      listener(dispatch, fieldData) {
+        let field = brim.field(fieldData)
+        let brimTime = brim.time(field.toDate())
+        if (field.type === "time") {
+          dispatch(tab.setFrom(brimTime.subtract(1, "minutes").toTs()))
+          dispatch(tab.setTo(brimTime.add(1, "minutes").toTs()))
+          SearchBar.clearSearchBar()
+          dispatch(SearchBar.changeSearchBarInput(""))
+          dispatch(submitSearch())
```
jameskerr, line 115: Like you mentioned in the description, the setTimeout below won't be a reliable way to run code after the Viewer (table) has rendered. For that, it might make sense to have `submitSearch()` return a promise that resolves when the table has received all it's events. You could then write `dispatch(submitSearch()).then(() => /* Code that scrolls to log */)` 

As for scrolling to the proper log. For this, I think we should save the scroll position (maybe the x and y values) in the redux store. This would be in the `state/Viewer/reducer.js`. Then we could simply dispatch an action to set that state and then update the Viewer component to "react" to the change. Then we don't have to worry about timing the events just right.

Maybe a new file called src/js/flows/scrollToLog.js which could contain a redux "thunk" that looks something like this:

```js
export default const (log) => (dispatch, getState) => {
  const state = getState()
  const logs = Viewer.getLogs(state)
  const index = logs.findIndex(log2 => Log.is(log, log2))
  const rowHeight = 25 // pixels
  const scrollY = index * rowHeight
  
  dispatch(Viewer.setScroll(0, scrollY))
}
```

The this function in the searchActions might look like:
```js
/* set time and search bar ... */

dispatch(submitSearch()).then(() => {
  dispatch(scrollToLog(log))
})
```

src/js/electron/menu/searchFieldContextMenu.js
```diff
@@ -64,6 +64,7 @@ export default function searchFieldContextMenu(
       menu.separator(),
       searchMenuActions.fromTime.menuItem([field], {enabled: isTime}),
       searchMenuActions.toTime.menuItem([field], {enabled: isTime}),
+      searchMenuActions.jumpToTime.menuItem([field], {enabled: isTime}),
```
jameskerr, line 67: If you want access to the log in your menu action listener,  you could add it after `field`: [field, log]. Then it will the last argument in the listener above.

src/js/electron/menu/actions/searchActions.js
```diff
@@ -101,6 +101,26 @@ function buildSearchActions() {
...
+    jumpToTime: action({
+      name: "search-cell-menu-show-context",
+      label: "View in full context",
+      listener(dispatch, fieldData) {
+        let field = brim.field(fieldData)
+        let brimTime = brim.time(field.toDate())
+        if (field.type === "time") {
+          dispatch(tab.setFrom(brimTime.subtract(1, "minutes").toTs()))
+          dispatch(tab.setTo(brimTime.add(1, "minutes").toTs()))
+          SearchBar.clearSearchBar()
```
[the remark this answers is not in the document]
r-franke: Changed

src/js/electron/menu/actions/searchActions.js
r-franke: So everything is working. I'm not really sure about my approach though.

I've tried to solve this in `flows/submitSearch.js` initially. My approach was periodically reading the `Viewer.endState` variable to see if it was done fetching. However that required doing something periodically again and it wasn't working reliably. I couldn't figure out why.

So I changed around to the suggested Promise:

```javascript
/* .... */
    return new Promise(function(resolve) {
      switch (args.type) {
        case "analytics":
          dispatch(executeTableSearch(tabId, resolve, args))
          break
        case "zoom":
          dispatch(executeTableSearch(tabId, resolve, args))
          break
        default:
          dispatch(executeTableSearch(tabId, resolve, args))
          dispatch(executeHistogramSearch(tabId, args, prevArgs))
      }
    })
```

Let me know if there is a better way.

r-franke:
I see that I'm getting some errors on both the CI side and on my side. After a full system upgrade my npm stopped working. Brim wouldn't boot until I did a npm update. I didn't push the new `package.json` could this be the case for the errors?

src/js/components/Viewer/Viewer.js
```diff
@@ -63,6 +65,12 @@ export default function Viewer(props: Props) {
     }
   })
 
+  useEffect(() => {
+    let view = ref.current
+    if (!view) return
+    view.scrollTo(props.scrollX, props.scrollY)
+  }, [props.scrollX, props.scrollY])
```
jameskerr, line 72: This is great.

src/js/electron/menu/actions/searchActions.js
```diff
@@ -101,6 +102,22 @@ function buildSearchActions() {
...
+      listener(dispatch, fieldData, log) {
+        let field = brim.field(fieldData)
+        let brimTime = brim.time(field.toDate())
+        if (field.type === "time") {
+          dispatch(tab.setFrom(brimTime.subtract(1, "minutes").toTs()))
+          dispatch(tab.setTo(brimTime.add(1, "minutes").toTs()))
+          dispatch(SearchBar.clearSearchBar())
+          dispatch(submitSearch()).then(() => {
+            dispatch(scrollToLog(log))
+          })
```
jameskerr, line 117: Awesome. The scrolling will be much more reliable now I believe.

src/js/flows/executeTableSearch.js
```diff
@@ -28,10 +29,10 @@ export default function executeTableSearch(
       )
       .stats((stats) => dispatch(Viewer.setStats(tabId, stats)))
       .error((error) => dispatch(Notice.set(ErrorFactory.create(error))))
-      .end((_id, count) =>
+      .end((_id, count) => {
+        resolve()
```
jameskerr, line 33: Yesterday I was thinking that this would be the way to do it, and I see you've found it too. But now I think we could make this simpler by having `executeTableSearch` return the promise. (I noticed we don't rely on the return value of `executeTableSearch` anywhere). Then in submitSearch(), have the last line be `return executeTableSearch(...)`. And `fetchNextPage()` wouldn't need to change at all.

For example:


```js
function executeTableSearch(...):  {
  return function(dispatch) {
    return new Promise((resolve, reject) => {
        ....
        .error((error) => {
          dispatch(Notice.set(ErrorFactory.create(error)))
          reject(error)
        })
        .end((_id, count) => {
          dispatch(Viewer.setEndStatus(tabId, endStatus(count)))
          resolve()
        })
      ...
    })
  }
}
```

And the switch statement at the end of submitSearch would now look like:

```js
switch (args.type) {
      case "analytics":
        return dispatch(executeTableSearch(tabId, args))
      case "zoom":
        return dispatch(executeTableSearch(tabId, args))
      default:
        dispatch(executeHistogramSearch(tabId, args, prevArgs))
        return dispatch(executeTableSearch(tabId, args))
    }
```

In the case where we bail early if the syntax isn't valid (line 25 of submitSearch.js), it might be good to return `Promise.reject()` to keep the return values the same type.

```js
if (!dispatch(SearchBar.validate())) return Promise.reject()
```

r-franke: Good suggestion. I think I've got this sorted now in the last commit. Reverted the changes in `fetchNextPage` as well.

philrz:
Verified in Brim commit `f90c2ea` talking to `zqd` commit `276030d`.

In the attached video, I drill down to a handful of events involving a particular hostname. When I right-click on the new "View in full context" option, you can see how the time range changes to 2 minutes (indicating the +/- 1-minute around the event I clicked on) and I'm placed in the middle of that range with the event I clicked on now at the top of the view. Note that the "View in full context" option currently only becomes available if I'm right-clicking on the `ts` field of the event.

[Verify.zip](https://github.com/brimsec/brim/files/4897627/Verify.zip)

Thanks @r-franke!

PR #898, opened 2020-06-26
Add error message for space renames

### LINKED ISSUE #900: Space name errors not being displayed to user
After the Space name constraints that came with https://github.com/brimsec/zq/pull/853, I tested out the user experience of what happens if I tried to violate those constraints, such as:

1. Attempting to rename a Space to one that already exists
2. Attempting to rename a Space to include invalid characters

As shown in the attached video, in both cases the app silently fails to process the rename. The error message that `zqd` floated shows up in DevTools but is not seen by the user. Surely it would be helpful to the user to have seen it so they can understand what they did wrong.

[Repro.zip](https://github.com/brimsec/zq/files/4834260/Repro.zip)

jameskerr:
Fixes #900 

If zqd responds with an error, we show it in the form next to the label.

Screen Shot 2020-06-25 at 8 32 46 PM

Whatever zqd returns as the {error: "message goes here"} is forwarded to the UI.

### DIFF
--- a/src/css/settings/_colors.scss
+++ b/src/css/settings/_colors.scss
@@ -36,6 +36,9 @@ $green: #399c81;
 $green-dark: darken(#399c81, 15%);
 
 $red: #e65835;
+:root {
+  --red: #e65835;
+}
 $red-light: lighten($red, 15%);
 $red-dark: darken($red, 15%);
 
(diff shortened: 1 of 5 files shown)


alfred-landrum:
I opened https://github.com/brimsec/zq/pull/945 so that the error displayed should be:
```
name may not contain / or non-printable characters
```

Issue #897, opened 2020-06-25
flaky chromedriver loaded a second chrome session, lingering zqd caused downstream test problems


mikesbrown:
https://github.com/brimsec/brim/runs/808064105?check_suite_focus=true

The sequence of test suites was:
1. histogram
1. type context menu
1. pcap
1. [the rest, not relevant here]

Every test suite expects to start in a new appDir with no spaces.

Pcap tests were the first to start failing. The first screen shot shows existing spaces when we didn't expect to have any.

![image](https://user-images.githubusercontent.com/16110130/85792143-e55d0280-b6e7-11ea-98cd-7cdae4a1e0bf.png)

It's clear from the artifacts that `histogram.test.js/1`'s `zqd` was handling requests for other tests. Type context menu tests still worked because there wasn't already a `types.tsv.brim` space. But other tests, which expect to be able to create `sample.pcap.brim`, failed because such a space already exists.

In addition to the screen shot that shows spaces already in place, `histogram.test.js/1/logs/zqd-access.log` shows requests being handled well after its test was finished.

Timestamps in console log:
```
[2020-06-25 16:16:53.633 info]: Running test: histogram deep inspection
PASS itest/tests/histogram.test.js (54.091s)
  Histogram tests
    ✓ histogram deep inspection (48941ms)

[2020-06-25 16:17:17.373 info]: Running test: FilterEq ensure ingest
```
Last request served by `histogram.test.js/1` `zqd`:
```
$ zq -f text -i ndjson 'last(ts)' histogram.test.js/1/logs/zqd-access.log
1593102647.54996
$ date -u -r 1593102647
Thu Jun 25 16:30:47 UTC 2020
```
The next question is why was `zqd` still up?

```
$ cat histogram.test.js/1/logs/zqd-core.log
{"level":"info","ts":1593101788.4474292,"msg":"Starting","datadir":"/var/tmp/brimsec/itest/histogram.test.js/1/data/spaces","pprof_routes":false,"zeek_supported":true}
{"level":"info","ts":1593101788.457112,"logger":"httpd","msg":"Listening","addr":"127.0.0.1:9867"}
{"level":"info","ts":1593101809.98069,"msg":"Starting","datadir":"/var/tmp/brimsec/itest/histogram.test.js/1/data/spaces","pprof_routes":false,"zeek_supported":true}
```
[...]

PR #901, opened 2020-06-26
Update zq to v0.16.0 and prep v0.12.0

philrz:
https://github.com/brimsec/brim/compare/v0.11.0...be52684


philrz:
@mikesbrown confirmed that the Actions failure appears to be due to #897. He'd already mentioned during a previous incident of it that we shouldn't hold up the release because of it, so I'm proceeding with the merge.

PR #903, opened 2020-06-27
Color fixes and default pane size

jameskerr:
I just wanted to make a few css color updates for the tabs and the welcome backgrounds. I also make the sidebar sections overflow: auto so that scrollbars do not appear if they are not needed.

image

And the welcome screen looks like this:

image

### DIFF
--- a/src/css/_search-page.scss
+++ b/src/css/_search-page.scss
@@ -16,7 +16,7 @@ $page-border-color: $white-4;
 }
 
 .pane-left {
-  min-width: 270px;
+  min-width: 230px;
   flex-shrink: 0;
 }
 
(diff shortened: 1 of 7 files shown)


src/css/_search-page.scss
```diff
@@ -16,7 +16,7 @@ $page-border-color: $white-4;
 }
 
 .pane-left {
-  min-width: 270px;
+  min-width: 230px;
```
jameskerr, line 19: I also thought this was a better size.

src/js/components/common/EmptySection.js
```diff
@@ -8,6 +8,7 @@ const EmptyWrapper = (styled.div`
   flex-direction: column;
   align-items: center;
   margin: 26px 0;
+  user-select: none;
```
mason-fish, line 11: Ah nice, I actually just noticed this issue today 👍

Issue #899, opened 2020-06-26
Store scrollposition in Viewer with searchhistory


r-franke:
In extension of #894 it would be nice if we could store the scroll position alongside the history entry. This would greatly enhance the user experience while they browse 'up' and 'down' through the history to examine data.

**Possible approach** 
`Viewer.js` already has `scrollhooks` in place. So it would be possible to update the scroll position when navigating away from the current page. 

What are you thoughts about this? 

I would be willing to work on this.

jameskerr:
I think this is a great idea, and the scroll hooks would be a perfect place to update the store.

There are two types of history in the app at the moment. The global history in the left pane, and the navigation history for the back and forward buttons next to the search bar. I think this would be a great addition to the navigation history, but I'm not convinced you'd want it in the the global history. For instance, if I pulled up a search I made yesterday, I think I'd want to see the top of my results. What do you think @r-franke? 

@brimsec/brim any of you have any thoughts?

r-franke:
Yeah I fully agree. Recalling a previous search in the left bar should yield fresh results. I did notice during exploration today that there is interaction between the back buttons and the global history by means of highlighting the query you are on. That's something to keep in sync when implementing this.

alfred-landrum:
hi @r-franke : we do think this is a good idea! Let us know if you'd like to work on it, otherwise we'll put it on our todo list and either @jameskerr or @mason-fish will work on it soon.

r-franke:
I'll get started on this.

r-franke:
I think i'm basically done except for the last step I can't seem to figure out.

In `submitSearch` I've leveraged the save routine to save the scrollposition of the current view in `History`:

```javascript
    if (save) {
      let record = Search.getRecord(state)
      let scrollPos = {x: 0, y: 0}
      if (view) {
        scrollPos.x = view.scrollLeft
        scrollPos.y = view.scrollTop
      }
      dispatch(History.push(record, time.toTs(), scrollPos))
      globalDispatch(Investigation.push(record, time.toTs()))
    }
```

Then I forward it to `brim.entries.push` to splice it in:

```javascript
    case "HISTORY_PUSH":
      return brim
        .entries(state)
        .push(action.entry, action.scrollPos)
        .data()
```
So far it's all working and I can see that the scrollpositions are registered on the state. I'm using `SearchBar/flows.js` to restore the scrollPosition:

```javascript
  goBack: (): Thunk => (dispatch, getState) => {
    dispatch(History.back())
[...]

PR #905, opened 2020-07-01
Remember scroll position

r-franke:
Functionality requested in #899 is done.

## Demo
scroll_demo

## Approach
In `submitSearch` I've leveraged the save routine to save the scrollposition of the current view in `History`.

Then I forward it to `brim.entries.push` to splice it in.

And then `Searchbar/flows` is responsible for setting the scrollposition when forward and backward buttons are pressed.

## Tests
Passing

### DIFF
--- a/src/js/brim/entries.js
+++ b/src/js/brim/entries.js
@@ -24,7 +24,11 @@ export default function({
     push(entry: *, scrollPos: *) {
       if (!isEqual(entry, this.getCurrentEntry())) {
         entries.splice(position + 1, entries.length, entry)
-        scrollPositions.splice(position + 1, scrollPositions.length, scrollPos)
+        scrollPositions.splice(position + 1, scrollPositions.length, {
+          x: 0,
+          y: 0
+        })
+        scrollPositions[position] = scrollPos
         position = entries.length - 1
       }
       return this
--- a/src/js/brim/entries.js
+++ b/src/js/brim/entries.js
@@ -4,20 +4,27 @@ import isEqual from "lodash/isEqual"
 
 type Args = {
   position: number,
-  entries: *[]
+  entries: *[],
+  scrollPositions: *[]
 }
 
-export default function({entries: initEntries, position}: Args) {
+export default function({
+  entries: initEntries,
+  position,
+  scrollPositions: initScrollPositions = []
+}: Args) {
   let entries = [...initEntries]
+  let scrollPositions = [...initScrollPositions]
 
   if (position < -1 || position >= entries.length) {
     throw new Error("Position out of bounds")
   }
 
   return {
-    push(entry: *) {
+    push(entry: *, scrollPos: *) {
       if (!isEqual(entry, this.getCurrentEntry())) {
         entries.splice(position + 1, entries.length, entry)
+        scrollPositions.splice(position + 1, scrollPositions.length, scrollPos)
         position = entries.length - 1
       }
       return this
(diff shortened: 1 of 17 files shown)


src/js/brim/entries.js
```diff
@@ -4,20 +4,31 @@ import isEqual from "lodash/isEqual"
...
   return {
-    push(entry: *) {
+    push(entry: *, scrollPos: *) {
       if (!isEqual(entry, this.getCurrentEntry())) {
         entries.splice(position + 1, entries.length, entry)
+        scrollPositions.splice(position + 1, scrollPositions.length, {
+          x: 0,
+          y: 0
+        })
+        scrollPositions[position] = scrollPos
```
r-franke, line 31: This is a point that's up for discussion. Technically it's a side effect, however splitting it off into it's own function leads to a lot of extra code.

src/js/flows/submitSearch.js
```diff
@@ -26,9 +26,16 @@ export default function submitSearch(
 
     const state = getState()
 
+    let view = document.getElementsByClassName("view")[0]
+
     if (save) {
```
jameskerr, line 31: Looking at this file made me realize we may have a bug in the code. If I submit a search with `save: false` it will not update the `History` state slice (which is what is used in the back and forward buttons) or the `Investigation` state slice (which is used for the history in the left sidebar). The main place we use `save: false` is when a user clicks on a previous search in the left sidebar.  When they click a previous search, it will update the searchbar and the viewer, but will not add an entry to the history. This results in:

```
change search input to "a",  hit enter, view results
change search input to "b",  hit enter, view results
click on a previous search in the left side bar, view results
click the "back" button

I'd expect the search input to be "b" but it is "a"
```

To fix this, here's what I think we should do.
* Always run `History.push(record, time.toTs())`
* Only run `Investigation.push(record, time.toTs())` if save === true.

As for making sure that the history entry has the correct scroll position, see my comment about adding a `History.update` call in the onScrollEnd callback in the `Viewer` component.

Note:
There may be a case when we do in fact not want to save to the History or the Investigation. Like when clicking on the "back" and "forward" buttons. If this appears to be true, then we may need to change the save argument consider both object like: `submitSearch({history: false, investigation: false})`.

src/js/flows/submitSearch.js
```diff
@@ -26,9 +26,16 @@ export default function submitSearch(
 
     const state = getState()
 
+    let view = document.getElementsByClassName("view")[0]
```
jameskerr, line 29: When using react, it's considered good practice to avoid grabbing an element like this. Instead, they would say set some state (like `Viewer.setScroll`), then have the component containing that element "react" to the state change.

And to get the x, y values. Set the state in the onScrollEnd callback in the Viewer.

This declarative style of building UI makes reasoning about things easier. However there are situations when it makes sense to break out of the react paradigms and use DOM methods like this.

src/js/state/History/types.js
```diff
@@ -1,11 +1,12 @@
...
 
-import type {SearchRecord} from "../../types"
+import type {ScrollPosition, SearchRecord} from "../../types"
 import type {Ts} from "../../brim"
 
 export type HistoryState = {
   position: number,
-  entries: SearchRecord[]
+  entries: SearchRecord[],
+  scrollPositions: ScrollPosition[]
```
jameskerr, line 9: As mentioned earlier, I think it would be better to add a `scrollPos` property to the `SearchRecord` objects, instead of adding another array here.

r-franke:
I will be submitting a new pull request based from the current master. This branch is getting to polluted.

PR #913, opened 2020-07-09
excess refresh fix part 2

### LINKED ISSUE #904: Excess "Refresh" pop ups
Repro is with Brim commit `5dc9a26` talking to `zqd` commit `b0c3b30`.

See the attached video. To repro starting from a "fresh" environment:

1. Drag a larger pcap into the first tab
2. While the pcap is being imported, perform a search, which will get you into the mode with the pop-up "More data is now available [Refresh]" as the import continues
3. Drag a small pcap into a second tab and wait for it to fully import
4. Come back to the first tab and wait for import to completely finish
5. Now with all data imported, hit the "Refresh" button for what should be the final time
6. Still in the first tab, click on the name of the Space for the small pcap that I'd imported into the second tab

As shown in the attached video, when I click on that Space, another "Refresh" pop-up appears, which I did not expect to see because I'd witnessed that data being fully imported, so the "More data is now available" is inaccurate. In fact, as I continue to perform searches and click back and forth between the two Spaces in the same tab, I've found I can make the "Refresh" pop-up keep coming back. However, I've not tried to capture every combination in the video, as I assume that once we've figured out that first one, the rest will go away as well.

[Repro.mp4.zip](https://github.com/brimsec/brim/files/4847048/Repro.mp4.zip)

mason-fish:
fixes #904 

...Or is another attempt at it. Highly encourage folks to check this out and follow the repro steps in the ticket. So far it has been working for me!

### DIFF
--- a/src/js/components/IngestRefresh.js
+++ b/src/js/components/IngestRefresh.js
@@ -56,7 +56,7 @@ export default function IngestRefresh() {
   useEffect(() => {
     if (currentSpace.id !== space.id) {
       cancelSnapshot()
-      ack()
+      setSnapshotAck(currentSpace.ingest.snapshot)
     }
     setSpace(currentSpace)
   }, [currentSpace])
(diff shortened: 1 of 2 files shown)


philrz:
I did a few repro attempts with @mason-fish's branch and the issue appears fixed from my perspective. 👍

PR #925, opened 2020-07-14
Column chooser makeover

### LINKED ISSUE #873: Can't maintain column headers on ZNG with diverse records
Per the attached video, as of Brim commit `e7b06f3` I can select column headers for records that contain diverse fields, e.g. `uri` (present in Zeek `http` but not in `dns`) and `query` (present in Zeek `dns` but not in `http`). Thanks to the fix in https://github.com/brimsec/zq/issues/852, I can now successfully Export the data to ZNG, then re-import it to a separate tab. The re-imported data starts out with no column headers, which I know to expect because the data shown requires multiple descriptors, and we always start out without column headers in that case. However, if I go into the column picker and start clicking off field names, it successfully adds the headers until I click on the last one, at which point it goes back to showing no headers at all.

Based on what I know of the implementation, I _understand_ why it's doing what it's doing, but since the user was able to see the headers pre-export, it feels like a mismatch that they have no way to see them after exporting and re-importing what the user sees as equivalent data. Do we need to change something related to how "Show All" works to cover this case?

[Repro.zip](https://github.com/brimsec/brim/files/4772661/Repro.zip)

mason-fish:
fixes #873 

Restyles the column chooser, and adds the `select all`, `deselect all`, and column view selector. The latter allows for the existing behavior to continue (and is the default), where column headers will automatically be shown or hidden based on the data, but the user may also explicitly set the headers on or off if they want.

image

image

### DIFF
--- a/src/js/components/ColumnChooserMenu.js
+++ b/src/js/components/ColumnChooserMenu.js
@@ -11,17 +11,17 @@ import CloseButton from "./CloseButton"
 import Columns from "../state/Columns"
 import TableColumns from "../models/TableColumns"
 import dispatchToProps from "../lib/dispatchToProps"
-import Viewer from "../state/Viewer"
 import styled from "styled-components"
 import Checkbox from "./common/Checkbox"
-import type {ColumnHeadersViewState} from "../state/Viewer/types"
 import SelectInput from "./common/forms/SelectInput"
+import Layout from "../state/Layout"
+import type {ColumnHeadersViewState} from "../state/Layout/types"
 
 const ControlListItem = styled.li`
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 4px 36px;
+  padding: 4px 12px;
   cursor: default;
 
   p {
@@ -34,7 +34,7 @@ const ColumnListItem = styled.li`
   display: flex;
   justify-content: space-between;
   align-items: center;
-  padding: 4px 36px;
+  padding: 4px 12px;
   cursor: default;
   ${(props) => props.theme.typography.labelNormal}
 `
@@ -100,7 +100,7 @@ export default class ColumnChooserMenu extends React.Component<Props> {
     const count = this.props.tableColumns.visibleCount()
 
     const onChangeColumnView = (e) => {
-      this.props.dispatch(Viewer.setColumnHeadersView(e.target.value))
+      this.props.dispatch(Layout.setColumnHeadersView(e.target.value))
     }
 
     return (
@@ -158,7 +158,7 @@ export default class ColumnChooserMenu extends React.Component<Props> {
 
 const stateToProps = (state: State) => ({
   tableColumns: Columns.getCurrentTableColumns(state),
-  columnHeadersView: Viewer.getColumnHeadersView(state)
+  columnHeadersView: Layout.getColumnHeadersView(state)
 })
 
 export const XColumnChooserMenu = connect<Props, OwnProps, _, _, _, _>(
--- a/src/js/components/ColumnChooserMenu.js
+++ b/src/js/components/ColumnChooserMenu.js
@@ -3,7 +3,6 @@
 import {CSSTransition} from "react-transition-group"
 import {connect} from "react-redux"
 import React from "react"
-import classNames from "classnames"
 
 import type {DispatchProps, State} from "../state/types"
 import {Fieldset, Paragraph, Subscript, Label} from "./Typography"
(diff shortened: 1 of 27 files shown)


src/js/components/ColumnChooserMenu.js
```diff
@@ -12,13 +11,66 @@ import CloseButton from "./CloseButton"
...
+const ControlListItem = styled.li`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 4px 36px;
+  cursor: default;
+
+  p {
+    text-decoration: underline;
+    color: ${(props) => props.theme.colors.havelock};
```
jameskerr, line 30: You know, for colors, it would be really cool to use css variables.

```css
color: var(--havelock)
```

src/js/components/ColumnChooserMenu.js
```diff
@@ -36,66 +88,90 @@ export default class ColumnChooserMenu extends React.Component<Props> {
...
     if (column.isVisible) {
-      if (this.allVisible()) {
-        this.props.dispatch(Columns.hideAllColumns(this.props.tableColumns))
-        this.props.dispatch(Columns.showColumn(this.tableId(), column))
-      } else if (this.props.tableColumns.visibleCount() === 1) {
-        this.props.dispatch(Columns.showAllColumns(this.props.tableColumns))
-      } else {
-        this.props.dispatch(Columns.hideColumn(this.tableId(), column))
-      }
+      this.props.dispatch(Columns.hideColumn(this.tableId(), column))
```
jameskerr, line 104: 👍

src/js/components/ColumnChooserMenu.js
```diff
@@ -36,66 +88,90 @@ export default class ColumnChooserMenu extends React.Component<Props> {
...
+      {
+        label: "Auto",
+        click: () => this.props.dispatch(Viewer.setColumnHeadersView("AUTO"))
+      }
+    ]
+
+    const openMenu = usePopupMenu(template)
+    const onClick = () => {
+      openMenu()
+    }
```
jameskerr, line 133: I'd suggest using a simple `<select>` input for this. There is a styled version in the `common/forms/SelectInput.js` file.  It might look something like this:

```jsx
const view = useSelector(Viewer.getColumnHeadersView)
const onChange = (e) => dispatch(Viewer.setColumnHeadersView(e.target.value))

<SelectInput value={view} onChange={onChange}>
  <option value="ON">On</option>
  <option value="OFF">Off</option>
  <option value="AUTO">Auto</option>
</SelectInput>
```

src/js/components/ColumnChooserMenu.js
```diff
@@ -36,66 +88,90 @@ export default class ColumnChooserMenu extends React.Component<Props> {
...
-            </li>
+            <ControlListItem>
+              <StyledLabel>Headers</StyledLabel>
+              <StyledViewSelect onClick={onClick}>
+                {columnHeadersView.toLowerCase()}
+                <DropdownArrow />
+              </StyledViewSelect>
+            </ControlListItem>
+            <ControlListItem>
+              <Paragraph onClick={this.selectAllColumns}>Select All</Paragraph>
```
jameskerr, line 161: I'd like to deprecate this old `<Paragraph />` component. These would be "Label Small"

src/js/components/LogRow.js
```diff
@@ -35,55 +34,28 @@ export default class LogRow extends React.Component<Props> {
...
-      </div>
-    )
-  }
-
-  renderFixedLayout() {
-    const {highlight, columns, log, dimens, index, rightClick} = this.props
+  render() {
+    const {dimens, highlight, index, log, rightClick, columns} = this.props
     const renderCell = (column, colIndex) => {
+      const width = dimens.rowWidth !== "auto" ? column.width || 300 : "auto"
```
jameskerr, line 40: Love how much more simple this got.

src/js/components/Toggle.js
```diff
@@ -13,7 +13,7 @@ export default class Toggle extends React.Component<Props> {
         onClick={this.props.onChange}
         // Pass-through props didn't work here. I had to set this directly.
         // Feel free to improve.
-        data-test-locator={this.props.dataTestLocator}
+        {...this.props}
```
jameskerr, line 16: Maybe remove that comment if it is no longer relevant.

src/js/components/common/Checkbox.js
```diff
@@ -0,0 +1,74 @@
...
+  height: 16px;
+  background: ${(props) =>
+    props.checked ? props.theme.colors.havelock : props.theme.colors.white};
+  border-radius: 2px;
+  transition: all 150ms;
+  border: 1px solid
+    ${(props) =>
+      props.checked ? props.theme.colors.white : props.theme.colors.lead};
+
+  ${Icon} {
```
jameskerr, line 49: wow, cool

src/js/components/common/Checkbox.js
```diff
@@ -0,0 +1,74 @@
...
+  ${(props) => props.theme.typography.labelSmall}
+`
+
+const StyledCheckbox = styled.div`
+  display: inline-block;
+  width: 16px;
+  height: 16px;
+  background: ${(props) =>
+    props.checked ? props.theme.colors.havelock : props.theme.colors.white};
+  border-radius: 2px;
```
jameskerr, line 43: I think a border-radius: 3px looks better now that I see it.

src/js/electron/extensions.js
```diff
@@ -3,22 +3,24 @@
...
+
+  install(REACT_DEVELOPER_TOOLS)
+    .then((name) => console.log(`Added Extension:  ${name}`))
+    .catch((err) => console.log("An error occurred: ", err))
+
+  install(REDUX_DEVTOOLS)
+    .then((name) => console.log(`Added Extension:  ${name}`))
+    .catch((err) => console.log("An error occurred: ", err))
+  BrowserWindow.removeDevToolsExtension("React Developer Tools")
+  BrowserWindow.removeDevToolsExtension("Redux DevTools")
```
jameskerr, line 23: 🤘

src/js/state/Viewer/types.js
```diff
@@ -19,7 +21,8 @@ export type ViewerState = {|
   status: SearchStatus,
   stats: ViewerStats,
   scrollX: number,
-  scrollY: number
+  scrollY: number,
+  columnHeadersView: ColumnHeadersViewState
```
jameskerr, line 25: You know, I think this should actually go in the global "user prefs" area of the state. In the future, I'd like not to persist things that seem temporal like the state of the viewer. This way we'd have to write fewer migrations. However, the column headers setting seems like it would be something we'd want to save between sessions. And it seems like something that should be either at the global or tab level. I think tab level would be good for now. Can we save it it the Layout slice?

src/js/components/ColumnChooserMenu.js
```diff
@@ -12,13 +11,66 @@ import CloseButton from "./CloseButton"
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 4px 36px;
+  cursor: default;
+  ${(props) => props.theme.typography.labelNormal}
+
+  &:hover {
+    background-color: ${(props) => props.theme.colors.ivory};
+  }
```
jameskerr, line 44: Now that is a checkbox, I think the hover state is a bit distracting. I'd be fine removing it.

src/js/components/ColumnChooserMenu.js
```diff
@@ -12,13 +11,66 @@ import CloseButton from "./CloseButton"
...
+    text-decoration: underline;
+    color: ${(props) => props.theme.colors.havelock};
+  }
+`
+
+const ColumnListItem = styled.li`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 4px 36px;
```
jameskerr, line 38: I think `12px` padding on left and right would look better for the items and the links.

src/js/state/Viewer/types.js
mason-fish: The current behavior is that this is stored per tab and persisted in the state (and I agree that we should be cautious adding stuff there to avoid maintenance complexity). Is that what you had in mind?

I did notice that we store the `Viewer` slice of state at both the root level, representing the current tab's view, and also nested in each of the Tabs states. As far as this component is concerned, the selectors are all using the nested Viewer state under Tabs, specifically the active tab, to read data here. I was actually curious what we use the root level Viewer for?

PR #938, opened 2020-07-22
Update npm dependencies

jameskerr:
Time to update some minor and patch versions.


package.json
```diff
@@ -63,94 +63,94 @@
...
+    "babel-eslint": "^10.1.0",
     "babel-jest": "^24.9.0",
-    "babel-plugin-inline-react-svg": "^1.1.0",
-    "babel-plugin-source-map-support": "^2.1.1",
+    "babel-plugin-inline-react-svg": "^1.1.1",
+    "babel-plugin-source-map-support": "^2.1.2",
     "commander": "^2.20.3",
     "cpx": "^1.5.0",
-    "electron": "^7.1.9",
+    "electron": "^7.3.2",
```
marktwallace, line 81: This is a super-generic question, but what are the reasons to stay on Electron 7, rather than move to 8 or 9? Are there upward compatibility problems with the move?

jameskerr: Nope, I don't have any reason to stay on Electron 7. I just looked through the electron releases pages and didn't find any obvious breaking changes that would effect our app. I'll upgrade electron to the latest version in this PR.

jameskerr:
@marktwallace good question. We have a suite of integration tests that test the important parts of the app. It's similar to selenium tests but for Electron apps. The library is called Spectron. If anything introduced here causes a dramatic breakage, those tests should pick it up. It was already run on this branch and passed it looks like.

To run the integration tests locally, you can `npm run itest`.

jameskerr:
As it turns out, updating to electron involves a few changes to the integration tests. Webdriverio version 6 has several breaking changes that we'll have to fix. I've opened an issue here. It should'nt be too bad, but I think it's best to punt it for now.
https://github.com/brimsec/brim/issues/952

PR #939, opened 2020-07-23
Mention logs location in Troubleshooting doc

philrz:
While working with a community user on the problem that led to opening https://github.com/brimsec/zq/issues/1039, I realized that we don't currently have any Support guidance as to where to find the logs that are generated while running Brim. These logs are admittedly a little sparse at the moment (i.e. they don't happen to give any insight into what's going on during a repro of https://github.com/brimsec/zq/issues/1039), but hopefully this will improve over time. Therefore in this PR I've added detail in the Troubleshooting doc to point the user at their location and ask that they be sent along when opening bug issues.


docs/Troubleshooting.md
```diff
@@ -170,6 +171,20 @@ Now perform the operation in the app that triggers the problem. Hopefully this
 will reveal some error messages in the **Console** area of the panel. If this
 is the case, please take a screenshot and include it with your issue.
 
+## Logs
+
+Logs are generated by the Brim app and the `zqd` process it launches to act as
```
jameskerr, line 176: Maybe...

> Logs are generated by the Brim app and the `zqd` process it launches (the engine for querying your data).

philrz: @jameskerr: Yeah, I agree that's an improvement. I've pushed a commit with that change. Thanks for the approval!

PR #951, opened 2020-07-27
Fix deleting a history item issuing a search

### LINKED ISSUE #947: Deleting a History entry also executes it
Repro is with Brim commit `fce9a46` and test data [wrccdc.2018-03-23.010014000000000.pcap.gz](https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz).

In the attached video, I've got two entries in my History panel. One was a query that returns results (`ocsp.pki.goog`) and the other contains a typo and returns nothing (`ocsup.pki.goog`). As you can see, if I have the results for the working query on my screen and then realize I want to delete the History entry for the one with the typo, and hence click its "X", it actually re-executes that query and shows me a blank set of results, which is not what I was expecting as a user.

[Repro.zip](https://github.com/brimsec/brim/files/4974466/Repro.zip)

jameskerr:
Now when you click the "x" on a history entry, it removes that entry and does not issue a search. Fixes #947 

The issue was that a nested click handler was not stopping the event propagation.

Unit tests have been added to ensure this doesn't regress.

### DIFF
--- a/src/js/components/Investigation/FindingCard.test.js
+++ b/src/js/components/Investigation/FindingCard.test.js
@@ -1,0 +1,43 @@
+/* @flow */
+
+import React from "react"
+
+import FindingCard from "./FindingCard"
+import brim from "../../brim"
+import loginTo from "../../test/helpers/loginTo"
+import provide from "../../test/helpers/provide"
+
+let store
+beforeEach(async () => {
+  const setup = await loginTo("cluster1", "space1")
+  store = setup.store
+})
+
+function getActionTypes() {
+  return store.getActions().map((a) => a.type)
+}
+
+const finding = {
+  ts: brim.time().toTs(),
+  search: {
+    program: "finding card test",
+    pins: [],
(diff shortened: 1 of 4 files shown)


src/js/components/Investigation/FindingCard.test.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
+
+import React from "react"
+
+import FindingCard from "./FindingCard"
+import brim from "../../brim"
+import loginTo from "../../test/helpers/loginTo"
+import provide from "../../test/helpers/provide"
+
+let store
```
mattnibs, line 10: Isn't it better to stray away from using a global here so these tests can run in parallel?

jameskerr: The tests do run in parallel at the file level, instead of the tests level. This pattern is used throughout the tests and works well enough.

PR #953, opened 2020-07-28
Column Visibility Defaults

jameskerr:
Now the logic is:

If all of the columns in the "set" are visible, any columns that come along without an explicit setting are "visible".
If NOT all of the columns in the "set" visible (only some are), then any columns without a setting are "not visible".

This replicates the behavior we had prior to the redesign.

JoZGtaowi2

### DIFF
--- a/src/js/models/TableColumns.js
+++ b/src/js/models/TableColumns.js
@@ -16,21 +16,23 @@ export default class TableColumns {
     columns: Column[] = [],
     tableSetting: ColumnSettingsMap = {}
   ) {
+    // $FlowFixMe
+    const defaultVisible = Object.values(tableSetting).every((c) => c.isVisible)
     this.id = id
     this.cols = columnOrder(columns)
       .map((col, index) => ({
         ...col,
-        ...TableColumns.columnDefaults(index),
+        ...TableColumns.columnDefaults(index, defaultVisible),
         ...tableSetting[columnKey(col)]
       }))
       .sort((a, b) => (a.position > b.position ? 1 : -1))
   }
 
-  static columnDefaults(index: number) {
+  static columnDefaults(position: number, isVisible: boolean) {
     return {
       width: undefined,
-      isVisible: true,
-      position: index
+      isVisible,
+      position
     }
   }
 


philrz:
I know this still needs code review and doesn't appear to be passing CI yet, but just chiming in to say that I tested it out on the branch and from a functional perspective it seems to be addressing the unexpected behaviors. 👍

src/js/models/TableColumns.js
```diff
@@ -16,21 +16,23 @@ export default class TableColumns {
     columns: Column[] = [],
     tableSetting: ColumnSettingsMap = {}
   ) {
+    // $FlowFixMe
+    const defaultVisible = Object.values(tableSetting).every((c) => c.isVisible)
```
mason-fish, line 20: I had tried this, but if you open the column chooser and deselect an option then it will deselect all of them since this will return false for every one

jameskerr: Thanks for the heads up @mason-fish. You were correct. The new commits fix this by immediately adding any columns from a search results to the "Column Preferences" state with sane defaults. Then the "select all" and deselect all" operate on all columns known to the preferences. Not just the columns that are in the current search results. Hope that makes sense.

src/js/state/Columns/touch.test.js
```diff
@@ -0,0 +1,70 @@
...
+  "1": [
+    {name: "_path", type: "string"},
+    {name: "duration", type: "interval"}
+  ],
+  "2": [
+    {name: "_path", type: "string"},
+    {name: "ts", type: "time"}
+  ]
+}
+let store
```
mattnibs, line 18: If you are using a global, doesn't this mean that the tests in this file cannot be run in parallel?

jameskerr: Correct, the tests cases in this file are not run in parallel. With Jest, each individual file gets it's own process and global state.

Here's a line from their homepage:

> By ensuring your tests have unique global state, Jest can reliably run tests in parallel.

https://jestjs.io/

mattnibs: ok

PR #956, opened 2020-07-28
Update devtools installer

### LINKED ISSUE #919: Uncomment Existing Devtools Code
Yesterday in the live stream we briefly discussed the dev tools. They where disabled because they may have caused performance issues at some times according to @jameskerr .

Maybe create a separate npm run scrip to enable the devtools. They aren't needed all the time but are sometimes handy to have up for grabs.

What are your thoughts?

jameskerr:
Fixes #919 

Update the dev tools installer package to the latest version. The following devtools should exist only in development mode.

image


philrz:
FWIW, I gave this a peek on the branch and I think I see all the expected additional menus. There's the React ones for Components and Profiler:

image

image

Then there's the Redux one. I had to expand the window to make it easier to find.

image

mattnibs:
Are the test failures on related to this change?

src/js/electron/extensions.js
```diff
@@ -1,26 +1,13 @@
...
-
-  install(REDUX_DEVTOOLS)
-    .then((name) => console.log(`Added Extension:  ${name}`))
-    .catch((err) => console.log("An error occurred: ", err))
-  BrowserWindow.removeDevToolsExtension("React Developer Tools")
-  BrowserWindow.removeDevToolsExtension("Redux DevTools")
+export function installExtensions() {
+  return installExtension([REDUX_DEVTOOLS, REACT_DEVELOPER_TOOLS])
+    .then(() => log.info("Devtools loaded"))
+    .catch((err) => log.error("An error occurred: ", err))
```
mattnibs, line 12: Perhaps this is clearer?
```suggestion
    .catch((err) => log.error("Error loading devtools: ", err))
```

PR #957, opened 2020-07-29
Wiki update for new Geolocation functionality

philrz:
In addition to just mentioning it in release notes, I'd like to get into the habit of adding wiki articles for significant new functionality. In this case I was also motivated by the fact that I have ideas of ways we could go further with geolocation functionality, but I'm not inclined to push for it unless users can really imagine themselves using it. Therefore, this article also serves as a way to see if users are noticing and paying enough attention to review & click on the issues.

If you want to see how it looks rendered in a wiki, see:

https://github.com/philrz/scratchwiki/wiki/Geolocation


docs/Geolocation.md
```diff
@@ -0,0 +1,42 @@
+# Geolocation
+
+- [Summary](#summary)
+- [Examples](#examples)
+- [Origin](#origin)
+- [Future Functionality](#future-functionality)
+
+# Summary
+
+Starting with `v0.14.0`, Brim now adds [geolocation](https://en.wikipedia.org/wiki/Geolocation) fields to the `conn` records
```
alfred-landrum, line 10:
```suggestion
Starting with `v0.14.0`, Brim adds [geolocation](https://en.wikipedia.org/wiki/Geolocation) fields to the `conn` records
```

docs/Geolocation.md
```diff
@@ -0,0 +1,42 @@
...
+of the geolocation data, see the [README](https://github.com/brimsec/geoip-conn/blob/master/README.md).
+
+# Future Functionality
+
+There are additional geolocation features in Brim that may be added in the future, depending on demand from the community.
+The following issues are currently being held open to gather interest:
+
+| **Issue**                                              |**Description**                                 |
+|--------------------------------------------------------|------------------------------------------------|
+| [brim/936](https://github.com/brimsec/brim/issues/936) | GeoLocation map visualization                  |
```
alfred-landrum, line 35: Are the dual 'G' & 'L' capitals of `GeoLocation` intended?

philrz: Fair point. I seem to have gotten into a bad habit of sometimes typing "GeoLocation", which I think was a side effect of typing "GeoLite2", which does, indeed, have the capital "L" in the middle. However, you're right that "geolocation" is not typically written that way, so I've fix it in my issue descriptions/comments and here in this branch. Thanks for flagging!

PR #961, opened 2020-07-31
Update zq to v0.18.0 and prep v0.14.0

philrz:
https://github.com/brimsec/brim/compare/v0.13.1...d88d1d4a128b90e29edcc93d4fe3313974311f82

There also were some packages being accessed via `http://`, and when I did `npm install https://github.com/brimsec/zq#v0.18.0` to update the `zq` pointer, it corrected those to `https://` as well. Since I know we'd been standardizing on `https://` in the past for obvious reasons, I let those changes roll into here as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## v0.14.0
 * Update zq to [v0.18.0](https://github.com/brimsec/zq/releases/tag/v0.18.0)
 * Add [geolocation](https://github.com/brimsec/brim/wiki/Geolocation) data to Zeek `conn` logs generated from imported pcaps (#959, #957, #935)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## v0.14.0
+* Update zq to [v0.18.0](https://github.com/brimsec/zq/releases/tag/v0.18.0)
+* Add [geolocation](https://github.com/brimsec/brim/wiki/Geolocation) data to Zeek `conn` logs generated from imported pcaps (#959, #957, #935)
+* Add developer documentation for [adding migrations](https://github.com/brimsec/brim/wiki/Adding-Migrations) (#921)
```
alfred-landrum, line 4: Could you call these 'internal state migrations' or something, to reinforce that this is not a user concern?

philrz: Sure thing. Went with your suggestion.

Issue #854, opened 2020-06-02
Add Index Search To Zealot


jameskerr:
The zealot client will need to be updated to support an index search.

Issue #853, opened 2020-06-02
Create Index Search Page


jameskerr:
Implement the design for a space that has a type of "archive". Keep in mind that we'll want to toggle back and forth between searching the data and searching the indexes.

alfred-landrum:
Closing this as we're no longer thinking of having separate search flows for data vs indexes.

PR #963, opened 2020-07-31
Create Space Form & Left Pane Polish

jameskerr:
A first step of implementing the UI for searching a ZAR archive. 

Part of #853 

Closes: #854 

* Add a form to create a space that points to a ZAR archive.
Screen Shot 2020-07-31 at 11 10 48 AM

* Add a box icon in the left side to indicate a ZAR archive.
Screen Shot 2020-07-31 at 11 08 07 AM

* Post a "Coming soon" message when a user selects that space.
Screen Shot 2020-07-31 at 11 07 54 AM

This PR also polishes the styles in the left side bar.

* Add a "plus" icon next the the spaces header to bring you back to the "new tab" page.Screen Shot 2020-07-31 at 11 08 28 AM
*  No more border radius or margins on the list items
Screen Shot 2020-07-31 at 11 09 01 AM

* Consistent hover state and active state
Screen Shot 2020-07-31 at 11 10 29 AM

* Consistent right click to remove behavior
Screen Shot 2020-07-31 at 11 16 47 AM

* A click region that is larger for expanding and collapsing the sections
Screen Shot 2020-07-31 at 11 10 24 AM

Since we don't really want to promote something that is not yet finished, I made the link to the new space form small and subtle at the bottom of the "new tab page".

aVPG4hMlo3


src/js/components/LeftPane.js
```diff
@@ -43,42 +45,73 @@ const SectionContents = styled.div`
 const SectionHeader = styled.div`
   display: flex;
   flex-direction: row;
-  background-color: ${(props) => props.theme.colors.coconut};
+  background-color: var(--coconut);
```
mason-fish, line 48: I do like how this simplifies how we select the color and gets rid of all the darn `${(props) => props.theme.finallySelectTheDamnColor}` boiler plate!

On the other hand, styled-components seems to really embrace defining css within js and this would mix the two paradigms a little. What if we just export the whole theme object and then import as needed? I also like being able to easily click into variables, which I don't think would work with css variables

Another thought (which doesn't yet apply to us, contradicts my last comment, but is worth thinking about for the future), is that the functionality that themes are supposed to provide is a restriction of sorts. So that certain components which inherit a given theme, can only use what's provided in that theme, while other components inherit different themes. Currently everything is already, and has been, global for us but it's just something to consider.

src/js/components/SavedSpacesList.js
```diff
@@ -37,33 +38,40 @@ export default function SavedSpacesList({spaces, spaceContextMenu}: Props) {
...
+                onClick={onClick(s.id)}
+                onContextMenu={() => {
+                  !s.ingesting() &&
+                    showContextMenu(spaceContextMenu(s.id, s.name))
+                }}
+                className={classNames("space-link", {
+                  "current-space-link": s.id === currentSpaceId
+                })}
+              >
+                {s.storage_kind === "archivestore" ? (
```
mason-fish, line 63: 👍

mason-fish: I'm predicting we will want to refactor this a little once we add dropdown functionality and hierarchy to the spaces

src/js/components/TabArchiveSearch.js
```diff
@@ -0,0 +1,83 @@
...
+        <SearchWrapper>
+          <form onSubmit={onSubmit}>
+            <SearchInput autoFocus value={pattern} onChange={onChange} />
+          </form>
+        </SearchWrapper>
+      </Header>
+      <ResultsWrapper>
+        {/* <SearchResults /> */}
+        <InfoCard>
+          <p>Archive search is comming soon.</p>
```
mason-fish, line 75: s/comming/coming/

src/js/components/TabCreateSpace.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+`
+
+const SubmitWrap = styled.div`
+  margin-top: 18px;
+  display: flex;
+  justify-content: flex-end;
+`
+
+export default function TabCreateSpace() {
```
mason-fish, line 35: This is a great new component and flow for space creation! Why don't combine the import files flow in here as well? We could still provide defaults for the name so that the user still only needs to import the logs, click `create` and go.

src/js/components/common/buttons/PrimaryButton.js
```diff
@@ -0,0 +1,33 @@
+/* @flow */
+import {darken} from "polished"
+import styled from "styled-components"
+
+import type {Styled} from "../../../types/styled"
+
+const bg = "#4b91e2"
```
mason-fish, line 7: Does this "magic color" correspond to anything we use anywhere else?

src/js/flows/initSpace.js
```diff
@@ -1,72 +1,34 @@
...
-    .then(checkSpacesExist)
-    .then((spaces) => getCurrentSpaceId(spaces, desiredID))
-    .then((spaceId) => zealot.spaces.get(spaceId))
-    .then((data) => setSpace(dispatch, data, clusterId))
-    .then((data) => setSearchDefaults(dispatch, data))
-    .then((data) => checkDataExists(dispatch, data, tabId))
-    .then(() => {
+    .get(spaceId)
+    .then(brim.interop.spacePayloadToSpace)
+    .then((space) => {
```
mason-fish, line 22: oh nice! I wonder if there was a noticeable speed bump from this

src/js/flows/submitArchiveSearch.js
```diff
@@ -0,0 +1,94 @@
...
+import Handlers from "../state/Handlers"
+import Notice from "../state/Notice"
+import SearchBar from "../state/SearchBar"
+import Tab from "../state/Tab"
+import Tabs from "../state/Tabs"
+import Viewer from "../state/Viewer"
+import brim from "../brim"
+import whenIdle from "../lib/whenIdle"
+
+export default (patterns: string[]): Thunk => (
```
mason-fish, line 13: 🔥

src/js/components/LeftPane.js
jameskerr: In addition to the reasons you found, I'm really into css variables for a few reasons:
* They are accessible through javascript and stylesheets. This makes them an excellent bridge between total "css in js" and the standard stylesheet approach. Since we will have both patterns for a while, we can store the values of these variables in one place and use them in both contexts.
* They have live bindings, meaning that changing the value of --coconut in javascript will immediately update the color on the whole page. This makes "themeing" the app really nice. It we wanted to support several "skins" for brim, we could store vars like --bg-color, --text-color, --font-size. The change them all in Javascript and the whole page will update immediatly.
* They have excellent chrome devtools support

Screen Shot 2020-08-04 at 1 02 16 PM


I think we could use the theme for things with more than one property, and use css vars with things that are only shared values.

Does that make sense?

src/js/components/TabArchiveSearch.js
jameskerr: Thanks! Will fix.

src/js/components/TabCreateSpace.js
jameskerr: In the future, I think an approach like you suggested here will make a lot of sense. Since it's a different scope that the "index search" feature, we'll address it in the future.

src/js/components/common/buttons/PrimaryButton.js
jameskerr: Oops, yeah I need to fix that.

jameskerr:
After the desktop Jam session, we added some to dos to this PR:

- [ ] Hide the create empty space menu if we're not developing
- [ ] Fix the magic bg color
- [ ] Make it so that switching tabs does not issue another search

Issue #926, opened 2020-07-14
design subspace creation from search tabular results


alfred-landrum:
From an index search, we want a user to be able to choose one or more records from the tabular search results, and create a subspace where they can search the full data. We've discussed allowing control-clicking to choose records, or making checkboxes available, but haven't narrowed down to a specific design yet.

philrz:
Verified in Brim commit `76f506a` talking to `zqd` commit `42103ef`.

First I followed the steps from the [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md) to create an archive, then used `zapi` to create an Archive Space for it in my locally-running Brim app.

```
$ zq zng/*.gz | zar import -s 25MB -

$ zar index :ip
file:///Users/phil/logs/20180324/1521912990.158766.zng: creating index file:///Users/phil/logs/20180324/1521912990.158766.zng.zar/microindex-type-ip.zng
file:///Users/phil/logs/20180324/1521912507.399929.zng: creating index file:///Users/phil/logs/20180324/1521912507.399929.zng.zar/microindex-type-ip.zng
file:///Users/phil/logs/20180324/1521912075.114273.zng: creating index file:///Users/phil/logs/20180324/1521912075.114273.zng.zar/microindex-type-ip.zng
file:///Users/phil/logs/20180324/1521911772.980384.zng: creating index file:///Users/phil/logs/20180324/1521911772.980384.zng.zar/microindex-type-ip.zng

$ zar index uri
file:///Users/phil/logs/20180324/1521912990.158766.zng: creating index file:///Users/phil/logs/20180324/1521912990.158766.zng.zar/microindex-field-uri.zng
file:///Users/phil/logs/20180324/1521912507.399929.zng: creating index file:///Users/phil/logs/20180324/1521912507.399929.zng.zar/microindex-field-uri.zng
file:///Users/phil/logs/20180324/1521912075.114273.zng: creating index file:///Users/phil/logs/20180324/1521912075.114273.zng.zar/microindex-field-uri.zng
file:///Users/phil/logs/20180324/1521911772.980384.zng: creating index file:///Users/phil/logs/20180324/1521911772.980384.zng.zar/microindex-field-uri.zng

$ zapi new -k archivestore -d $ZAR_ROOT MyArchive
MyArchive: space created
```

[...]

PR #971, opened 2020-08-05
Viewer Multi Select

jameskerr:
Relates to #926 

The viewer is now able to select rows via "click", "cmd+click", and "shift+click".

image

Since multi-select doesn't make sense in the normal log search viewer at the moment (it would just create more work that is not priority), we will disable it there. It will be enabled in the archive search results table.

Things to finish before merging:

- [x] Adding a handler for "select all"
- [x] Changing the background color of the selected cell
- [x] Disabling it on the main viewer
- [x] Writing migration to remove "viewer" slice of state from being persisted
- [x] Removing "viewer" slice in the root reducer (it's just hanging out there doing nothing)


src/js/state/Tabs/reducer.js
```diff
@@ -119,3 +120,14 @@ function updateTab(state, action: Object) {
 function indexOf(data, id) {
   return data.findIndex((t) => t.id === id)
 }
+
+function reduxAction(action) {
+  return action.type.startsWith("@@")
```
mason-fish, line 125: I wasn't aware that this is associated with an internal redux event. Would you mind adding a little comment here providing some color on what this is doing, for posterity?

src/js/state/Viewer/reducer.js
```diff
@@ -1,25 +1,31 @@
...
+  scrollPos: {x: 0, y: 0},
+  selection: {
+    rows: {},
+    currentRange: [0, 0]
+  }
+})
 
 export default function(
-  state: ViewerState = init,
+  state: ViewerState = init(),
```
mason-fish, line 21: 👍 In addition to being careful about ensuring initial state does not reuse a pointer to an old object, the changes here are reminding me that we should prob choose to continue to adopt `immer` with the same policy we've decided for `styled-components`

src/js/state/Viewer/actions.js
```diff
@@ -8,42 +8,88 @@ import type {
...
+
+export const selectMulti = (index: number): VIEWER_SELECT_MULTI => {
+  return {type: "VIEWER_SELECT_MULTI", index}
+}
+
+export const selectRange = (index: number): VIEWER_SELECT_RANGE => {
+  return {type: "VIEWER_SELECT_RANGE", index}
+}
+
+export const selectRangeNext = () => ({
```
mason-fish, line 77: Why no return type annotations for the remaining actions here?

src/js/state/Viewer/reducer.js
jameskerr: Yes, that makes sense to me.

src/js/state/Viewer/actions.js
jameskerr: Oops, forgot to add those.

PR #972, opened 2020-08-06
841 remove blank record flash

### LINKED ISSUE #841: Remove brief blank record table periods during import
When a pcap is imported into Brim, if no query is used, Brim updates the records shown as import progress is made. This currently causes brief periods of blank space in the records list while updating to the new set of records available. We should remove these blank periods.

mason-fish:
fixes #841 

Don't update the table unless a different search has been submitted

brim

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@
 /coverage
 /releases
 /zealot
+/run
--- a/src/js/flows/submitSearch.js
+++ b/src/js/flows/submitSearch.js
@@ -8,10 +8,10 @@ import Search from "../state/Search"
 import SearchBar from "../state/SearchBar"
 import Tab from "../state/Tab"
 import Tabs from "../state/Tabs"
-import Viewer from "../state/Viewer"
 import brim from "../brim"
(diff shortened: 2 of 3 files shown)


src/js/flows/submitSearch.js
```diff
@@ -51,3 +51,13 @@ export default function submitSearch(
...
   }
 }
+
+function shouldClearTable(args, prev) {
+  let duration = ([from, to]) => to - from
+
+  return !(
+    args.tableProgram === prev.tableProgram &&
+    args.spaceId === prev.spaceId &&
+    duration(args.span) === duration(prev.span)
```
jameskerr, line 61: Will you also add an `if this space is ingesting` check. I think if the space is not ingesting and I hit "enter", I'd want to see it clear and re-populate. What do you think?

.eslintignore
```diff
@@ -4,3 +4,4 @@
 /coverage
 /releases
 /zealot
+/run
```
jameskerr, line 7: Same thing was happening in one of my branches. Glad you found it too.

PR #973, opened 2020-08-06
Fix double click bug

### LINKED ISSUE #966: Double clicking on two different rows, acts like double clicking on the second row
**Describe the bug**
Quickly clicking two different logs will trigger the "onDoubleClick" handler for the latter log. 

**Expected behavior**
I'd expect that you'd only get the detail view by clicking twice, quickly on the same log line.

mason-fish:
fixes #966 

Track previously selected index, and compare on double click

### DIFF
--- a/src/js/components/LogRow.js
+++ b/src/js/components/LogRow.js
@@ -54,14 +54,7 @@ const LogRow = (props: Props) => {
       )
     }
     if (dimens.rowWidth !== "auto") {
-      return (
-        <div
-          className="log-cell"
-          key={key}
-          style={{width}}
-          onClick={handleClick}
-        />
-      )
+      return <div className="log-cell" key={key} style={{width}} />
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/js/components/LogRow.js
```diff
@@ -19,51 +20,69 @@ type Props = {
...
+
+export default memo<Props>(LogRow, (prevProps: Props, nextProps: Props) => {
+  return !(
+    !Log.isSame(prevProps.log, nextProps.log) ||
+    !isEqual(prevProps.columns, nextProps.columns) ||
+    prevProps.highlight !== nextProps.highlight ||
+    prevProps.dimens.rowWidth !== nextProps.dimens.rowWidth ||
+    prevProps.timeZone !== nextProps.timeZone ||
+    prevProps.timeFormat !== nextProps.timeFormat
+  )
```
jameskerr, line 87: How about inverting this to avoid the double negatives?

```js
return Log.isSame(prevProps.log, nextProps.log) &&
  isEqual(prevProps.columns, nextProps.columns) &&
  prevProps.highlight === nextProps.highlight &&
  prevProps.dimens.rowWidth === nextProps.dimens.rowWidth &&
  prevProps.timeZone === nextProps.timeZone &&
  prevProps.timeFormat === nextProps.timeFormat
```

jameskerr: I'm glad the inverted the API, I think it makes more sense to say, "use the memo'ed version when this condition is true."

src/js/components/LogRow.js
```diff
@@ -19,51 +20,69 @@ type Props = {
...
+        />
+      )
+    }
+    if (dimens.rowWidth !== "auto") {
+      return (
+        <div
+          className="log-cell"
+          key={key}
+          style={{width}}
+          onClick={handleClick}
```
jameskerr, line 62: I don't see why the log cell needs to have the click handler? Is that intentional?

src/js/components/LogRow.js
```diff
@@ -19,51 +20,69 @@ type Props = {
...
+      className={classNames("log-row", {highlight, even: index % 2 === 0})}
+      style={Styler.row(dimens)}
+      onClick={handleClick}
+    >
+      {columns.getVisible().map(renderCell)}
+    </div>
+  )
 }
+
+export default memo<Props>(LogRow, (prevProps: Props, nextProps: Props) => {
```
jameskerr, line 79: Thanks for doing this. It looks simpler that I'd thought it would be.

src/js/components/LogRow.js
mason-fish: Ah shoot, no this is a mistake. I put it here at first and realized it belonged on the whole row but forgot to remove this. Nice catch

src/js/components/LogRow.js
mason-fish: I had the same thought actually. Glad you agree, will change

jameskerr:
👍 I see what you mean. We needed the click event now. Thanks for fixing.

PR #978, opened 2020-08-10
Customize Brim Zeek artifact

### LINKED ISSUE #958: mention our zeek as starting point for zeek customization
The zeek customization wiki should document that users can start with our pre-built zeek releases & try to add their zeek scripts onto a copy of it. See:

https://brimsec.slack.com/archives/C010DR0HHMF/p1596037523492400?thread_ts=1596031961.491000&cid=C010DR0HHMF

philrz:
Users may want to avoid having to install/compile Zeek from scratch in order to make small customizations. This is particularly true on Windows where the lack of official Zeek support presents an extra challenge.

In this article I provide steps for starting from one of our Zeek artifacts and making minor customizations. I use Windows as the example platform since I expect this to be the most challenging for users, so hopefully readers that are on macOS/Linux will be able to easily intuit the equivalent commands.

In the prior version of this doc, the "Contact Us" section was in the middle, which I was in favor of at the time because it was a short doc and we really wanted to get the attention of users that were attempting to use this feature. However, now that the doc is longer it was feeling a bit awkward to still have it in the middle, so here I've moved it to the end to make it similar to the other docs in the wiki.

Closes #958.


docs/Zeek-Customization.md
```diff
@@ -69,3 +66,115 @@ runner without altering its location, any changes will take effect on the next p
...
+If you're experienced at [installing Zeek](https://docs.zeek.org/en/current/install/install.html), you may have the best luck
+by building and testing your Zeek installation via those standard instructions. This will give you the most flexibility in
+terms of being able to to compile plugins, use the [Zeek Package Manager](https://docs.zeek.org/projects/package-manager/en/stable/)
+to add extensions, or even make changes to the core Zeek C++ code if you choose.
+
+However, situations may arise when this is difficult. Some examples:
+
+1. **Microsoft Windows**. Zeek is not officially supported on Windows, and Brim's
+[fork of Zeek](https://github.com/brimsec/zeek) that includes limited Windows support is unfortunately not easy to compile
+outside of a customized build environment.
```
alfred-landrum, line 81:
```suggestion
[fork of Zeek](https://github.com/brimsec/zeek) that includes limited Windows support requires a complex build setup.
```

docs/Zeek-Customization.md
```diff
@@ -69,3 +66,115 @@ runner without altering its location, any changes will take effect on the next p
...
+2. **Download and unpack a current Brim Zeek artifact** for the platform from the
+[releases](https://github.com/brimsec/zeek/releases) page.
+
+```
+PS C:\temp\zeek> Invoke-WebRequest -Uri "https://github.com/brimsec/zeek/releases/download/v3.2.0-dev-brim6/zeek-v3.2.0-dev-brim6.windows-amd64.zip" -OutFile zeek.zip
+PS C:\temp\zeek> Expand-Archive zeek.zip
+```
+
+3. **Add/modify the Zeek scripts as necessary.** In this case, we're going to be enabling the Zeek scripts from the
+[zeek-log-all-http-headers](https://github.com/sethhall/zeek-log-all-http-headers) package. Since the scripts are in a GitHub
```
alfred-landrum, line 117:
```suggestion
3. **Add/modify the Zeek scripts as necessary.** In this case, we're going to add the Zeek scripts from the
[zeek-log-all-http-headers](https://github.com/sethhall/zeek-log-all-http-headers) package. Since the scripts are in a GitHub
```

PR #984, opened 2020-08-12
Uniqly Generate Space Names

jameskerr:
It was a poor user experience to import a pcap with the same name and get a "there is already a space with this name" error. Now we add a numeric suffix to the name if it already exists. I've also removed the `.brim` suffix since "brim folders" are no longer a thing. This uniq function will also help when we are create subspaces.

Screen Shot 2020-08-11 at 4 59 39 PM

It also works if you provide a custom data dir option to Brim.

Screen Shot 2020-08-11 at 4 59 48 PM

### DIFF
--- a/src/js/brim/ingest/getParams.js
+++ b/src/js/brim/ingest/getParams.js
@@ -1,9 +1,11 @@
 /* @flow */
-import type {IngestFileType} from "./detectFileType"
-import fileList, {type FileListData} from "./fileList"
-import time from "../time"
-import lib from "../../lib"
 import path from "path"
+
+import type {IngestFileType} from "./detectFileType"
+import {getUniqName} from "../../lib/uniqName"
+import fileList, {type FileListData} from "./fileList"
+import lib from "../../lib"
+import time from "../time"
 
 export type IngestParams = {
   dataDir: string,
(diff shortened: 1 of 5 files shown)


src/js/brim/ingest/getParams.js
```diff
@@ -18,6 +20,7 @@ export type IngestParamsError = {
 export default function getParams(
   data: FileListData,
   dataDir?: string,
+  existingNames?: string[] = [],
```
mason-fish, line 23: Not sure if you need the `?` optional specifier when you already use an initializer

Also, we might want to convert these parameters to be keys in a single object instead now that we've hit 4 params

src/js/flows/ingestFiles.js
```diff
@@ -63,7 +64,7 @@ const createDir = () => ({
   }
 })
 
-const createSpace = (client, dispatch, clusterId) => ({
+const createSpace = (client, gDispatch, clusterId) => ({
```
mason-fish, line 67: Good call to clarify this 👍

src/js/brim/ingest/getParams.js
jameskerr: Yeah, good instincts. I think 4 is close to max for this. I think you do need the optional parameter for flow to let you not provide a value. I could be wrong though.

PR #986, opened 2020-08-12
fix link in contributing

alfred-landrum:


alfred-landrum:
Yes, the link failed in a test yesterday, and then when it came up, the URL
to get to the content we intended is the one in the PR. Since that seems to
have reverted, I'll close the PR.

PR #989, opened 2020-08-12
Create a new "Current" Reducer

jameskerr:
This PR refactors the getters and setters of the current space and the current connection.

The `Current` namespace will be used for housing the tab's current space and current connection ids.  We'll now be able to get the current space like `Current.getSpace` and `Current.getConnection`. They will throw errors if they do not exist. There are also selectors like `Current.getSpaceId` and `Current.getConnectionId` which will return a string if present, or null if not set.

To Dos:

- [x] Write a migration to move the current ids into this reducer. Remove them from the `Search` reducer.


src/js/components/RightPane.js
```diff
@@ -94,14 +94,22 @@ export default class RightPane extends React.Component<Props, S> {
...
-const stateToProps = (state) => ({
-  isOpen: Layout.getRightSidebarIsOpen(state),
-  width: Layout.getRightSidebarWidth(state),
-  prevExists: LogDetails.getHistory(state).prevExists(),
-  nextExists: LogDetails.getHistory(state).nextExists(),
-  currentLog: LogDetails.build(state),
-  space: Tab.space(state)
-})
+const stateToProps = (state) => {
+  let space
```
mason-fish, line 98: `getSpaceSafe()` instead here?

src/js/components/LogDetailsWindow/index.js
```diff
@@ -2,24 +2,24 @@
 import {useDispatch, useSelector} from "react-redux"
 import React from "react"
 
+import {Center, Left, PaneHeader, PaneTitle, Right} from "../Pane"
```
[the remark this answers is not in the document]
jameskerr: I'm using "import-js" https://github.com/Galooshi/import-js for now. I've got an atom plugin that calls out to it. By default it groups and sorts the imports automatically. It's a little annoying when the diff is bigger just because of the sorting. We have a `.importjs.js` config file too at the root of the project.

src/js/components/RightPane.js
jameskerr: Definitely, thanks.

PR #995, opened 2020-08-14
841 b fix table flash

mason-fish:
Second stab at fixing #841


src/js/flows/executeTableSearch.js
```diff
@@ -37,6 +47,11 @@ export default function executeTableSearch(
         })
         .abort(resolve)
         .end((_id, count) => {
+          if (isBlocking) {
+            dispatch(Viewer.setRecords(tabId, collectedRecords))
+            dispatch(Viewer.updateColumns(tabId, collectedColumns))
+            dispatch(Columns.touch(collectedColumns))
+          }
```
jameskerr, line 54: As soon as I woke up today I thought of doing it this way. Yeah that works too!

src/js/state/Viewer/reducer.js
```diff
@@ -36,6 +36,8 @@ export default function(
       return {...state, status: action.status}
     case "VIEWER_COLUMNS":
       return {...state, columns: {...state.columns, ...action.columns}}
+    case "VIEWER_SET_COLUMNS":
+      return {...state, columns: {...action.columns}}
```
jameskerr, line 40: Nice, that makes sense you'd need that.

src/js/flows/executeTableSearch.js
```diff
@@ -14,16 +14,26 @@ import executeSearch from "./executeSearch"
...
         .id("Table")
         .status((status) => dispatch(Viewer.setStatus(tabId, status)))
         .chan(0, (records, types) => {
           const columns = hashDescriptorKeys(types)
+
+          if (isBlocking) {
+            Object.assign(collectedColumns, columns)
+            collectedRecords = collectedRecords.concat(records)
+            return
+          }
```
jameskerr, line 35: Great. I think this is the quickest way to get the feature in without changing too much code. Perfect for getting this release out. Next time we visit this, I have some ideas on ways to re-arrange it.

philrz:
I've tested this one out on the branch and it looks good to me as well. FWIW, I did notice that the blank white flash still happens if one has been doing queries and hence is presented with the "Refresh" button and clicks it. However, that kind of flash seems to be in-line with what a user might expect to see when hitting a Refresh button, so this does not bother me. I checked with @alfred-landrum and he's in agreement. So, it's great that we've addressed the main problem with the flashes during "lean-back" auto-refreshes. Looking forward to seeing this merged! 👍

PR #997, opened 2020-08-14
Tabbable Links on Import Page

jameskerr:
A quick clean up of some code for the import view. Also make the link's "tabbable".

image

### DIFF
--- a/src/js/components/LoadFilesInput.js
+++ b/src/js/components/LoadFilesInput.js
@@ -37,6 +37,7 @@ export default function LoadFilesInput({onChange}: Props) {
       {...bindDropzone()}
     >
       <input
+        tabIndex="-1"
         ref={setInput}
         type="file"
         multiple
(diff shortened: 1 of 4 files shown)


src/js/lib/cssVar.js
```diff
@@ -8,7 +8,7 @@ So we provide a fallback for the unit tests
 
 export function cssVar(name: string, fallback: string | number) {
   try {
-    return polished.cssVar(name)
+    return polishedCssVar(name)
```
jameskerr, line 11: This was not loading in the right order. It now works when I import the named function instead of importing the default.

src/js/components/LoadFilesInput.js
```diff
@@ -37,6 +37,7 @@ export default function LoadFilesInput({onChange}: Props) {
       {...bindDropzone()}
     >
       <input
+        tabIndex="-1"
```
jameskerr, line 40: Makes this not tabbable since it's hidden.

mason-fish: neat trick

PR #998, opened 2020-08-14
Persist event in useDoubleClick

jameskerr:
Two small changes here:

1. Because of React's [Event Pooling](https://reactjs.org/docs/events.html#event-pooling) which actually will be removed in [Version 17](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#no-event-pooling), anytime we save an event in an asycronous way, we need to call `event.persist()`.

2. The mousetrap bindings to switch tabs were not working consistently with the new constructor pattern, so I added them to the document Element. All seems to work well now.

### DIFF
--- a/src/js/components/hooks/useDoubleClick.js
+++ b/src/js/components/hooks/useDoubleClick.js
@@ -26,7 +26,8 @@ export default function useDoubleClick(
     return () => clearTimeout(singleClickTimer)
   }, [clicks])
 
-  return (e: MouseEvent) => {
+  return (e: SyntheticEvent<HTMLElement>) => {
+    e.persist()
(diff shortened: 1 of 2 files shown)


src/js/components/hooks/useDoubleClick.js
```diff
@@ -26,7 +26,8 @@ export default function useDoubleClick(
     return () => clearTimeout(singleClickTimer)
   }, [clicks])
 
-  return (e: MouseEvent) => {
+  return (e: SyntheticEvent<HTMLElement>) => {
+    e.persist()
```
mason-fish, line 30: Good call!

PR #1001, opened 2020-08-15
Create Subspace Flow

jameskerr:
This PR adds a new thunk called `createSubspace()`. It grabs the selected records from the viewer and issues the api call to zqd with the expected parameters. It doesn't add any UI, just preps the app with the new function and corresponding tests.

This PR also puts the `globalDispatch` method in the third parameter of the Thunk function signature. This makes it much easier to stub out in tests. Thunks now look like:

```js
type Thunk = (dispatch, getState, {zealot, globalDispatch}) => any
```

In the tests, we mock `globalDispatch` with the normal `dispatch` function so that it does not make any ipc calls to electron. The testing framework doesn't know about the electron APIs.

### DIFF
--- a/src/js/brim/record.js
+++ b/src/js/brim/record.js
@@ -7,8 +7,8 @@ export type $Record = {|
   columns: () => Column[],
   values: () => FieldValue[],
   data: () => RecordData,
-  find: (string) => ?$Field,
-  get: (string) => $Field
+  get: (string) => ?$Field,
+  mustGet: (string) => $Field
 |}
 
 export default function record(data: RecordData): $Record {
@@ -22,12 +22,12 @@ export default function record(data: RecordData): $Record {
     data() {
       return data
     },
-    find(name: string) {
+    get(name: string) {
       let fieldData = data.find((field) => field.name === name)
       return fieldData ? brim.field(fieldData) : null
     },
-    get(name: string) {
-      const f = this.find(name)
+    mustGet(name: string) {
+      const f = this.get(name)
       if (f) return f
       else throw new Error(`Missing field: ${name}`)
     }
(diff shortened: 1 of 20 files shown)


src/js/brim/record.js
```diff
@@ -14,9 +22,14 @@ export default function record(data: RecordData): $Record {
     data() {
       return data
     },
-    field(name: string) {
+    get(name: string) {
```
mason-fish, line 25: maybe `getField` ?

src/js/initializers/initStore.js
```diff
@@ -27,7 +28,10 @@ export default async () => {
     initialState,
     composeWithDevTools(
       applyMiddleware(
-        reduxThunk.withExtraArgument({zealot: createZealot("localhost:9867")})
+        reduxThunk.withExtraArgument({
+          zealot: createZealot("localhost:9867"),
+          globalDispatch
```
mason-fish, line 33: Are you thinking we should be limiting the scope of `globalDispatch` to be less global with this addition?

src/js/state/Spaces/types.js
```diff
@@ -27,7 +27,8 @@ export type Space = {
   max_time: Ts,
   pcap_support: boolean,
   storage_kind: "filestore" | "archivestore",
-  ingest: SpaceIngest
+  ingest: SpaceIngest,
+  parent_id?: string
```
mason-fish, line 31: 👍

src/js/brim/record.js
jameskerr: I had that at first but later felt that it is clear enough when I see it in context: `record.get("_path")`.

PR #1006, opened 2020-08-17
Double click bug

### LINKED ISSUE #1005: Opening Log Detail as first action throws an error
Repro is on Brim commit `1eeeb44`.

Starting from clean app state, I import a small pcap. Before doing anything else (i.e. I've not yet single-clicked any event row) if I double-click an event to bring up a Log Detail window, this error is thrown:

image

jameskerr:
Fixes #1005 

We were opening the log detail window before we were saving the log to the state. This now makes sure we push that log detail into the state before opening the detail window.

It also prevents the log detail state from pushing a duplicate log entry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "release": "node scripts/release --darwin",
     "prerelease": "npm run build",
     "prebuild": "npm run clean",
-    "postinstall": "npm-run-all -p deps:*",
+    "postinstall": "npm-run-all -s deps:*",
     "deps:zdeps": "node scripts/download-zdeps",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "release": "node scripts/release --darwin",
     "prerelease": "npm run build",
     "prebuild": "npm run clean",
-    "postinstall": "npm-run-all -p deps:*",
+    "postinstall": "npm-run-all -s deps:*",
```
jameskerr, line 36: Running these parallel makes it hard to know why an npm install is taking so long.

PR #1007, opened 2020-08-18
New Connection page

### LINKED ISSUE #849: Create the New Connection Page

mason-fish:
#closes #849 

The new connection modal! There were a lot of little corner cases to work out here (switching tabs/windows mid ingest across multiple connections, etc.) but I think this is in a good spot now. Some of the changes include changing our cluster/connection id to always reflect the `host:port`. For that reason I also included a migration to swap out all of the existing "zqd" ids we defaulted to previously to use this new format. We can add a `name` field later on to support more flexible titles.

image

Not much to fill in here yet...
image

image


src/js/test/initTestStore.js
```diff
@@ -23,7 +23,9 @@ export default (zealot: * = createZealotMock()): TestStore => {
     undefined,
     compose(
       applyDispatchAll(),
-      applyMiddleware(reduxThunk.withExtraArgument({zealot})),
+      applyMiddleware(
+        reduxThunk.withExtraArgument({zealot, createZealot: createZealotMock})
```
jameskerr, line 27: For this I was thinking....

```js
function initTestStore(zealot = createZealotMock()) {
  //...
  const createZealot = () => zealot
  //...
  reduxThunk.withExtraArgument({createZealot})
  //...
}
```

This way we could keep the method signature of `iniTestStore` the same in all the tests, and continue to stub responses in the same way we have been doing.

src/js/flows/submitArchiveSearch.js
```diff
@@ -13,8 +13,9 @@ import whenIdle from "../lib/whenIdle"
 export default (patterns: string[]): Thunk => (
   dispatch,
   getState,
-  {zealot}
+  {createZealot}
 ) => {
+  const zealot = createZealot(Current.getConnectionId(getState()))
```
jameskerr, line 18: I wonder if it would be worth dispatching a selector to make this more conscise, since we're doing it so many places.

```js
const getZealot = () => (dispatch, getState, {createZealot}) => {
  return createZealot(Current.getConnectionId(getState()))
}
```

```js
// then in all the places we could just do
const zealot = dispatch(getZealot())
```

src/js/flows/submitAutoRefreshSearch.js
```diff
@@ -10,10 +10,12 @@ export default function submitAutoRefreshSearch(): Thunk {
   return function(dispatch, getState) {
     const state = getState()
     const prevArgs = Search.getArgs(state)
+    console.log("prevArgs: ", prevArgs)
```
jameskerr, line 13: A few stragglers...

src/js/test/initTestStore.js
```diff
@@ -23,12 +23,20 @@ export default (zealot: * = createZealotMock()): TestStore => {
...
+  // $FlowFixMe
   store = createStore(
     rootReducer,
     undefined,
     compose(
       applyDispatchAll(),
-      applyMiddleware(reduxThunk.withExtraArgument({zealot, globalDispatch})),
+      applyMiddleware(
+        reduxThunk.withExtraArgument({
+          zealot,
```
jameskerr, line 35: I don't think we need the old zealot in here now.

src/js/components/ClusterPicker.js
```diff
@@ -25,13 +29,37 @@ const ClusterPickerWrapper = (styled.div`
...
+      checked: isCurrent,
+      click: () => !isCurrent && dispatch(setConnection(c))
+    }
+  })
+
+  template.push(
+    {type: "separator"},
+    {
+      label: "+ New Connection",
+      click: () => dispatch(Current.setConnectionId(""))
```
jameskerr, line 50: setConnectionId now accepts `null` for when there is no connection. I'd prefer that over the empty string we used to have.

src/js/components/ClusterPicker.js
```diff
@@ -25,13 +29,37 @@ const ClusterPickerWrapper = (styled.div`
...
+    electronIsDev && openMenu(e.currentTarget)
+  }
 
   return (
-    <ClusterPickerWrapper>
-      <label>
-        {current.host}:{current.port}
-      </label>
+    <ClusterPickerWrapper onContextMenu={onContextMenu}>
+      <label>{current.id}</label>
```
jameskerr, line 62: Right now the connection id happens to the be the host:port combo, but it seems unreliable to assume that will be the case in the long-term. I'd suggest that when we actually need the host, we call a function with a name like getHost. 

For example, when we need to construct the zealot client, I think it's more clear to have a function called `Current.getConnectionHost` instead of `Current.getConnectionId`, even thought at the moment the result is the same.

The same goes for the line above. I'd suggest keeping the {host}:{port} join or making a little function somewhere that does it.

src/js/state/Modal/reducer.js
```diff
@@ -1,26 +1,20 @@
...
+      draft.name = action.name
+      draft.args = action.args
+      return
     case "MODAL_HIDE":
-      return {
-        ...init
-      }
-    default:
-      return state
+      Object.assign(draft, init())
```
jameskerr, line 17: You can just return a new object from the produce method. So this could be:
```js
return init()
```

src/js/flows/openLogDetailsWindow.js
```diff
@@ -3,7 +3,15 @@
...
 import type {Thunk} from "redux-thunk"
+import {mustGetConnection} from "../state/Current/selectors"
 
 export const openLogDetailsWindow = (): Thunk => (dispatch, getState) => {
-  invoke(ipc.windows.open("detail", {size: [700, 600]}, getState()))
+  const {host, port} = mustGetConnection(getState())
+  invoke(
+    ipc.windows.open(
+      "detail",
+      {size: [700, 600], query: {host, port}},
```
jameskerr, line 13: Good catch!

src/js/components/NewConnectionModal.js
```diff
@@ -89,7 +100,16 @@ export default function NewConnectionModal() {
       let form = brim.form(f, config)
 
       if (await form.isValid()) {
-        form.submit()
+        const {host} = form.getFields().reduce((obj, field) => {
```
jameskerr, line 103: Maybe this reduce call could be in `form.getFields`? It mike make it a little easier to work with.

Issue #982, opened 2020-08-11
"Syntax docs" window is unusable


philrz:
Repro is in Brim `v0.14.0`.

There's a link to the ZQL docs from the pull-down menu via **Help > Query Syntax Docs** that works as I'd expect, in that it opens up the docs in my browser. However, I recently noticed (per the attached video) that there's also a "Syntax docs" option available in the vertical "..." menu to the right of the search bar. When clicked, I find I can't move the window, nor can I click on any of the links in it that go to the other docs pages. As a user, I find it unusable in its current form.

[Repro.zip](https://github.com/brimsec/brim/files/5059762/Repro.zip)

I'd suggest that we either make it behave more like the one from the **Help** pull-down or remove it entirely.

philrz:
Current thinking is we should just remove it for now.

r-franke:
Working on it.

philrz:
Verified in Brim commit `7cf09ff`. Per the attached video, the option remains in the menu, but now when clicked it behaves like the one in the **Help** pull-down such that it opens in my regular browser window that I can move around, click links in it, etc.

[Verify.zip](https://github.com/brimsec/brim/files/5133366/Verify.zip)

Thanks @r-franke!

Issue #975, opened 2020-08-06
Search tool in column chooser


philrz:
For addressing the problem described in #949, another approach proposed by the Dev team is to offer some kind of search tool. This could prove handy in situations where the number of fields is long enough that even searching through an alphabetized list might be tedious. This would also help with finding common substrings in the middle of field names. To offer an example with Zeek data, search for "tcp" would lead to field names like `active_tcp_conns`, `reassem_tcp_size`, and `tcp_conns`.

r-franke:
Working on it.

philrz:
Verified in Brim commit `7cf09ff`.

As shown in the attached video, this enables a use case where I've isolated just the Zeek `files` logs, and I want to pick a couple favorite fields, then easily find & include all the ones that have `uid` anywhere in the field names. Very handy!

[Verify.zip](https://github.com/brimsec/brim/files/5133339/Verify.zip)

That said, the verification made me think of a follow-on enhancement #1029 we may want to add at some point. But what we have here is already a huge improvement.

Thanks @r-franke!

Issue #949, opened 2020-07-27
Sort field names in column chooser


philrz:
As a user, I was recently working with the column picker. I was starting with the full diverse set of Zeek logs with intent to click "Deselect All" in the picker and then select just a handful. Having imported test data [wrccdc.2018-03-23.010014000000000.pcap.gz](https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz) (after `gunzip`'ing it), the picker in that case starts out with 120 field names to choose from. Knowing the names of the fields I wanted to select, I found it a little difficult to pick them out of the list.

The Dev team has explained there's deterministic logic to how the list is presented currently. Specifically, we match the order of the columns in the column chooser with the viewer results column order. While that's certainly better than being random, I suspect that may not come through in an obvious way to the user.

In a conversation with @alfred-landrum, we agreed that a likely place to start with improving this would be to sort the field names alphabetically in the column picker.

philrz:
Verified in Brim commit `7cf09ff`.

For comparison, using the pcap described above, here's how the unsorted column chooser contents were looking in Brim GA tagged `v0.15.1` before the enhancement:

![image](https://user-images.githubusercontent.com/5934157/91369906-cd762d80-e7c1-11ea-89f1-5d0c8cf56799.png)

Now we can see it sorted in Brim commit `7cf09ff`:

![image](https://user-images.githubusercontent.com/5934157/91369842-a4ee3380-e7c1-11ea-9ae2-806cdfb88e52.png)

Thanks @r-franke!

PR #1012, opened 2020-08-20
Column chooser sorting and searching

r-franke:
This PR implements #949 and #975.

I think it still needs a few last touches and discussion. 

I wasn't sure what to do with the styling of the input field. I considered styling it like the main search field, but that takes up a lot of space. So input is appreciated on this.

Second part is the sorting of the column names. From Phil's issues I know there is some discussion on this. So maybe we can continue that discussion here and choose an approach. 

I also think that the top of the column chooser is taking up quite a lot of real estate due to white space as it is right now. 

Maybe we could replace the select all and deselect row with a checkbox above the other check-boxes and before the new search input field. I believe it's done more often that way and compresses the header a bit.

**Example**

```
[ x ]  [ search input field                                ]
[   ]  AA
[ x ]  RA
..... and so on
```

**Demo**

filter_demo

**Disclaimer**
Still need to fix some flow errors, but wanted to get the discussion done first.

**Out-of-scope-ninja-fixes** 
#982 is fixed.

### DIFF
--- a/src/css/_column-chooser.scss
+++ b/src/css/_column-chooser.scss
@@ -63,6 +63,15 @@
     padding-bottom: 12px;
     height: 100%;
   }
+
+  .search-input {
+    width: 100%;
+  }
+
+  #column-search {
+    width: 100%;
+  }
+
 }
 
 .slide-in-right-appear {
@@ -73,3 +82,4 @@
   transition: transform 150ms;
   transform: translate(0, 0);
 }
+
--- a/src/css/_column-chooser.scss
+++ b/src/css/_column-chooser.scss
@@ -77,7 +77,6 @@
     right: -2px;
     top: -2px;
   }
-
 }
 
 .slide-in-right-appear {
@@ -88,4 +87,3 @@
   transition: transform 150ms;
   transform: translate(0, 0);
 }
-
--- a/src/css/_column-chooser.scss
+++ b/src/css/_column-chooser.scss
@@ -67,6 +67,7 @@
   .search-input {
     width: 100%;
     align-content: center;
+    margin-bottom: 12px;
   }
 
   #column-search {
--- a/src/css/_column-chooser.scss
+++ b/src/css/_column-chooser.scss
@@ -66,10 +66,16 @@
 
   .search-input {
     width: 100%;
+    align-content: center;
   }
 
   #column-search {
     width: 100%;
+  }
+  #column-search-clear {
+    position: absolute;
+    right: -2px;
+    top: -2px;
   }
 
 }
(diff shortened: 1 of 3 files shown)


jameskerr:
Hey Remy, this is looking great. Here is how you'll fix those flow errors you're encountering:

```js
type LocalState = {
  searchValue: string
}

export default class ColumnChooserMenu extends React.Component<Props, LocalState> {
  state: LocalState = {
    searchValue: ""
  }
  // ...
}
```

As for styling, there is a nice search input component that can be swapped out with the input element very easily. I also don't think the close button will be necessary for this feature.

```diff
               <div className="search-input">
-                <input
+                <SearchInput
                   id="column-search"
                   type="text"
                   value={this.state.searchValue}
                   onChange={this.handleChange}
                 />
-                <CloseButton
-                  id="column-search-clear"
-                  onClick={this.clearSearch}
-                />
               </div>
```

src/css/_column-chooser.scss
```diff
@@ -63,6 +63,20 @@
     padding-bottom: 12px;
     height: 100%;
   }
+
+  .search-input {
+    width: 100%;
+    align-content: center;
```
jameskerr, line 69: I think a `margin-bottom: 12px` looks good here.

jameskerr:
I think it's fine to sort the list alphabetically for now.

src/css/_column-chooser.scss
r-franke: Done

r-franke:
New `SearchInput` in use:
image

src/js/components/ColumnChooserMenu.js
```diff
@@ -139,16 +159,39 @@ export default class ColumnChooserMenu extends React.Component<Props> {
...
-            {columns.map((c) => (
-              <ColumnListItem key={`${c.name}-${c.type}`}>
-                <Checkbox
-                  label={c.name}
-                  checked={c.isVisible}
-                  onChange={(e) => this.onColumnClick(e, c)}
+            <ControlListItem>
+              <div className="search-input">
+                <SearchInput
+                  id="column-search"
```
jameskerr, line 165:
```suggestion
                  autoFocus
```

Looks like the id isn't necessary.

And I think an autoFocus would be helpful here.

src/css/_column-chooser.scss
```diff
@@ -63,6 +63,21 @@
...
+  }
+
+  #column-search {
+    width: 100%;
+  }
+  #column-search-clear {
+    position: absolute;
+    right: -2px;
+    top: -2px;
+  }
```
jameskerr, line 80:
```suggestion
```

It doesn't look like these will be necessary anymore.

r-franke:
I think all changes are done, dunno why it still shows the requested changes?

PR #1016, opened 2020-08-20
Extract contextmenu builder

jameskerr:
This PR just hoists some of the data needed to build the right click menu on the cells of the table into a component higher in the tree. This way, depending on what type of results are rendered, we can use different kinds of right click menus.


mason-fish:
Lgtm. The results table component now accepts a context menu via props instead of building it internally. I think you'll also want to adjust the stateProps type here (https://github.com/brimsec/brim/pull/1016/files#diff-d54648f0d6e8db1fe2e2b2e0e038fe2eR41-R42) since it no longer uses the `program` and `space`

jameskerr:
Closing this for now. I'll revisit this PR when we actually need two different right click menus.

PR #1020, opened 2020-08-25
Additional debug steps for pcap extraction problems

philrz:
I recently helped a community user debug a problem where Brim was not extracting flows from their pcap ([Slack thread](https://brimsec.slack.com/archives/C010DR0HHMF/p1597642269030000)). Almost immediately after getting the debug process started for that user, yet another user reported a similar type of symptom. That made me realize that we could probably benefit from a wiki article I can reference next time this comes up.


docs/Troubleshooting.md
```diff
@@ -25,13 +25,85 @@ Brim invokes to extract flows from your pcap handles the most common variations
...
-
-If you don't see an error message, or the error seems terse and unhelpful, it
-may help us get to the bottom of it faster if you include the debug info from
-[Developer Tools](#developer-tools).
+Such problems are often due to file format variations that are unique to the
+system that generated the capture. In such situations, the most expedient debug
+path would be to send us the capture file so our developers can use it to
+reproduce the problem, fix it, and create automated tests to ensure the problem
+doesn't arise again in the future. We understand that concerns regarding
+[sensitive information](http://0.0.0.0:6419/#sensitive-information-important)
```
alfred-landrum, line 33: What's the intended destination for this link?

philrz: @alfred-landrum: Gah! Thanks for catching that. It was supposed to be a link to an anchor within the same article. I just made a stupid mistake and pasted an absolute URL from when I was rendering the markdown locally with `grip`. I just pushed a commit that should fix the link.

docs/Troubleshooting.md
```diff
@@ -25,13 +25,85 @@ Brim invokes to extract flows from your pcap handles the most common variations
...
+However, if you cannot share your data, proceed with the following steps.
+
+First, if you don't see an error message in Brim, or the error seems terse and
+unhelpful, it may help us get to the bottom of it faster if you include the
+debug info from [Developer Tools](#developer-tools).
+
+Beyond that, the most comprehensive way to debug is to use the
+[zq](https://github.com/brimsec/zq) toolset at the command line to perform
+steps that are functionally equivalent to those initiated by the Brim
+application to extract flows from your pcap. This will allow you to better
```
alfred-landrum, line 52: Just a suggestion: say up front that if you go through this process, you expect it to fail. You mention this after all the steps, but I think its worth mentioning in some form beforehand, like:
```suggestion
application to extract flows from your pcap, and report at which point you see a failure. This will allow you to better
```

philrz: Thanks @alfred-landrum. I adjusted the wording using your suggestion as a starting point.

PR #1024, opened 2020-08-26
Index Search

### LINKED ISSUE #926: design subspace creation from search tabular results

jameskerr:
Fixes #926 

This PR allows searching an archive's index from within brim. After opening an archive, select "Index" from the search target selector on the left of the search bar. After issuing a valid search and getting some results, you'll be able to select one or more "chunks". Click the new "Subspace" button in the toolbar to create a new subspace based on that chunk.

Some notable changes in the code:

* Moved components related to the SearchBar and the Toolbar into their own folders with `src/js/components`
* Converted some stylesheet+className into styled-components
* Created a new reducer to store that "last" search record. We could not rely on the history because there are valid cases when we need to issue a search without saving it to the history. However, there are many places where we need to know what the actual last search was in order to render or not render certain components.
* Added three different icons for a space (archive, space, subspace)

image

image

image

Forgot that I need to write a migration to add the search target state.

- [x] Write the necessary migrations


src/js/components/FilterTree.js
```diff
@@ -15,113 +15,97 @@ import Current from "../state/Current"
...
+    {
+      label: "Delete All",
+      click: () => globalDispatch(Investigation.clearInvestigation())
+    }
+  ])
 
-  function renderNode(node: Node, i: number) {
-    const findingSpaceName = get(
+  function onNodeClick() {
+    dispatch(Search.restore(node.data.finding.search))
```
jameskerr, line 54: Removed some duplication here and used the shared Search.restore method.

src/js/components/FilterTree.js
```diff
@@ -15,113 +15,97 @@ import Current from "../state/Current"
...
+        onClick={onNodeClick}
+        onContextMenu={() => menu.open()}
+        title={findingSpaceName}
+      >
+        <FilterNode filter={node.data.filter} />
+        {renderWarning()}
+      </div>
+      <div className="filter-tree-children">
+        {node.children.map((node, i) => (
+          <NodeRow node={node} i={i} key={i} />
```
jameskerr, line 99: The NodeRow component used to be a plain function. I converted it to Components so that the hooks in them will be associated with each Row, rather than the parent component.

src/js/components/Investigation/FindingCard.js
```diff
@@ -66,7 +61,7 @@ export default React.memo<Props>(function FindingCard({finding}: Props) {
     )
   }
 
-  const template = [
+  const menu = usePopupMenu([
```
jameskerr, line 64: Changed the return value of usePopMenu to include an "onClick" handler along with some other nice things like "isOpen".

src/js/components/FilterTree/helpers.test.js
```diff
@@ -11,7 +11,8 @@ function search(pins, program) {
       program,
       spaceName: "default",
       spaceId: "defaultId",
-      spanArgs: brim.dateTuple([new Date(1), new Date(2)]).toSpan()
+      spanArgs: brim.dateTuple([new Date(1), new Date(2)]).toSpan(),
+      target: "events"
```
jameskerr, line 15: Added a `target` property to the SearchRecord type.

src/js/components/LogRow.js
```diff
@@ -62,9 +59,8 @@ const LogRow = (props: Props) => {
     <div
       className={classNames("log-row", {highlight, even: index % 2 === 0})}
       style={Styler.row(dimens)}
-      onClick={(e) => {
-        handleClick(e)
-      }}
+      onClick={onClick}
+      onDoubleClick={onDoubleClick}
```
jameskerr, line 63: Due to decoupling the row selection from the logDetail stack, we can fire the click handler twice on double click without issue. This makes selecting the rows feel much snappier now since we don't have to wait for the 250ms timeout.

src/js/components/SearchBar/Button.js
```diff
@@ -0,0 +1,45 @@
+/* @flow */
+import {darken} from "polished"
+import styled from "styled-components"
+
+import type {Styled} from "../../types/styled"
+import {cssVar} from "../../lib/cssVar"
+
+const Button: Styled<> = styled.button`
```
jameskerr, line 8: Used for the Left and Right history arrows in the search bar.

src/js/components/SearchBar/Error.js
```diff
@@ -0,0 +1,41 @@
...
+
+import {useSelector} from "react-redux"
+import React from "react"
+import styled from "styled-components"
+
+import type {Styled} from "../../types/styled"
+import SearchBar from "../../state/SearchBar"
+import Warning from "../icons/warning-sm.svg"
+
+const Wrap: Styled<> = styled.div`
```
jameskerr, line 11: The error message that appears if the syntax is invalid.

image

src/js/components/SearchBar/InputAction.js
```diff
@@ -0,0 +1,39 @@
+/* @flow */
+import {darken} from "polished"
+import styled from "styled-components"
+
+import type {Styled} from "../../types/styled"
+import {cssVar} from "../../lib/cssVar"
+
+const InputAction: Styled<> = styled.button`
```
jameskerr, line 8: Used for the buttons on the right side of the search input.

image

src/js/components/SearchBar/InputBackdrop.js
```diff
@@ -0,0 +1,24 @@
+/* @flow */
+import styled from "styled-components"
+
+import type {Styled} from "../../types/styled"
+
+const Input: Styled<> = styled.div`
```
jameskerr, line 6: This is fun. I realized I could have a much easier time laying out the search bar if I made a containing "div" look like the search input. Then the actual input where the user types is completely transparent. I could then use flexbox to layout the buttons instead of padding and absolute positions.

This component is the background.

image

src/js/components/SearchHeaderChart.js
```diff
@@ -2,17 +2,18 @@
...
 
+import Last from "../state/Last"
 import MainHistogramChart from "./charts/MainHistogram/Chart"
-import Tab from "../state/Tab"
 import brim from "../brim"
 
 export default function SearchHeaderChart() {
-  let searchRecord = useSelector(Tab.currentEntry)
-  let chartable = true
+  const search = useSelector(Last.getSearch)
```
jameskerr, line 10: Here is the first usage of the new "get actual last search record" feature. We check to see if we should show the histogram component.

src/js/components/charts/MainHistogram/useMainHistogram.js
```diff
@@ -48,12 +48,12 @@ export default function(width: number, height: number): HistogramChart {
 
     function onFocus(dates: DateTuple) {
       dispatch(search.setSpanFocus(time.convertToSpan(dates)))
-      dispatch(submitSearch({history: false, investigation: false}))
+      dispatch(submitSearch({history: true, investigation: false}))
```
jameskerr, line 51: We do want to save the search to the "Back and Forward" history when you zoom in or focus on the histogram..

src/js/components/hooks/useDelayedMount.js
```diff
@@ -1,17 +1,19 @@
...
     if (show) {
-      id.current = setTimeout(() => setReady(true), delay)
+      id = setTimeout(() => setReady(true), delay)
     } else {
-      clearTimeout(id.current)
       setReady(false)
     }
+    return () => {
+      clearTimeout(id)
+    }
```
jameskerr, line 16: This fixes that warning we'd see in the console at times from the SearchSpinner.

src/js/flows/submitSearch.js
```diff
@@ -1,54 +1,22 @@
...
-      case "zoom":
-        return dispatch(executeTableSearch(tabId, args))
-      default:
-        dispatch(executeHistogramSearch(tabId, args, prevArgs))
-        return dispatch(executeTableSearch(tabId, args))
-    }
+    return dispatch(
+      record.target === "index"
+        ? submitIndexSearch(save, ts)
+        : submitEventsSearch(save, ts)
```
jameskerr, line 19: There is a lot of duplication in both these methods. I plan to write some new and better tests for them each, then dry them out in a later PR.

src/js/lib/bytes.js
```diff
@@ -0,0 +1,12 @@
...
+export function formatBytes(bytes: number, decimals?: number = 2) {
+  if (bytes === 0) return "0 Bytes"
+
+  const k = 1024
+  const dm = decimals < 0 ? 0 : decimals
+  const sizes = ["Bytes", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"]
+
+  const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + " " + sizes[i]
```
jameskerr, line 11: Copied and pasted strait from stack overflow.

src/js/state/Current/selectors.js
```diff
@@ -43,7 +44,7 @@ export const mustGetSpace = createSelector<
   if (!spaces[connId]) throw new Error(`No spaces in connection id: ${connId}`)
   if (!spaces[connId][spaceId]) throw new Error(`Missing space id: ${spaceId}`)
 
-  return spaces[connId][spaceId]
+  return brim.space(spaces[connId][spaceId])
```
jameskerr, line 47: Instead of returning just the data, we now construct a brim.space object with some handy methods.

src/js/state/getPersistable.js
```diff
@@ -11,6 +11,7 @@ export default function getPersistable(state: *) {
     for (const tab of draft.tabs.data) {
       delete tab.viewer
       delete tab.chart
+      delete tab.last
```
jameskerr, line 14: Dont need to persist the last reducer.

philrz:
I was just testing the branch at Brim commit `46bed5f` which is talking to `zqd` commit `42103ef`. After following the steps in the `zar` README to create microindexes, I did an Index Search for `uri=/file` and tried to create a subspace out of the found chunks. This produced the error shown below, presumably due to the scheme we're currently using to auto-name subspaces.

image

philrz (approved):

PR #1031, opened 2020-08-27
zqd: Prevent orphaned zqd (unix)

### LINKED ISSUE #1018: Leaked zqd process on ungraceful brim shutdown
If brim is ungracefully shutdown the zqd process launched by brim continues on. There should be some functionality in zqd (or mode for running zqd) where zqd will terminate if the parent Brim process ceases to exist.

Steps to reproduce:
1. Start brim
2. Find brim pid
3. Run `kill -9 $brimpid`
4. Observe that the brim launched zqd process is still running `ps aux | grep zqd`

Note: This does not appear to be a problem on the Windows version of Brim.

mattnibs:
Prevent orphaned zqd when Brim exits with SIGKILL.

If platform is not windows, create posix pipe and pass read fd
to zqd via the -brimfd argument. Should brim be closed with SIGKILL
the fd will be closed and zqd, seeing the fd has closed will exit.

Closes #1018


mattnibs:
zqd changes here: brimsec/zq#1184

src/js/zqd/zqd.js
```diff
@@ -124,6 +124,12 @@ export class ZQD {
       "-zeekrunner",
       zeekRunnerCommand(this.zeekRunner)
     ]
+
+    if (process.platform !== "win32") {
+      const {readfd} = require("node-pipe").pipeSync()
+      opts.stdio.push(readfd)
+      args.push("-brimfd", opts.stdio.length - 1)
```
alfred-landrum, line 131:
```suggestion
      args.push("-brimfd", opts.stdio[opts.stdio.length - 1])
```

mattnibs: I originally thought this to be the case as well but because you pass the readfd as the fourth arg in `stdio`, the fd on the newly created process ends up being 4.

src/js/zqd/zqd.js
```diff
@@ -124,6 +124,12 @@ export class ZQD {
       "-zeekrunner",
       zeekRunnerCommand(this.zeekRunner)
     ]
+
+    if (process.platform !== "win32") {
```
mattnibs, line 128:
```suggestion
    // For unix systems, pass posix pipe read file descriptor into zqd process.
    // In the event of Brim getting shutdown via `SIGKILL`, this will let zqd
    // know that it has been orphaned and to shutdown.
    if (process.platform !== "win32") {
```

src/js/zqd/zqd.js
```diff
@@ -124,6 +124,15 @@ export class ZQD {
...
       zeekRunnerCommand(this.zeekRunner)
     ]
+
+    // For unix systems, pass posix pipe read file descriptor into zqd process.
+    // In the event of Brim getting shutdown via `SIGKILL`, this will let zqd
+    // know that it has been orphaned and to shutdown.
+    if (process.platform !== "win32") {
+      const {readfd} = require("node-pipe").pipeSync()
+      opts.stdio.push(readfd)
+      args.push(`-brimfd=${opts.stdio.length - 1}`)
```
jameskerr, line 134: I'm not sure how this works so I may be off, but it looks like you'll pass in the integer 3 here. Were you intending to access the array `opts.stdio[opts.stdio.length - 1]`?

mattnibs: This is intended. See https://github.com/brimsec/brim/pull/1031#discussion_r478654813

jameskerr (approved):
This works when I tested it out on my machine.

alfred-landrum:
Merging for @mattnibs who's out at the moment.

PR #1035, opened 2020-08-31
Render an error message if the initialize fails

jameskerr:
## UPDATE:

I changed the colors a bit. It now looks like this:

Screen Shot 2020-10-02 at 2 47 55 PM

If initialization fails, render an error message instead of nothing. The try again link will relaunch the app and the contact support link will email support@brimsecurity.com.

image

12Es8n47rt


jameskerr:
After talking about to everyone at the Desktop Jam, we decided to not make a call to zqd until we've rendered the app. This message will be reserved for completely unexpected failures during startup.

src/js/search.js
```diff
@@ -2,29 +2,34 @@
...
+        <Provider store={store}>
+          <ThemeProvider theme={theme}>
+            <App />
+          </ThemeProvider>
+        </Provider>
+      </AppErrorBoundary>,
+      lib.doc.id("app-root")
+    )
+  })
+  .catch((e) => {
```
mason-fish, line 33: 💯

mason-fish (approved):

src/js/components/StartupError.tsx
```diff
@@ -0,0 +1,97 @@
...
+import styled from "styled-components"
+import {remote} from "electron"
+import Link from "./common/Link"
+import ToolbarButton from "./Toolbar/Button"
+
+const Wrap = styled.div`
+  padding: 24px;
+  padding-top: 54px;
+  min-height: 100vh;
+  background: linear-gradient(to bottom right, #f8eee3, #fae1e0);
```
mason-fish, line 11: want to use a css var here? This gradient will also be used again in the connection detail designs iirc

mason-fish (approved):
Nice! Just left one comment

src/js/components/StartupError.tsx
jameskerr: Yeah good point.

philrz:
Verified in Brim commit `6f7c3be`.

For a real world example, I did a repro of #1099. A better fix for that is in the works (indeed, it's our goal to never have bugs that land us in this state!) but here we can see the incremental improvement. With config left behind for an inaccessible remote `zqd`, now when I start up, instead of the blank white screen I get:

image

The "Relaunch" unfortunately doesn't help much on its own, since the problem is persistent. But the "Contact Support" button does initiate an email.

image

This is all better than a blank white screen and expecting users to find their way into DevTools to deliver us the error messaage, so, a good improvement.

Thanks @jameskerr!

PR #1069, opened 2020-09-11
Add file signature for nanosecond pcap to accepted list

### LINKED ISSUE #1068: Support import of nanosecond pcap files
The capture I used in https://github.com/brimsec/zq/issues/1240 was initially generated via:

```
$ tcpdump -i ens4 -w pings.pcapnano --time-stamp-precision=nano net 192.168.1.0/24
```

[pings.pcapnano.zip](https://github.com/brimsec/brim/files/5209849/pings.pcapnano.zip)

That format is actually a variant that Brim is not yet trained to recognize:

```
$ capinfos pings.pcapnano
File name:           pings.pcapnano
File type:           Wireshark/tcpdump/... - nanosecond pcap
```

Therefore when dragged it into Brim commit `73d51c1`, the user experience:

image

Wireshark and Zeek both happily read it, though. I was able to turn it into usable test data for https://github.com/brimsec/zq/issues/1240 by opening it in Wireshark and saving it as a pcapng.

I can see by:

https://github.com/brimsec/brim/blob/63233ef59e46ddf1f4630fe205f3474fe7fd33cb/src/js/brim/ingest/detectFileType.js#L4-L7

...where we recognize the file signatures of capture formats, and the signature `4d 3c b2 a1 ` I observe for this capture is not in that list.

```
hexdump pings.pcapnano | head -1
0000000 4d 3c b2 a1 02 00 04 00 00 00 00 00 00 00 00 00
```

The format does not seem to be widely discussed online, but I do see one of the well-known Wireshark developers speaking of this signature at https://osqa-ask.wireshark.org/questions/51702/magic-numbers-for-supported-capture-files-for-wireshark, so it seems appropriate for us to add it to our list.

philrz:
Fixes #1068.

Now the (unzipped) [pings.pcapnano.zip](https://github.com/brimsec/brim/files/5209860/pings.pcapnano.zip) can be successfully imported.

[Success.zip](https://github.com/brimsec/brim/files/5209873/Success.zip)

I'll also look at adding an automated test in the zq repo that works with this file format.

### DIFF
--- a/src/js/brim/ingest/detectFileType.js
+++ b/src/js/brim/ingest/detectFileType.js
@@ -4,7 +4,8 @@ import fs from "fs"
 const PCAP_1_HEX = "d4c3b2a1"
 const PCAP_2_HEX = "a1b2c3d4"
 const PCAPNG_HEX = "0a0d0d0a"
-const PCAP_HEXES = [PCAP_1_HEX, PCAP_2_HEX, PCAPNG_HEX]
+const PCAPNANO_HEX = "4d3cb2a1"
+const PCAP_HEXES = [PCAP_1_HEX, PCAP_2_HEX, PCAPNG_HEX, PCAPNANO_HEX]
 
 export type IngestFileType = "pcap" | "log"
 


jameskerr (approved):
Cool, makes sense. If you give me a small pcap sample with this format, we can add an integration test as well. I think we have integration tests for all these other ones.

philrz:
Thanks @jameskerr. There's a link to a sample pcap in my opening text of the issue. However you'd not have luck getting an integration test working yet because there's a backend issue https://github.com/brimsec/zq/issues/1240 such that even when we let these nano pcaps be imported the user won't be able to yet extract flows from them. I've made a note-to-self to circle back with you once that issue is fixed.

Issue #1036, opened 2020-08-31
Migrate to Typescript


jameskerr:
I recommend migrating the whole codebase in one PR. Trying to make [flow and typescript work together](https://medium.com/@michaelsholty/migrating-500k-lines-of-flow-code-to-typescript-15a8cad43fec) in the same project seemed like the same amount of work as doing it all at once. It might make sense for bigger projects, but we have about 30K lines of javascript that need to be ported over.

**Strategy**
I think we should follow the instructions given in this [recent blog post.](https://skovhus.github.io/blog/flow-to-typescript-migration/) The author has ported several flow projects to typescript. He says a 30K line project he recently did took him a few days.

**Babel Typescript**
Since we already use babel to transform files in the src to the dist directory, let use the babel-typescript plugin instead of ditching babel completely for tsc. This means that we will still use tsc to type check, but the code will be built using babel. The workflow will be very similar to what we have now.

PR #1075, opened 2020-09-11
Move from Flow to Typescript

jameskerr:
Flow is gone and TypeScript is here!

Babel still transpiles the code, but instead of typechecking with `npm run flow`, we'll now use `npm run tsc`.

TypeScript caught more errors that Flow did, so I've fixed those it found.

All files must now end in `.ts` and all JSX files must end in `.tsx`


alfred-landrum:
@jameskerr : fyi, I pulled this onto my Windows dev machine, and hit the same issue as the CI, notice this line:
`Successfully compiled 0 files with Babel (222ms).`. For some reason, when I run `npm run build`, I don't see any files created under `dist`.

I also hit https://github.com/brimsec/brim/issues/1076 while trying to build locally on my windows machine, btw: if you try to repro this on your Windows VM, you'll need to run `choco install mingw` and edit package.json to use `mingw32-make.exe` instead of `make`.

jameskerr:
@alfred-landrum thanks for doing the windows debugging. I'll see if I can get babel to make those files.

mason-fish:
Awesome! I've been playing with this branch for a couple days now and everything seems solid to me. For such a large change I think we will have to lean on our tests a bit, but then again this is only changing the typing layer which has always been superimposed on the actual js code. So seems good to merge to me 👍

mason-fish (approved):

alfred-landrum:
The windows changes work for me locally, and I verified that the git history for the files locally (not in github) reflects the edits:
`924 files changed, 6167 insertions(+), 35460 deletions(-)`
:shipit: from me

jameskerr:
Fixes #1036

Issue #1055, opened 2020-09-05
Option to "Delete All" Spaces


philrz:
A community user asked:

> Can you add an option to "delete all" in the Spaces picker (like there is for History)? I must delete it one by one.

There's already an "Are you sure...?" pop-up when the user tries to delete any single Space, so we'll make sure a similar pop-up appears for the "Delete All".

![image](https://user-images.githubusercontent.com/5934157/93272155-bd6ebf80-f769-11ea-9a46-87138f78ad40.png)

philrz:
Side note: The group discussion on this topic led to a suggestion that we do a similar "Are you sure...?" pop-up for the "Delete All" that's available in the History panel. @mason-fish has another PR open where he's going to slip that in.

philrz:
Verified in Brim commit `0b26536`.

Per the attached video, now right-clicking on any Space reveals a "Delete All" option. If clicked, the user is shown an "Are you sure?" prompt, then if they confirm, all the Spaces are deleted.

[Verify.zip](https://github.com/brimsec/brim/files/5285865/Verify.zip)

Thanks @mason-fish!

PR #1078, opened 2020-09-15
1021 investigations space filter

### LINKED ISSUE #1021: Filter investigations by space
Currently we collect history entries (or investigations) globally and show them all at once regardless of what connection or space you are on. Clicking on an investigation will rerun the search in the space that it had originally been run in. If you are working across multiple spaces then this can cause unwanted context switches.

Instead of showing all of the entries, let's filter them by space. Later, we can also add a "Saved Queries" section to the left sidebar which are space-agnostic and may be run against you current context.

mason-fish:
fixes #1021, #1055 

Organizes investigations by connectionId and spaceId!

brim


src/js/state/migrations/202009121941_refactorInvestigations.ts
```diff
@@ -0,0 +1,19 @@
...
+  for (const s of getAllStates(state)) {
+    const newState = {}
+    s.investigation.forEach((f) => {
+      if (!f) return
+      if (!newState[f.search.spaceId]) newState[f.search.spaceId] = []
+
+      newState[f.search.spaceId].push(f)
+    })
+
+    s.investigation = {
```
mason-fish, line 13: Since we don't know the connId associated with a given finding, the *best we can do is just put them all on `localhost:9867` and hope that since the new connection feature is so new that folks haven't yet leveraged it :( 

*this got me thinking, @jameskerr we had discussed organizing the state as I've done in these commits 
```
"investigations": {
    "connId": {
      "spaceId": [{finding}]
    }
}
```
...which allows us to remove all of a connection's spaces' investigations more simply/quickly, BUT it creates this migration problem AND given the way we are generating spaceIds we should actually be able to rely on them all being unique across connections anyway.

So now I'm thinking of refactoring the state to the following, and on removing a connection we instead pay the cost of removing each of the connection's spaces/investigations sequentially

```
"investigations": {
  "spaceId": [{finding}]
}
```

wdyt?

mason-fish: We decided offline to keep as is for the benefits mentioned above, and assume that folks have not created new connections

src/js/flows/deleteSpaces.ts
```diff
@@ -0,0 +1,32 @@
...
+import Investigation from "../state/Investigation"
+
+const deleteSpaces = (ids: string[]): Thunk => (
+  dispatch,
+  getState,
+  {globalDispatch}
+) => {
+  const zealot = dispatch(getZealot())
+  const clusterId = Current.getConnectionId(getState())
+  return Promise.all(
```
jameskerr, line 14: It'd be awesome if this could just be.

```js
Promise.all(ids.map(id => dispatch(deleteSpace(id))))
  .catch(e => handle(e))
  .finally(() => dispatch(refreshSpaceNames()))
```

src/js/state/Investigation/selectors.ts
```diff
@@ -1,12 +1,23 @@
...
 
-export default {
-  getInvestigation(state: State) {
-    return state.investigation
-  },
+type Id = string | null
 
-  getCurrentFinding(state: State) {
-    return last(state.investigation)
+export const getInvestigation = (connId: Id, spaceId: Id) => (
```
jameskerr, line 7: What do you think about a `Current.getInvestigation()` method as well, instead of having to get all these spaces and connection ids?

src/js/state/Investigation/reducer.ts
```diff
@@ -4,46 +4,46 @@ import {Finding, InvestigationAction, InvestigationState} from "./types"
...
+import produce from "immer"
 
-const init: InvestigationState = []
+const init = (): InvestigationState => ({})
 
-export default function reducer(
-  state: InvestigationState = init,
-  a: InvestigationAction
-): InvestigationState {
+export default produce((draft, a: InvestigationAction) => {
```
jameskerr, line 11: 👍

jameskerr (changes requested):
Great improvement to the UX to now have the history correspond to each space. I have a few suggestions. Thanks!

src/js/components/FilterTree.tsx
```diff
@@ -16,35 +14,54 @@ import FilterNode from "./FilterNode"
...
-    ["data", "finding", "search", "spaceName"],
-    ""
-  )
   const menu = usePopupMenu([
     {
       label: "Delete",
       click: () => {
-        const multiTs = node.mapChildren((node) => node.data.finding.ts)
-        globalDispatch(Investigation.deleteFindingByTs(multiTs))
+        remote.dialog
```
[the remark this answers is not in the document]
philrz: Yes, @jameskerr's take matches mine.

src/js/state/Investigation/selectors.ts
mason-fish: I had thought of that but saw that we needed the connId and spaceId for deleting individual and groups of investigations anyway so decided to just keep current focused on the current context (defining "context" as a connection + space)

src/js/components/Investigation/FindingCard.tsx
```diff
@@ -27,47 +22,55 @@ const StyledMagnifyingGlass = styled(MagnifyingGlass)`
...
+            title: "Delete History Entry",
+            message: `Are you sure you want to delete this history entry?`,
+            buttons: ["OK", "Cancel"]
+          })
+          .then(({response}) => {
+            if (response === 0)
+              globalDispatch(
+                Investigation.deleteFindingByTs(connId, spaceId, finding.ts)
+              )
+          })
```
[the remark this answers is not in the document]
mason-fish: I found that the recursive functionality of the "FindingTree" nodes vs the iterative "FindingCard" made consolidation a little tricky. This menu, for example, I had thought to pass in from both components' parent but since the "Delete" functionality is different between each (FindingCard just deletes that one entry, where the NodeRow first finds its children and then deletes the whole branch) I decided to leave as is. I also thought to just pass in the common menu element (Delete All, which functions the same for both) but ultimately thought that keeping each array of menu logic together within the corresponding component and having a single piece of duplication was better than breaking the logic across component boundaries. Lmk if you prefer the latter though!

jameskerr (approved):
Right on! Your explanations make sense.

PR #1084, opened 2020-09-16
fix modal close

### LINKED ISSUE #1080: Provide UI indication for newConnection status check
On the newConnection modal, after hitting 'Connect', we keep the modal open until the status check succeeds (and if it doesn't we display errors within the modal), but during this status check there is no indication that anything is happening. Let's add a spinner next to the connect button while we wait for the response.

mason-fish:
fixes #1080 

I added some more context in ticket ^^

brim

### DIFF
--- a/src/js/components/ModalBox/ModalBox.tsx
+++ b/src/js/components/ModalBox/ModalBox.tsx
@@ -34,7 +34,7 @@ export default function ModalBox({name, children, ...props}: ModalBoxProps) {
   }
 
   return (
-    <Animate show={active === name} enter={enter} exit="reverse">
+    <Animate show={active === name} enter={enter}>
       <ModalContents {...props}>{children}</ModalContents>
     </Animate>
   )
--- a/src/js/components/NewConnectionModal.tsx
+++ b/src/js/components/NewConnectionModal.tsx
@@ -89,6 +89,7 @@ export default function NewConnectionModal() {
   const dispatch = useDispatch()
   const [f, formRef] = useCallbackRef()
   const [errors, setErrors] = useState([])
+  const [isFetching, setIsFetching] = useState(false)
   const config: FormConfig = {
     host: {
       name: "host",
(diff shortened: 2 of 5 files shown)


src/js/components/NewConnectionModal.tsx
```diff
@@ -110,22 +111,25 @@ export default function NewConnectionModal() {
...
       }
     },
-    [f, config]
+    [f, config, setIsFetching]
   )
 
   const buttons = [
     {label: "Cancel", click: (closeModal) => closeModal()},
-    {label: "Connect", click: onSubmit}
+    {label: "Connect", click: onSubmit, showSpinner: isFetching}
```
jameskerr, line 132: Instead of adding a `showSpinner` option, why not add a more generic "icon" option where you could inject the MacSpinner.

This does make me think about buttons with a loading state. I'd like to design a first-class "loading" state into the buttons. I always feel comforted as a user when I see those in other apps.

src/js/components/NewConnectionModal.tsx
```diff
@@ -110,22 +111,25 @@ export default function NewConnectionModal() {
...
+        } finally {
+          setIsFetching(false)
         }
         closeModal()
       } else {
         setErrors(form.getErrors())
       }
     },
-    [f, config]
+    [f, config, setIsFetching]
```
jameskerr, line 127: I don't think setIsFetching is necessary to include the deps here. I think the setters won't get stale like the values will.

src/js/components/ModalBox/ModalBox.tsx
```diff
@@ -34,7 +34,7 @@ export default function ModalBox({name, children, ...props}: ModalBoxProps) {
   }
 
   return (
-    <Animate show={active === name} enter={enter} exit="reverse">
+    <Animate show={active === name} enter={enter}>
```
jameskerr, line 37: Yes, a faster "exiting" state is nice.

jameskerr:
Here's a design I like for the loading state of a button: https://www.sketch.com/s/4b2957a2-910e-403c-9fda-617de055c061/a/OEY3dp#Inspector

image

jameskerr (changes requested):
A few suggestions.

jameskerr (approved):
Very cool! Thanks Mason

PR #1091, opened 2020-09-22
refresh all connection spaces

### LINKED ISSUE #1079: Window load does not set host and port query params
The result of this is that if you close a window that was looking at a non-default connection's space and re-open, it loads up the default zqd into zealot but then cannot find the previously viewed space and the user sees this

mason-fish:
fixes #1079 

When loading up the app from a previously saved state, we have still been assuming that the connection will either default to localhost:9867, or that the app has just created a new connection, but it was not properly handling the case where it had already created a new connection from a previous session. This PR aims to fix that by checking for and using any previously set connectionIds and will also populate spaces for all known connections.

### DIFF
--- a/src/js/flows/refreshConnectionsSpaceNames.ts
+++ b/src/js/flows/refreshConnectionsSpaceNames.ts
@@ -1,0 +1,18 @@
+import {Thunk} from "../state/types"
+import Clusters from "../state/Clusters"
+import Spaces from "../state/Spaces"
+
+export default function refreshConnectionsSpaceNames(): Thunk<Promise<void[]>> {
+  return (dispatch, getState, {globalDispatch, createZealot}) => {
+    const clusters = Clusters.all(getState())
+    return Promise.all(
+      clusters.map((c) => {
+        const zealot = createZealot(`${c.host}:${c.port}`)
+        return zealot.spaces.list().then((spaces) => {
+          const addSpaces = spaces || []
+          if (c.id) globalDispatch(Spaces.setSpaces(c.id, addSpaces))
+        })
+      })
+    )
+  }
+}
--- a/src/js/initializers/initIpcListeners.ts
+++ b/src/js/initializers/initIpcListeners.ts
@@ -44,10 +44,6 @@ export default (store: Store) => {
     ipcRenderer.send(channel, getPersistable(store.getState()))
   })
 
-  ipcRenderer.on("getState", (event, channel) => {
-    ipcRenderer.send(channel, getPersistable(store.getState()))
-  })
-
   ipcRenderer.on("showPreferences", () => {
     store.dispatch(Modal.show("settings"))
   })
--- a/src/js/initializers/initZqdConnection.ts
+++ b/src/js/initializers/initZqdConnection.ts
@@ -22,7 +22,9 @@ export default async function(store: Store) {
       username: "",
       password: ""
     }
-    await globalDispatch(Clusters.add(cluster))
+
+    store.dispatch(Clusters.add(cluster))
+    globalDispatch(Clusters.add(cluster))
     store.dispatch(Current.setConnectionId(cluster.id))
   }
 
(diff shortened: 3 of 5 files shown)


src/js/initializers/initIpcListeners.ts
```diff
@@ -44,10 +44,6 @@ export default (store: Store) => {
     ipcRenderer.send(channel, getPersistable(store.getState()))
   })
 
-  ipcRenderer.on("getState", (event, channel) => {
-    ipcRenderer.send(channel, getPersistable(store.getState()))
-  })
-
```
jameskerr, line 50: I am curious about the deletion here. It is used currently for gathering up all the window states when you click "Save Session for Testing Migrations".

src/js/flows/refreshConnectionsSpaceNames.ts
```diff
@@ -0,0 +1,18 @@
+import {Thunk} from "../state/types"
+import Clusters from "../state/Clusters"
+import Spaces from "../state/Spaces"
+
+export default function refreshConnectionsSpaceNames(): Thunk<Promise<void[]>> {
```
jameskerr, line 5: Since we are refreshing all the data about the space and not just the names anymore, a better name might be "refreshSpacesForAllConnections" ?

src/js/initializers/initZqdConnection.ts
```diff
@@ -3,26 +3,32 @@ import {initSpace} from "../flows/initSpace"
...
+    const clusterId = `${clusterHost}:${clusterPort}`
+    const cluster = {
+      id: clusterId,
+      host: clusterHost,
+      port: clusterPort,
+      username: "",
+      password: ""
+    }
+
+    store.dispatch(Clusters.add(cluster))
```
jameskerr, line 26: It looks like there is a case when we'll never get a cluster variable defined?

Is this the logic?

1. If the host/port was provided in the url params, create a connection from them, add it, then set it to current.
2. Else if there was a current connection already, do nothing more in this file
3. Else create a default host and port, add it, then set it to current

Actually, do we always need to create the default one here?

jameskerr (changes requested):

src/js/initializers/initIpcListeners.ts
jameskerr: https://github.com/brimsec/brim/blob/master/src/js/electron/menu/appMenu.ts#L241

mason-fish: Oh it's a duplicate, look at the lines right above it ;)

src/js/initializers/initZqdConnection.ts
mason-fish: Yes, exactly. And I think we do need the default for the first time anyone loads the app.

src/js/initializers/initIpcListeners.ts
jameskerr: Oh right 😅, thanks!

src/js/initializers/initZqdConnection.ts
jameskerr: This conditional logic was getting hard to understand at first glance. What do you think about a version like this where we extract some of the connection setup into a new function?

```ts

const setupConnection = (host, port) => (dispatch, _, {globalDispatch}) => {
  const cluster = {
    host,
    port,
    id: [host, port].join(":"),
    username: "",
    password: ""
  }
  dispatch(Clusters.add(cluster))
  globalDispatch(Clusters.add(cluster))
  dispatch(Current.setConnectionId(cluster.id))
}

export default async function(store: Store) {
  const {space, host, port, id} = getUrlSearchParams()
  global.windowId = id

  const existingConnection = Current.getConnection(store.getState())

  if (host && port) {
    store.dispatch(setupConnection(host, port))
  } else if (!existingConnection) {
    store.dispatch(setupConnection("localhost", "9867"))
  }

  await store.dispatch(refreshConnectionsSpaceNames())

  const spaceId = space || Current.getSpaceId(store.getState())

  if (spaceId) store.dispatch(initSpace(spaceId))
}
```

mason-fish: Sure thing, will use this instead!

jameskerr (approved):
Right on

PR #1106, opened 2020-10-01
upgrade electron 7.3.2 -> 10.1.2

### LINKED ISSUE #1044: Electron upgrade
We should update Electron to most recent stable release, v10.1.1 currently. @jameskerr notes that last time we attempted an Electron upgrade, we had some issues with the Spectron framework, so the Electron update may require updating several other dependencies. (https://github.com/electron-userland/spectron/pull/631)

mason-fish:
fixes #1044 

Upgrades Electron and Spectron by 3 major versions. Also of note, is the underlying webdriverio dep is upgraded by 2 major versions (4.xx -> 6.xx).

Here are some of the breaking changes:

**Electron**
* `remote` module is disabled by default now, and must be explicitly enabled for each browser window, this module is flagged to be deprecated and moved. More information on it's proposed future [here](https://github.com/electron/electron/issues/21408)
* as of electron 9, sending complex objects over ipc will now throw an error. Previously the pieces which could not be serialized would just be silently dropped and was the source of a few bugs we've encountered in the past. So this is a welcome change, we now add `.serialize()` and `deserialize()` methods to classes when we have to send them over the ipc wire. This is also a useful pattern for persisting app state, since we encounter the same problem there.

**Spectron**
* It's difficult to say how much of this is spectron and how much is chromedriver, but there are several new OS dependencies we need to install on centOS for our iTests to even begin to run. Thanks @jameskerr for chasing that down!
* We also need to set the "--no-sandbox" flag since our tests are being run by the `root` user inside our CI container.
* Previous versions of Spectron would create the directory provided as `webdriverLogPath` if it didn't exist but the current version does not, now we create it ourselves

**Webdriverio** (provides the primary api we use to programmatically interact with a browser)
* Entire api has been "promisified", in some cases this allows us to simply `await apiMethod()` instead of using our own retry layers. Though this is not always the case.
* api has also been refactored from `browser.method(selector, methodOptions)` -> `browser.$(selector).method(methodOptions)` in terms of the method input pattern. Since we are using `async` mode, this also means each method must be awaited/handled as a promise.
* there are also several name changes, and in some cases methods have completely been removed and require us to write our own, see the changelist [here](https://github.com/webdriverio/webdriverio/blob/v5/CHANGELOG.md#v500-2018-12-20)


zealot/index.js
```diff
@@ -0,0 +1,42 @@
+// Copyright 2018-2020 the Deno authors. All rights reserved. MIT license.
```
jameskerr, line 1: I think this file can be deleted now. The zealot stuff lives in dist/index.js

jameskerr (approved):
Just one comment to delete the old zealot/index.js but then ship it!

PR #1111, opened 2020-10-05
UnexpectedServerResponseError if NDJSON is Malformed

### LINKED ISSUE #992: Better handling of non-NDJSON in NDJSON response stream
In working on #1087, the team was surprised by the fact that the text stack trace appeared in the middle of the ZJSON response. We've fixed issues in this area on the `zqd` side, such as to not have such text show up in ZJSON (https://github.com/brimsec/zq/issues/1098) and to make sure the stack traces go to the `zqd` logs (https://github.com/brimsec/zq/issues/1097).

In addition to these, @jameskerr also had the good idea of putting in some kind of extra protection at the Brim app level such that if non-NDJSON should leak out in the future, we'd handle it better than we saw in #1087 where it caused the opaque error message `Unexpected token p in JSON at position 31`.

In a group discussion regarding this, @jameskerr indicated that we should likely be able to output the first block of non-NDJSON text that's seen in the stream. This would be very helpful to display to the user since that would allow them to quickly communicate the nature of the error to us via cut & paste or screenshot while still in the app.

jameskerr:
Fixes #992 

Based off the `zjson` branch, tracked in: #1108 

If there is an error on the server and some non-ndjson appears in a response stream, the error used to be:

```
SyntaxError: JSON Parse error: Unexpected identifier "p"
```

With this change, it is now:

```
UnexpectedServerResponse: Expected ndjson but received "panic: Major problem here man, no json in sight
  - src/js/components/bad.ts
  - src/js/components/real.ts
  - src/js/components/this.ts
"
```

Unit tests have been added to ensure the fix is long-lived.


mason-fish (approved):
Looks good! I trust the error from zqd is not verbatim with what we are mocking in the tests :P

jameskerr:
@mason-fish haha no, it's made up.  It did start with "panic:" though : )

PR #1112, opened 2020-10-05
Handle PcapPostWarning in Brim

### LINKED ISSUE #1028: app should handle pcap post warnings
Zqd has a new API message with type field "PcapPostWarning" that has the same form and use as LogPostWarning.  The app should be ready to receive this on PcapPostRequest.

The warning shouldn't be modal and interrupt the user but there should be a visual cue in the UI.

This was added to delay with funkiness we're finding in various pcaps and the warnings could be useful in diagnosing problems from users who can't share their pcaps with us.

jameskerr:
Fixes #1028 

Based off #1108 (that must go in first, but you can look at the last commit here to see the changes)

The app will now respond to PcapPostWarning the same way that it responds to LogPostWarning. The warning will appear in the bottom right of the page after ingest has finished.

A unit test has been added to avoid regressions.

### DIFF
--- a/src/js/flows/ingestFiles.test.ts
+++ b/src/js/flows/ingestFiles.test.ts
@@ -129,4 +129,19 @@ describe("error case", () => {
     expect(Spaces.getSpaces(cluster)(state)).toEqual([])
     expect(Current.getSpaceId(state)).toEqual(null)
   })
+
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
Great, the back and FE now provide and display warnings for pcaps too 👍

PR #1123, opened 2020-10-08
add specific test for alpha and post-alpha zng

alfred-landrum:
Move the integration tests file "sample.zng" to "sample.azng" since its a alpha binary zng file, and add a new sample.zng that's post alpha.


alfred-landrum:
fyi, the test failure is a spurious problem while checking markdown links.

jameskerr (approved):
Makes sense to me

PR #1134, opened 2020-10-13
Pop Notice

jameskerr:
This was inspired by a bug in the integration tests that @alfred-landrum discovered on master in `zar.test.ts`. He noticed that the integration tests were not waiting for the ingest to finish before moving on to other parts of the test.

So I've added a new component called `PopNotice` that pops up a small message at the bottom of the screen. 

image

The integration tests now wait until this message appears before moving on through the tests.

**Technical Details**

To create a new pop notice you imperatively call the `popNotice("message")` function. Since it is such a temporary piece of UI, I think it's easier to understand and use in a thunk or effect if it's imperative instead of declarative like most UI in react.

It also has these constrains:
* On a string message is accepted
* The dismiss "x" is always displayed
* It always is automatically dismissed and unmounted after 3 seconds

```
dispatch(ingestFile("sample.tsv")).then(() => popNotice("Import complete.")
```

We could reuse this pop notice other places. In this PR I've added one after deleting a space, and after deleting all the spaces.


mason-fish (approved):
Looks great! Feels like this new notice will be handy in other areas of the app moving forward 👍

PR #1137, opened 2020-10-13
update brim/ast.js to work with recent zq ast changes

### LINKED ISSUE #1131: brim test failures after field model zq changes
`npm run test` fails after the zq side changes for this PR:

https://github.com/brimsec/zq/pull/1468

The CI automation automatically created a brim side branch with zq pointing to that change, so this is an easy branch to pull & run the tests to see the failure:
https://github.com/brimsec/brim/pull/1127

mccanne:
This commit updates the ast logic in brim/ast.js to deal with recent ast changes in zq.

I put an XXX comment regarding the logic around using "." in field names.

Not urgent, but fieldExprToName() should check for non-identifier characters 
and build a field access using `rec[".@weird"]` instead of building `rec..@weird`.

Fixes #1131


jameskerr (approved):
Cool with me. Just curious, what does an ast look like now with `lhs` and `rhs`?

alfred-landrum:
@mccanne : I'm going to merge this now since it's green & approved, so I can update & get #1138 in.

PR #1138, opened 2020-10-13
fixes for brim integration tests

alfred-landrum:
This Brim tests looks for an error message that changed in brimsec/zq#1461. 

While we should update this test to work differently, this change should allow tests to pass against the zq master after #1137 lands. I'll update this PR then to point zq at zq master.


jameskerr (approved):
Right on

alfred-landrum:
@jameskerr : this is now on top of #1134 so that it includes the fixes from there.

PR #1142, opened 2020-10-14
Flatten columns when comparing to table columns

### LINKED ISSUE #1141: "Pivot to logs" grayed out for entry containing dotted record field
I've narrowed this down to having shown up with Brim commit `c3e0caa` (associated with #1108, cc: @jameskerr), and the symptom is still present through current `master` (commit `5f46417` at the moment.)

If I create an aggregation table that contains a dotted record field such as `id.resp_p`, when I right click with intent to click "Pivot to logs", I find the option is grayed out.

image

I don't have this problem with top-level fields, such as `method`.

image

jameskerr:
In the ZJSON changes, we retained the nested-ness of the data, but added a "flatten" method to convert all the nested field to be flat with "." separator like Zeek. We use this flatten method for the columns in the results table.

The "pivot to logs" field menu option is only enabled when the columns in the table match the columns in the right-clicked record's. This bug revealed that the records columns were still nested. The fix is to flatten first, then compare.

Bug example:
```
Table columns: ["count", "id.resp_h"]
Record columns: ["count", "id"]
```

Fixes #1141

### DIFF
--- a/src/js/electron/menu/searchFieldContextMenu.ts
+++ b/src/js/electron/menu/searchFieldContextMenu.ts
@@ -21,9 +21,9 @@ export default function searchFieldContextMenu(
     const isConn = log.try("_path")?.toString() === "conn"
     const isGroupBy = hasGroupByProc(program)
     const isIp = ["addr", "set[addr]"].includes(field.data.getType())
-
     const hasCol = columns.includes(field.name)
-    const sameCols = isEqual(log.type.map((d) => d.name).sort(), columns.sort())
+    const flatColNames = log.flatten().getColumnNames()
+    const sameCols = isEqual(flatColNames.sort(), columns.sort())
     const hasPackets = space && space.pcap_support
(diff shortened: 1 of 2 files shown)


src/js/electron/menu/searchFieldContextMenu.ts
```diff
@@ -21,9 +21,9 @@ export default function searchFieldContextMenu(
     const isConn = log.try("_path")?.toString() === "conn"
     const isGroupBy = hasGroupByProc(program)
     const isIp = ["addr", "set[addr]"].includes(field.data.getType())
-
     const hasCol = columns.includes(field.name)
-    const sameCols = isEqual(log.type.map((d) => d.name).sort(), columns.sort())
+    const flatColNames = log.flatten().getColumnNames()
```
mason-fish, line 25: 👍 nice catch

mason-fish (approved):

PR #1143, opened 2020-10-14
Pin Bugs

### LINKED ISSUE #1140: Bad pin starting at commit 7046367
I came across this while going through our regular demo flow for the first time in a while. I narrowed down the symptom as appearing starting at Brim commit `7046367` (associated with #1049, cc: @jameskerr) but it reproduces the same in current Brim `master`, which is commit `5f46417` at the moment.

Repro is shown in the video below and described in the steps.

[Repro.zip](https://github.com/brimsec/brim/files/5378860/Repro.zip)

Repro steps:

1. Start with the Zeek logs associated with the common `all.pcap` test data
2. Create individual pinned items for search terms `192.168.0.51` and `_path=http`
3. Execute variations on additional search terms `GET`, `OPTIONS`, `POST`
4. Change the History mode from Linear to Tree
5. Click on the pin for `192.168.0.51`, change it to `192.168.0.53`, and re-execute the search

The problem you can see in the History panel is that the lowest level search term now shows `192.168.0.53`. Prior to commit `7046367`, that same entry would show `POST`.

jameskerr:
fixes #1140 

**Bug Reason**

This bug appears when editing pins. When you edited a pin, then submited the search, we generated an incorrect "search record" object and saved it to history.

If the search bar looked like this: `"http" (pinned), "GET" (search bar)`, then you clicked on "http" and changed it to "http2", then clicked submit, the search record was:

```
program: "http2"
pins: ["http2"]
```

This PR checks if you are editing a pin and will now return the correct search record:

```
program: "GET",
pins: ["http2"]
```

**Future Implementation**

I think in the future we ought to do away with the notion of "previous" and "current" and "pins". All zql fragments could always be in an array. Then you'd have an index of which "fragment" is loaded in the search bar.

```
parts: ["10.10.10.10", "http", "GET"]
editing: 2
```

Then if you click a pin, we just change the editing index:

```
parts: ["10.10.10.10", "http", "GET"]
editing: 0
```

A search bar with no pins would be:

```
parts: ["count() by _path"]
editing: 0
```

**Style Fixes**

I noticed the "remove" buttons for the pins were not styled well so I've updated that here. I also noticed the "pin" button was not aligned so I've updated it too.

Before:
Screen Shot 2020-10-14 at 12 50 51 PM

After:
Screen Shot 2020-10-14 at 12 48 13 PM


src/js/state/Search/selectors.ts
```diff
@@ -24,13 +24,25 @@ export default {
...
+    if (searchBar.editing === null) {
+      return {
+        program: searchBar.current,
+        pins: SearchBar.getSearchBar(state).pinned,
+        spanArgs: Tab.getSpanArgs(state),
+        spaceName: space.name,
+        spaceId: space.id,
+        target: SearchBar.getTarget(state)
+      }
+    } else {
```
mason-fish, line 37: nit: no need for this `else` block with the returning `if` above

mason-fish (approved):
looks good! just left one nit

src/js/state/Search/selectors.ts
jameskerr: True dat, I like the else here though. Feels symmetrical 😆

PR #1144, opened 2020-10-14
Fix the broken link

jameskerr:
So that all our builds will run.


philrz (approved):

jameskerr:
This was fixed in Al's PR #1138

PR #1148, opened 2020-10-15
Keep Brim on Screen

### LINKED ISSUE #1041: Offer a "Move Brim to Current Display" option
@orochford reported that when he removed his laptop from two external displays and then tried to open Brim, and it's no longer visible. He pointed out that apps like Slack offer a right-click option that can repair this, which leads us to guess that maybe this is a common problem with Electron apps.

image

In past situations when this has happened to me with other Windows apps I'd had success hitting Alt-Tab repeatedly until I was effectively "on" the window of the app I couldn't see, then hit Windows-LeftArrow or Windows-RightArrow enough times to step it onto my visible desktop. However, @orochford reports that did not work for him, so I've pointed him at the [Troubleshooting article for clearing app state](https://github.com/brimsec/brim/wiki/Troubleshooting#brim-seems-unable-to-restart-normally-such-as-after-a-bad-crash) to get him unstuck.

jameskerr:
Fixes #1041 

The ticket requests a new task be added to the “[Jump List](https://www.electronjs.org/docs/tutorial/windows-taskbar)” on a Windows OS. 

Clicking these tasks re-executes the brim exe file with specified arguments. By default this creates more than one instance of the Brim app, meaning there are multiple main processes and multiple “global” stores. I didn’t know this, since I develop on a Mac which maintains a single instance for us. However, if you open a new terminal and manually run the brim executable again, it will indeed start a new instance, even on Mac.

Electron provides the “[instanceLock](https://www.electronjs.org/docs/api/app#apprequestsingleinstancelock)” api to help with this. Only one instance of the app can have the lock, so any other instances will immediately quit and trigger a “[second-instance](https://www.electronjs.org/docs/api/app#event-second-instance)” event on the primary instance.

This is now how we handle the Windows JumpList tasks.

Clicking on the “Move Brim to Current Display” Task will move all Brim windows over to the current display and stack them so that you can see each one. 

Screen Shot 2020-10-16 at 1 33 42 PM

Screen Shot 2020-10-16 at 1 35 23 PM

I’ve created a SearchWindow class that is responsible for creating the window and making sure that it will fit on one of the screens. This prevents windows from hiding way off screen if a user unplugs a display. 

Lastly, opening a new brim window will use the `stack` function to automatically position it 25 pixels below and to the right of the previously focused window (before it would have the same size and position so you couldn’t see the previously focused window).

Screen Shot 2020-10-16 at 1 30 54 PM

**PS** 
I’ve withheld some refactors here in favor of getting it done faster, but I am excited to make a BrimApp class to contain many of the variables in `main.ts` and the ways they need to interact. 

I also have some ideas on how to simplify the “WindowManager” class. More to come.

### DIFF
--- a/src/js/electron/tron/window.ts
+++ b/src/js/electron/tron/window.ts
@@ -1,4 +1,6 @@
 import {BrowserWindow} from "electron"
+import {dimensFromSizePosition} from "../window/dimens"
+import {SearchWindow} from "../window/SearchWindow"
 
 import {WindowName} from "./windowManager"
 
@@ -24,33 +26,14 @@ export default function window(name: WindowName, params: WindowParams) {
 
 function searchWindow(params) {
   const {size, position, query, id} = params
-  const win = new BrowserWindow({
-    titleBarStyle: "hidden",
-    resizable: true,
-    minWidth: 480,
-    minHeight: 100,
-    webPreferences: {
-      nodeIntegration: true,
-      experimentalFeatures: true,
-      enableRemoteModule: true
-    }
-  }).on("close", (e: any) => {
+  const dimens = dimensFromSizePosition(size, position)
+  const win = new SearchWindow(dimens, query, id).ref
+
+  win.on("close", (e: any) => {
     // Close handled by the search renderer
     e.preventDefault()
     e.sender.webContents.send("close")
   })
-
-  if (size) {
-    const [width, height] = size
-    win.setSize(width, height)
-  }
-  if (position) {
-    const [x, y] = position
-    win.setPosition(x, y)
-  } else {
-    win.center()
-  }
-  win.loadFile("search.html", {query: {...query, id}})
 
   return win
 }
--- a/src/js/electron/userTasks.ts
+++ b/src/js/electron/userTasks.ts
@@ -6,21 +6,29 @@ export default function(app: any) {
     app.setUserTasks([
       {
         program: process.execPath,
-        arguments: getArguments(),
+        arguments: getArguments("--new-window"),
         iconPath: process.execPath,
         iconIndex: 0,
         title: "New Window",
         description: "Create a new window"
+      },
+      {
+        program: process.execPath,
+        arguments: getArguments("--move-to-current-display"),
+        iconPath: process.execPath,
+        iconIndex: 0,
+        title: "Move Brim to Current Display",
+        description: "Move brim windows to the current display."
       }
     ])
   }
 }
 
-function getArguments() {
+function getArguments(arg) {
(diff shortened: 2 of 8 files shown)


src/js/electron/tron/window.ts
```diff
@@ -24,34 +26,15 @@ export default function window(name: WindowName, params: WindowParams) {
...
-    resizable: true,
-    minWidth: 480,
-    minHeight: 100,
-    webPreferences: {
-      nodeIntegration: true,
-      experimentalFeatures: true,
-      enableRemoteModule: true
-    }
-  }).on("close", (e: any) => {
+  const dimens = dimensFromSizePosition(size, position)
```
mason-fish, line 29: I like the dimens refactoring here and below, it was becoming a bit unwieldy

src/js/electron/userTasks.ts
```diff
@@ -6,21 +6,29 @@ export default function(app: any) {
...
         title: "New Window",
         description: "Create a new window"
+      },
+      {
+        program: process.execPath,
+        arguments: getArguments("--move-to-current-display"),
+        iconPath: process.execPath,
+        iconIndex: 0,
+        title: "Move Brim to Current Display",
+        description: "Move brim windows to the current display."
```
mason-fish, line 21: nit: capitalize Brim like you did for `title`

mason-fish (approved):
Awesome! I am especially a fan of the SearchWindow and dimens refactoring. There are a lot of magic numbers we use in that part of the code base and since it is a bit much to document/define them all then it's at least nice that we can herd them into one place.

src/js/electron/userTasks.ts
jameskerr: Oops yes, thanks!

src/js/electron/userTasks.ts
```diff
@@ -18,7 +18,7 @@ export default function(app: any) {
         iconPath: process.execPath,
         iconIndex: 0,
         title: "Move Brim to Current Display",
-        description: "Move brim windows to the current display."
+        description: "Move Brim windows to the current display"
```
mason-fish, line 21: shoot sorry, I also should have mentioned the inconsistency w "Current Display". Not trying to be capitalization police, they're just right next to each other so I can't not see it haha 🙈

jameskerr: I think the title is like title cased like a book cover, but the description is a normal sentence. So the lower cased words work there.

mason-fish: Oh gotcha, yes that works!

PR #1153, opened 2020-10-16
Move link checker to separate Actions Workflow

philrz:
The automation for the link checker has been doing its job insofar as it's noticed when external links go bad or act flaky. However, it's also been correctly noted that having it run as a Job in the same Action with the rest of `ci.yml` can be troublesome, since it makes it more difficult to debug when there's link problems at the same time as legit code breakages, or CI infra instability.

To address this, here we break out the link checker to its own separate Actions Workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,9 +203,3 @@ jobs:
         with:
           name: artifacts-${{ matrix.os }}-node-${{ matrix.node-version }}
           path: /var/tmp/brimsec/itest
-
-  markdown-link-check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v1
-      - uses: gaurav-nelson/github-action-markdown-link-check@v1
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,14 @@
+name: Link checker
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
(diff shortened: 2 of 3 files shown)


philrz:
In a coincidence that validates why this change is justified [😄 ], a link to an external article failed when the checker was invoked by this PR. As I manually attempted to access the URL in my browser, it continued to fail. I ultimately found the same content at a link on another site and changed the link in this branch in hopes it will be more reliable. And of course by the time _that_ Actions run had completed successfully, the original link was working also. So... 🤷‍♂️ 

This could surely trigger a debate about how many links we should have in our materials to external sources that are out of our control, and/or if we should care if they break. I'm personally of the opinion that:

1. They're useful and provide good context, hence still worthwhile to include.
2. It's good for us to at least know if they fail or are flaky, even if we don't make it our life's work to debug them at every turn.

By having them handled in a separate job here, it seems like the optics of how the automation runs would make it easier to argue for exceptions during occasional link failures. That is, if all the CI runs fine with the exception of a link checker failure, and that failure is attached to a site that's been flaky in the past, or it has the look of a one-time failure that's immediately been confirmed to have passed, reviewers could turn a blind eye to that failure and consent to merging despite a small part of the automation having finished in "red" status.

alfred-landrum (approved):
thanks for doing this!

jameskerr:
Cool, makes sense.

philrz:
Thanks guys! I'll put up something similar for zq next.

Issue #1152, opened 2020-10-16
Disruptive "The user aborted a request" notification


philrz:
Repro is with Brim commit `07dda7a` talking to `zqd` commit `0c6b642`.

In the attached video, I perform rapid-fire queries that take a few seconds to execute, such that the second one starts executing while the first one is in progress. As it shows, when the second one is entered, a yellow pop-up saying "The user aborted a request" is presented, and it remains until it's clicked to Dismiss it. This feels like a bug since being able to start something new before something else finishes feels like a convenience of which the user should not have to be informed, let alone explicitly acknowledge.

[Repro.zip](https://github.com/brimsec/brim/files/5394429/Repro.zip)

philrz:
Verified in Brim commit `f52056a` talking to `zqd` commit `702d877`.

Repeating the original repro steps, now the second query quickly entered no longer triggers a notification.

[Verify.mov.zip](https://github.com/brimsec/brim/files/5411811/Verify.mov.zip)

Thanks @jameskerr!

PR #1155, opened 2020-10-17
Search Aborted Error 

jameskerr:
If you issue a search, then issue another search before the first has completed, Brim will abort the fetch request of the first. This usually throws an error with the message: `DOMException: the user aborted the request`. However, in our app, we don't want to consider that an error case, since it's clearly what the user wanted to do. However, the error was still being thrown because we were not checking for the abort message in all three places an error can happen in a search:

1. The "fetch" promise could reject with an error
2. The zqd server could return an http error code
3. The zqd server could end the search with an "error" key in the TaskEnd payload.

Previously, we were only checking one of those places for the abort error and skipping the error. Now I've made a single `errored` function that we use in all three places.

This will fix: #1152 

I also inlined the callbacks to make it all more readable. I think it's better than scrolling up just to see what the handler does.


src/js/flows/search/handler.ts
```diff
@@ -8,71 +8,53 @@ function abortError(e) {
...
-
-  function warnings(payload) {
-    response.emit("warnings", payload.warning)
-  }
+    const flushBufferLazy = whenIdle(flushBuffer)
+
+    function errored(e) {
+      flushBufferLazy.cancel()
+      if (abortError(e)) {
+        resolve()
```
mason-fish, line 24: 👍

mason-fish (approved):

PR #1157, opened 2020-10-19
Connection names

### LINKED ISSUE #1085: Add and use name field for connections
We are currently using connection id's as the display name, and that id is the `{host}:{port}` combo. Let's add a name field which can default to just use the host but otherwise be assigned any string value.

mason-fish:
fixes #1085 

builds off #1150 

Adds names to connections, and displays that value instead of id.

image

image

image

### DIFF
--- a/bin/generator/index.js
+++ b/bin/generator/index.js
@@ -9,4 +9,11 @@ cli.command("style <name>").action(handleStyle)
 cli.command("reducer <name>").action(handleReducer)
 cli.command("icon <path>").action(handleIcon)
 cli.command("migration <name>").action(handleMigration)
+
+cli.on('command:*', function () {
+  console.error('Invalid command: %s', cli.args[0]);
+  cli.outputHelp()
+  process.exit(1);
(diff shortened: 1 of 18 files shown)


bin/generator/index.js
```diff
@@ -9,4 +9,11 @@ cli.command("style <name>").action(handleStyle)
 cli.command("reducer <name>").action(handleReducer)
 cli.command("icon <path>").action(handleIcon)
 cli.command("migration <name>").action(handleMigration)
+
+cli.on('command:*', function () {
+  console.error('Invalid command: %s', cli.args[0]);
+  cli.outputHelp()
+  process.exit(1);
+});
+
```
jameskerr, line 18: Tight!

jameskerr (changes requested):
My only request is to add a margin-bottom: 12px to the <InputField>. Or transform it into a styled component with a margin-bottom: 12px.

jameskerr (approved):

PR #1159, opened 2020-10-21
Update electron and webdriverio

jameskerr:
Maybe upgrading WebDriver and electron will help with our intermittent test failures. 🤞


mason-fish (approved):
🤞

PR #1163, opened 2020-10-22
update status globally

### LINKED ISSUE #1161: Additional Brim windows show spinners
The repro in the video below is with Brim at commit `a563c51` on Windows, but I've confirmed via binary search that the place where this started misbehaving was commit `8636e29` (associated with #1150, cc: @mason-fish), and I reproduced at that commit on macOS.

What I've observed is that when I select "New Window" (either using the drop-down menu as shown in the video or by selecting the right-click option in the taskbar icon), the new window always appears "active" while the ones I'd open previously all change to show only a spinner.

[Repro.zip](https://github.com/brimsec/brim/files/5419152/Repro.zip)

mason-fish:
fixes #1161 

brim

needed to update status globally

### DIFF
--- a/src/js/flows/initCurrentTab.ts
+++ b/src/js/flows/initCurrentTab.ts
@@ -4,7 +4,11 @@ import {initSpace} from "./initSpace"
 import Clusters from "../state/Clusters"
 import {Thunk} from "../state/types"
 
-export const initCurrentTab = (): Thunk => async (dispatch, getState) => {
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Makes sense, I wonder if there is a more fail safe way we could do global dispatch. Use normal dispatch everywhere, then have a special reducer that forwards the global actions. Maybe?

PR #1164, opened 2020-10-22
Stack windows correctly on non-zero x,y screens

### LINKED ISSUE #1041: Offer a "Move Brim to Current Display" option

jameskerr:
Third try fix #1041 

If a screen had an x or y that was not zero, some parts of the code broke. So we subtract the screen.x from the desired window.x, to effectively make the screen x and y both 0. Then all the math is easier. Then we add back the screen.x at the end.

Added more tests to fix the bugs in the comments of the ticket.

### DIFF
--- a/src/js/electron/window/dimens.test.ts
+++ b/src/js/electron/window/dimens.test.ts
@@ -226,3 +226,38 @@ test("moving to current display", () => {
   const next = stack({x, y, width, height}, screen2, 25)
   expect(next).toEqual({x: 1920 + 25, y: 77 + 25, width: 1250, height: 750})
 })
+
+test("moving to current display when display is smaller", () => {
+  const _screen1 = {x: 0, y: 0, width: 1920, height: 1160}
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
Thanks for including tests, helps understand the logic better!

PR #1167, opened 2020-10-26
Connection detail

mason-fish:
Adds three modals to facilitate operations on connection details

New connection
image

New Connection with errors:
image

Connection details:
image

Editing existing connection:
image


src/js/components/ConnectionModal/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+              conn={mode === "edit" && conn}
+            />
+          </>
+        )}
+      </TextContent>
+    </StyledModalBox>
+  )
+}
+
+export default ConnectionModal
```
jameskerr, line 114: I think the code would be more revealing if you switched on the mode at the very top, then had two different components for each mode. 

Something like:

```js
function ConnectionModal({mode}) {
  if (mode === "edit")
     return EditConnectionModal
  else 
    return ShowConnectionModal
}

```

src/js/components/ConnectionModal/StatusLight.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <circle cx="50" cy="50" r="40" strokeWidth="3" filter="url(#f1)" />
+    </svg>
+  )
+}
+
+export default styled(StatusLight)<{status: ClusterStatus}>`
+  fill: ${(p) => {
+    switch (p.status) {
+      case "connected":
+        return p.theme.colors.green
```
jameskerr, line 22: We should soon decide on the one place to store color variables. We have the same colors in scss vars, the styled-component theme, and in css var(--s).

src/js/components/ConnectionModal/StatusLight.tsx
```diff
@@ -0,0 +1,29 @@
...
+const StatusLight = ({...props}: any) => {
+  return (
+    <svg {...props} viewBox="0 0 100 100" width="10">
+      <defs>
+        <filter id="f1" x="-50%" y="-50%" width="200%" height="200%">
+          <feGaussianBlur in="SourceGraphic" stdDeviation="5" />
+        </filter>
+      </defs>
+      <circle cx="50" cy="50" r="40" strokeWidth="3" filter="url(#f1)" />
+    </svg>
```
jameskerr, line 14: If you didn't want to mess with svgs, I think this could be done using css.

pseudo-css
```
width: 10px;
height: 10px;
border-radius: 50%;
background: $color;
box-shadow: 0 0 5px 3px transparentize($color, a little bit);
```

src/js/state/Clusters/types.ts
```diff
@@ -6,6 +6,7 @@ export type Cluster = {
   username: string
   password: string
   status: ClusterStatus
+  version?: string
```
jameskerr, line 9: Cool!

zealot/src/zealot.ts
```diff
@@ -29,6 +29,9 @@ export function createZealot(
     status: () => {
       return promise({method: "GET", path: "/status"})
     },
+    version: () => {
+      return promise({method: "GET", path: "/version"})
+    },
```
jameskerr, line 34: 👍

jameskerr (changes requested):
This is looking really nice. I like the file organization. Two things I found while testing.

1. It appears that the form validation is not being run when in the Edit Modal. The form submitted just fine when I entered a blank name.

2. It's a little jarring right now when I open the connection detail view, then click "edit". The modal animation replays for the edit modal. I think the edit modal contents should be part of the detail modal body. When the user clicks edit, the contents are just swapped out on the existing modal background.

mason-fish:
Ah I set the default name, if left blank, to just be the same as the host:port as we had before

src/js/components/ConnectionModals/ViewConnectionModal.tsx
```diff
@@ -107,6 +108,7 @@ const ViewConnectionModalContents = ({onClose, onEdit}) => {
   const conn = useSelector(Current.getConnection)
   const spaceIds = useSelector(Spaces.ids(conn.id))
   const spaceCount = spaceIds.length
+  useEnterKey(onClose)
```
jameskerr, line 111: 👍

jameskerr (approved):
🚢📦:shipit:

src/js/zqd/zqd.ts
```diff
@@ -150,9 +150,32 @@ export class ZQD {
...
+        giveUp = true
+      }, 5000)
+
       this.zqd.kill("SIGTERM")
+      while (!giveUp && (await isUp())) {
+        await new Promise((res) => setTimeout(res, 500))
+      }
+
+      if (giveUp) {
+        throw new Error("gave up waiting for zqd to shutdown")
```
jameskerr, line 167:
```suggestion
        log.error("zqd did not shutdown after 5 seconds")
```

src/js/electron/quitter.ts
```diff
@@ -21,8 +22,13 @@ export function handleQuit(manager: $WindowManager, store, session, zqd) {
     }
   })
 
-  app.on("quit", () => {
-    zqd.close()
+  app.on("quit", async () => {
+    try {
+      await zqd.close()
```
jameskerr, line 27: Move this `await zqd.close()` to after manager.quit() and before app.quit() in the "before-quit" handler.

src/js/zqd/zqd.ts
```diff
@@ -150,9 +150,32 @@ export class ZQD {
...
+      }
```
jameskerr, line 168: I think it should just log what happened here.

If giveUp, log.error("zqd failed to shutdown in time")
else log.info("zqd has shutdown")

PR #1168, opened 2020-10-27
fix detail window spinning

mason-fish:
Opening the detail window was also causing connections to re-initialize, but we don't need to do that for this page

### DIFF
--- a/src/js/initializers/initConnectionParams.ts
+++ b/src/js/initializers/initConnectionParams.ts
@@ -2,9 +2,13 @@ import {Store} from "../state/types"
 import Clusters from "../state/Clusters"
 import Current from "../state/Current"
 import getUrlSearchParams from "../lib/getUrlSearchParams"
-import {Cluster} from "../state/Clusters/types"
+import {Cluster, ClusterStatus} from "../state/Clusters/types"
 
-const setupConnection = (host, port) => (dispatch, _, {globalDispatch}) => {
+const setupConnection = (host, port, status: ClusterStatus = "initial") => (
+  dispatch,
+  _,
+  {globalDispatch}
+) => {
   const hostPort = [host, port].join(":")
   const cluster: Cluster = {
     host,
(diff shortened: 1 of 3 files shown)


jameskerr:
I see. If the brim app opens and we have 10 search windows that get opened, they will all reset the connections and run the connection function, right? Is that a problem?

jameskerr:
These problems seem to be arising from mixing the concept of "app level state" with the concept of "window level state". Each window is re-setting the "app-level" connection status.

I wonder if we need to keep the connection status local to each window, and we don't need to persist it, or keep it in the global store.

Either that or, make the connection initialize in the main process and update the global store once on initialize. 

Either of those would solve a lot of these issues that keep coming up, right?

mason-fish:
Ultimately, yes I think you've got the right of it: windows keep resetting app level state.

I like the idea of re-initializing in the main process on start. Or maybe even more simply, we can reset when saving the state on quit.

jameskerr:
After thinking about it, I think the least complex way to solve this to add a ConnectionStatus reducer that only maps a connection id to it's status for that window. It would not be persisted, eliminating the need to tweak state when saving or loading.

Downsides:
* Adding a new reducer
* Window's making duplicate status checks on startup

Upsides:
* Follows existing patterns
* No new patterns to manage reseting the connection statuses like we have now
* If the current window disconnects, all the other open windows don't change (this seems like a plus if you're using those windows as references)
* Will only check the connections as needed, not all of them

I know we discussed this as a possible solution before, but now with a little more experience with the problem, it seems like these downsides are small compared to the win of staying in the existing patterns.

If we did the checks in the main process we'd have these downsides:

* Extra code to use zealot in the node process, or write our own request outside of zealot
* Still need to munge state
* Only would run on app start
* Updating the status would re-render all windows and clearing out whatever context they had

mason-fish:
closing in favor of #1196 which implements the above solution

PR #1171, opened 2020-10-28
A modal dialog experiment

jameskerr:
Here is an experiment for a new way to structure our modals. The goal is to give them more flexibility, while making it easy to still look consistent.

As part of the experiment, I re-did the "Debug Query" modal to use this new "ModalDialog" component. It has an enter and exit animation using CSSTransition, it pops down below the tab bar, it has a "Scrollable" container for content that can get very long, it has a footer component for the row of buttons, and it locks the rest of the screen using "pointer-events: none" on the body element when it's opened.

### DIFF
--- a/src/js/components/ModalDialog/ModalDialog.tsx
+++ b/src/js/components/ModalDialog/ModalDialog.tsx
@@ -19,8 +19,8 @@ const Overlay = styled.div`
 `
 
 const Background = styled.div`
-  --blur-color: hsla(28, 5%, 85%, 0.5);
-  --blur-shadow: hsla(28, 5%, 20%, 0.5);
+  --blur-color: hsla(32, 5%, 85%, 0.5);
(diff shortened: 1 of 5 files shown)


jameskerr:
I don't think this error is related to the changes.

src/js/components/ModalDialog/ModalDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+      document.body.style.pointerEvents = ""
+    }
+    return () => {
+      document.body.style.pointerEvents = ""
+    }
+  }, [show])
+
+  return ReactDOM.createPortal(
+    <Overlay>
+      <CSSTransition
```
mason-fish, line 119: Awesome, excited to get away from the `anime` lib and use this one more

mason-fish (approved):
Love it!

PR #1177, opened 2020-10-29
1022 Style Tweeks

jameskerr:
Styles tweeks to the edit, show, and new connection modals. They look like this now after these tweeks:

Screen Shot 2020-10-29 at 3 24 01 PM
Screen Shot 2020-10-29 at 3 24 05 PM
Screen Shot 2020-10-29 at 3 24 23 PM


mason-fish (approved):
Right on, these lgtm!

PR #1189, opened 2020-11-05
Activate Bug

### LINKED ISSUE #1118: Clicking Brim dock icon brings up an additional window
Starting with GA Brim tagged `v0.18.0`, I've noticed that on macOS when the Brim app is already running and there's a window open, clicking the Brim icon in the dock opens yet another separate window. In GA Brim tagged `v0.17.0` and other earlier versions, this did not happen. With those older versions, clicking the icon in the dock would simply make sure that the Brim window that's already open is maximized and in the foreground, which is more in line with how other macOS apps tend to behave,

Here's a video reproducing the problem in `v0.18.0`:

[Repro.zip](https://github.com/brimsec/brim/files/5341821/Repro.zip)

Here's a video showing the behavior in `v0.17.0`:

[v0.17.0.zip](https://github.com/brimsec/brim/files/5341847/v0.17.0.zip)

jameskerr:
Fixes #1118 

Clicking the Brim icon on Mac now does nothing if there is already a window open (it focuses it). But if there are no open windows, it will create a new one when clicked.

The problem was some unnecessary parens that got added when we migrated to typescript. It tried to fix some where syntax. The first commit is the fix. The second commit is a small refactor and tests to ensure it doesn't appear again.

### DIFF
--- a/src/js/electron/brim.test.ts
+++ b/src/js/electron/brim.test.ts
@@ -1,0 +1,42 @@
+jest.mock("./extensions", () => ({
+  installExtensions: jest.fn()
+}))
+import {installExtensions} from "./extensions"
+import {Brim} from "./brim"
+
+test("activate when zero windows open", () => {
+  const brim = new Brim()
+  expect(brim.windows.count()).toBe(0)
+  brim.activate()
+  expect(brim.windows.count()).toBe(1)
+})
+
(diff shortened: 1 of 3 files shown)


mason-fish (approved):
lgtm 👍

PR #1196, opened 2020-11-06
use connectionStatuses reducer

### LINKED ISSUE #1183: Opening Log Detail window causes main window to spin (commit 8636e29)
I've narrowed down via binary search that this bug seemed to crop up at commit `8636e29` (associated with #1150, cc: @mason-fish).

Repro is with the attached [hello.pcapng.zip](https://github.com/brimsec/brim/files/5478705/hello.pcapng.zip) (unzip before importing). Per the attached video, when I click on an event (such as the `conn` one as shown here) to bring up the Log Detail in a separate window, the main window becomes a never-ending spinner.

[Repro.zip](https://github.com/brimsec/brim/files/5478707/Repro.zip)

mason-fish:
fixes #1183 

Refactors the way we store statuses to be in it's own piece of non-persisting window-level state.

I also noticed a couple of other smaller bugs which I fixed along the way, I'll leave comments next to the corresponding code for those.

brim

### DIFF
--- a/src/js/electron/ipc/windows/messages.ts
+++ b/src/js/electron/ipc/windows/messages.ts
@@ -10,6 +10,9 @@ import {
 } from "../types"
 
 export type NewTabSearchParams = {
+  host: string
+  port: string
+  connId: string
   program: string
   span: SpanArgs
   spaceId: string
--- a/src/js/electron/tron/windowManager.ts
+++ b/src/js/electron/tron/windowManager.ts
@@ -124,7 +124,9 @@ export default function windowManager() {
       winParams: Partial<WindowParams> = {},
       initialState: any = undefined
     ): BrimWindow {
-      const lastWin = last<BrimWindow>(this.getWindows())
+      const lastWin = last<BrimWindow>(
+        this.getWindows().filter((w) => w.name === name)
+      )
       const params = defaultWindowParams(winParams, lastWin && lastWin.ref)
       const id = params.id
 
--- a/src/js/electron/tron/windowManager.ts
+++ b/src/js/electron/tron/windowManager.ts
@@ -97,7 +97,7 @@ export default function windowManager() {
     openSearchTab(searchParams: NewTabSearchParams) {
       let isNewWin = true
       const existingWin = this.getWindows()
-        .sort((a, b) => a.lastFocused - b.lastFocused)
+        .sort((a, b) => b.lastFocused - a.lastFocused)
         .find((w) => w.name === "search")
       if (existingWin) {
         isNewWin = false
@@ -109,7 +109,8 @@ export default function windowManager() {
         return
       }
 
-      const {win} = this.openWindow("search", {})
+      const {host, port} = searchParams
+      const win = this.openWindow("search", {query: {host, port}})
       win.ref.webContents.once("did-finish-load", () => {
         sendTo(
           win.ref.webContents,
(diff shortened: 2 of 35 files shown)


src/js/initializers/initNewSearchTab.ts
```diff
@@ -7,12 +7,13 @@ import SearchBar from "../state/SearchBar"
...
 
 export default function(store: Store, params: NewTabSearchParams) {
-  const {spaceId, span, program, isNewWin} = params
+  const {connId, spaceId, span, program, isNewWin} = params
 
   if (!isNewWin) {
     store.dispatch(Tabs.new())
   }
 
+  store.dispatch(Current.setConnectionId(connId))
```
mason-fish, line 16: this was assuming localhost:9867 before

src/js/electron/tron/windowManager.ts
```diff
@@ -97,7 +97,7 @@ export default function windowManager() {
     openSearchTab(searchParams: NewTabSearchParams) {
       let isNewWin = true
       const existingWin = this.getWindows()
-        .sort((a, b) => a.lastFocused - b.lastFocused)
+        .sort((a, b) => b.lastFocused - a.lastFocused)
```
mason-fish, line 100: this was sorted chronologically resulting in selection of first search window that had been focused on, instead of last

src/js/electron/ipc/windows/messages.ts
```diff
@@ -10,6 +10,9 @@ import {
 } from "../types"
 
 export type NewTabSearchParams = {
+  host: string
```
mason-fish, line 13: These are for when we initiate a search from the detail window (possible from field context menus). When there are no existing search windows to create a new tab on, we already create a new window but now we also pass in the detail window's `host` and `port` as query args so that the search window's initialization can use these instead of defaulting to localhost:9867 for adding a cluster. `connId` will now also be sent along through ipc to the new/existing window much like spaceId already is.

src/js/initializers/initNewSearchTab.ts
jameskerr: Oh very good. Nice catch.

src/js/electron/tron/windowManager.ts
```diff
@@ -190,10 +193,10 @@ export default function windowManager() {
       if (win) {
         win.ref.webContents.send("showPreferences")
       } else {
-        const {win} = this.openWindow("search", {})
+        const newWin = this.openWindow("search", {})
```
jameskerr, line 196: Oh shoot, was this broken before?

src/js/electron/tron/windowManager.ts
```diff
@@ -123,7 +124,9 @@ export default function windowManager() {
       winParams: Partial<WindowParams> = {},
       initialState: any = undefined
     ): BrimWindow {
-      const lastWin = last<BrimWindow>(this.getWindows())
+      const lastWin = last<BrimWindow>(
+        this.getWindows().filter((w) => w.name === name)
+      )
```
jameskerr, line 129: 🙏

src/js/electron/ipc/windows/messages.ts
jameskerr: Awesome

jameskerr (approved):
Awesome Mason, thanks for finding all those bugs in the window manager too. I think TypeScript is confused what "this" is in the window manager. There is a way to annotate "this" and get type errors for things like. I'd like to do that soon.

 

This is not a blocker but just something I noticed. We have a few different "flows" for initializing tabs and connections and spaces and things like that.

* initConnection
* initCurrentTab
* initNewSearchTab
* initConnectionParams

At first look, it's confusing to me (even though I wrote some of them) to know the responsibilities of each. And it looks like the bodies of those functions do some of the same things. Now that we have some solid use cases like (open tab on same window, on new window, from detail view... etc), it's probably time to draw some sequence diagrams to clarify what is happening, then re-arrange some of these methods.

src/js/state/getPersistable.ts
```diff
@@ -5,6 +5,7 @@ export default function getPersistable(state: any) {
     delete draft.errors
     delete draft.notice
     delete draft.handlers
+    delete draft.connectionStatuses
```
jameskerr, line 8: 👍

PR #1200, opened 2020-11-10
retrying connection status

### LINKED ISSUE #1160: Clicking "Retry" on a new tab after zqd restart attempts to GET a null Space
This is a loose end I stumbled onto while verifying #1048. Repro is on Brim commit `a563c51` talking to `zqd` commit `e26ffa3`.

Repro steps as shown in the attached video:

1. Connected successfully to a local `zqd` and view the contents of a Space
2. `kill` the running `zqd` process
3. Click to a new tab and click the Space name to trigger the "The service could not be reached..." message
4. Manually restart `zqd` using the exact same command line that Brim launched it with
5. Click "Retry Now" (as shown in the video ), or let it count down all the way
6. Observe a "Space does not exist" message
7. If the Space name is clicked again manually, now it does come up successfully

[Repro.mov.zip](https://github.com/brimsec/brim/files/5419005/Repro.mov.zip)

Since it came up successfully in step 7, that implies the `zqd` started in step 4 was healthy and we should not have seen the "Space does not exist" message seen in step 6. That's the bug.

I sniffed the traffic in Wireshark, and I see a `GET /space/null` being sent at step 6.

Request:

```
GET /space/null HTTP/1.1
Host: localhost:9867
Connection: keep-alive
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Brim/0.18.0 Chrome/85.0.4183.121 Electron/10.1.4 Safari/537.36
Accept: */*
Sec-Fetch-Site: cross-site
Sec-Fetch-Mode: cors
Sec-Fetch-Dest: empty
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US
```

Response:

```
HTTP/1.1 404 Not Found
X-Request-Id: 1
Date: Wed, 21 Oct 2020 22:33:21 GMT
Content-Length: 77

{"type":"Error","kind":"item does not exist","error":"space does not exist"}
```

mason-fish:
fixes #1160 

adds a "retrying" connection status which is completely managed end to end by the `NetworkErrorNotice` component. While that component is present, the connection which triggered it's rendering will switch to a "retrying" status. This new status is distinct from "disconnected" because it indicates we are actively trying to reconnect for you and we don't want you to lose your previous space and search context.

If the user hits "dismiss" or navigates to a different tab (or any action which would cause the notification to dismount) then the connection status will be switched from "retrying" to "disconnected".

This should now only be triggered when encountering network errors for searches. All promises issued from clients obtained from `getZealot` will now automatically update connection statuses accordingly whenever network errors are encountered. If a flow would like to step outside of this logic, it may create it's own raw zealot client as I am doing in the `checkStatus` flow.


src/js/components/NetworkErrorNotice.tsx
```diff
@@ -15,21 +16,39 @@ const MAX_BACKOFF = 128
 
 export default function NetworkErrorNotice({error}: Props) {
```
jameskerr, line 17: I'm having difficulty understanding the contents of this component now.  What are the use cases that each 'useEffect' call are intended for?

jameskerr (commented):
I had a few nit comments which are just suggestions. Take them or leave them. Functionality looks good.

I just want help understanding the use cases that each "effect" accounts for in the NetworkErrorNotice. Can you clarify that for me?

I see that when it unmounts it always sets it to "disconnected", and I assume there is some handling when the connection changes.

src/js/flows/initSpace.ts
```diff
@@ -27,10 +27,14 @@ export const initSpace = (spaceId: string): Thunk => (dispatch, getState) => {
       if (!space.hasIndex()) {
         dispatch(SearchBar.setTarget("events"))
       }
-      dispatch(submitSearch({history: true, investigation: false}))
+      dispatch(submitSearch({history: true, investigation: false})).catch(
+        (e) => {
+          dispatch(Notice.set(ErrorFactory.create(e)))
+          throw e
+        }
+      )
```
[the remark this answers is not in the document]
mason-fish: I wanted to specifically catch and ~release~ handle the error from search differently than the outer catch-all here. In other words, if search threw an error, we set the notice and then throw the error again so that the outer catch will also log out that error (and if there were any other non-search thrown errors, then we would still log those). I just realized two things though:

1) I opted to use a simpler statusCheck flow for retrying instead of using this for our retrying, so this flow is no longer used in situations where we want to keep context for the user.

2) If there is a non-network error thrown, we still want to display that in a notice.

So I'll push something up with a single catch at the end like:
```js
    .catch((error) => {
      console.error(error)
      const e = ErrorFactory.create(error)
      if (e.type === "NetworkError") return

      dispatch(Notice.set(ErrorFactory.create(e)))
    })
```

src/js/state/ConnectionStatuses/types.ts
```diff
@@ -2,7 +2,7 @@ export type ConnectionStatusesState = {
   [clusterId: string]: ConnectionStatus
 }
 
-export type ConnectionStatus = "connected" | "disconnected"
+export type ConnectionStatus = void | "connected" | "disconnected" | "retrying"
```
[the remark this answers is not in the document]
mason-fish: Oh sure thing, thanks for sharing that distinction 👍

src/js/components/NetworkErrorNotice.tsx
```diff
@@ -15,21 +16,39 @@ const MAX_BACKOFF = 128
...
+  const dismiss = () => {
+    dispatch(Notice.dismiss())
+  }
 
   useEffect(() => {
+    if (connId === connRef.current) {
+      statusRef.current = status
+      if (status === "disconnected") dismiss()
+    }
+  }, [status, connId])
```
mason-fish, line 38: this effect does two things. it updates the statusRef used when the component exits, and it will also force the component to exit if it sees that status has changed to "disconnected". It will only do those things if the connection matches the one set when the component was first created (and is referencing throughout it's lifecycle).

I had originally not paid any attention to the connId for the status monitoring, and ended up in situations where we if we switch the connection then this component would begin to operate against the wrong conn.

src/js/components/NetworkErrorNotice.tsx
```diff
@@ -15,21 +16,39 @@ const MAX_BACKOFF = 128
...
+
+  useEffect(() => {
+    connRef.current = connId
+    dispatch(ConnectionStatuses.set(connId, "retrying"))
```
mason-fish, line 42: on first render, the component sets the connId, which we will continue to refer to in the above effect so we are sure that we are only managing the connection in question. we then set it's status to retrying.

src/js/components/NetworkErrorNotice.tsx
```diff
@@ -39,7 +58,12 @@ export default function NetworkErrorNotice({error}: Props) {
     }
     tick()
 
-    return () => clearTimeout(id)
+    return () => {
+      clearTimeout(id)
+      if (statusRef.current === "retrying") {
```
mason-fish, line 63: when the component exits, which may happen because the user has disregarded the notice and clicked somewhere else in the app which triggers a Notice.dismiss(), we check if the status is still set to "retrying" and if it is then we set it to "disconnected".

The idea here is that since they can't do anything with the conn while it is down, we give the options of retrying and keeping context, or if they choose to navigate away then they are already changing context and we can fall back on the <ConnectionError /> at that point.

jameskerr:
I found another bug in here:

If there is a connection that has gone down, I can't switch to it using the connection picker.

vqeNJStZd9

In the dev tools

Screen Shot 2020-11-12 at 1 46 04 PM

Looks like something needs changing in the initCurrentTab function.

mason-fish:
@jameskerr thanks for catching that, it actually came from a change I had unintentionally made in the clusterPicker click handler. I returned it back to what it was before, and should be working correctly again

jameskerr (approved):
Let's do it. This works as expected when I tested it just now. I did find a bug unrelated to this that I may file an issue for:

When I switch connections. I get this behavior. Granted, I'm starting a zqd that points to the exact same data dir as the first zqd. That probably won't happen in practice.

image

Maybe when someone switches connections, we should drop them in with no spaces selected, I think. That would solve this.

mason-fish:
Sounds good, clear current space on connection switch 👍

PR #1205, opened 2020-11-11
Wait For Windows to Close

### LINKED ISSUE #1204: closing last window leaves lingering process, unable to start Brim afterwards
On my Windows 10 machine with the v0.19.0 Brim release, if I close the last open Brim window either by using the taskbar 'Close Window' option, or by closing the window via the upper-right "X", the window disappears, and Brim looks like it has quit. However, I then can't start Brim up again by clicking on its icon.

If I check Task Manager, I see that after quitting the last window, there are still some lingering 'Brim' processes left running, and I assume that means that the application did not in fact quit. If I manually quit those tasks in Task Manager, I can then launch Brim with the usual icon double click.

jameskerr:
fixes #1204 

When you close the last open Brim window, we automatically quit the app (unless you're on a Mac).

We are notified that the last window has closed using electron's event `app.on("window-all-closed")` [docs](https://www.electronjs.org/docs/api/app#event-window-all-closed).

When a window closes, it fires a "closed" event [docs](https://www.electronjs.org/docs/api/browser-window#event-closed). We use this event to dereference the window.

Strangely, electron fires these events in this order:

1. app => "window-all-closed"
2. browserWindow => "closed"

This led to calling a method on a browser window that was destroyed, but had not yet fired "closed" which would have caused us to dereference it.

To fix, I added a `windowManager.whenAllClosed()` method that will resolve when all windows have dereferenced.

### DIFF
--- a/src/js/electron/quitter.ts
+++ b/src/js/electron/quitter.ts
@@ -22,7 +22,12 @@ export function handleQuit(manager: $WindowManager, store, session, zqd) {
     }
   })
 
-  app.on("window-all-closed", () => {
-    if (process.platform !== "darwin") app.quit()
+  app.on("window-all-closed", async () => {
+    if (process.platform === "darwin") return
+    // Strangely, this event fires before the "closed" event on the window is fired,
+    // where we dereference the window. Here we check to make sure all the windows
+    // have indeed been dereferenced before we quit the app.
+    await manager.whenAllClosed()
+    app.quit()
   })
 }
--- a/src/js/electron/tron/windowManager.ts
+++ b/src/js/electron/tron/windowManager.ts
@@ -1,15 +1,15 @@
 import {BrowserWindow, screen} from "electron"
-
-import {NewTabSearchParams} from "../ipc/windows/messages"
-import {SessionState} from "./formatSessionState"
-import {WindowParams} from "./window"
+import {last} from "lodash"
 import brim from "../../brim"
 import ipc from "../ipc"
 import sendTo from "../ipc/sendTo"
-import tron from "./"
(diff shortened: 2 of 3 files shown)


src/js/electron/tron/windowManager.ts
```diff
@@ -67,7 +77,10 @@ export default function windowManager() {
         this.getWindows().map((w: BrimWindow) => w.confirmClose())
       )
         .then((oks) => oks.every((ok) => ok))
-        .catch(() => false)
+        .catch((e) => {
+          log.error(e)
```
jameskerr, line 81: This is where the error was silently being handled. We now log it, and say we're ok to quit. Better to allow the quit, than to hang.

src/js/electron/quitter.ts
```diff
@@ -22,7 +22,12 @@ export function handleQuit(manager: $WindowManager, store, session, zqd) {
     }
   })
 
-  app.on("window-all-closed", () => {
-    if (process.platform !== "darwin") app.quit()
+  app.on("window-all-closed", async () => {
+    if (process.platform === "darwin") return
+    // Strangely, this event fires before the "closed" event on the window is fired,
```
mason-fish, line 27: why?? so counterintuitive...glad you figured that out

src/js/electron/tron/windowManager.ts
```diff
@@ -54,6 +54,16 @@ export default function windowManager() {
       }
     },
 
+    whenAllClosed() {
```
mason-fish, line 57: dang, it's so strange that there is not an event fired that does this for us

mason-fish (approved):
Nice catch!

PR #1222, opened 2020-11-16
Cookbook for remote zqd

philrz:
For some time I've been looking to provide initial user-facing docs to encourage the use of remote `zqd` configurations. With the coming release of Brim `v0.20.0` and zq `v0.24.0`, the caveats are now minimal enough that I think the article attached to this PR is ready for consumption by the community. It's written as if these artifacts already exist. I can provide pointers to release candidate artifacts for anyone who wants to walk through the steps in the article themselves.


docs/Remote-zqd.md
```diff
@@ -0,0 +1,310 @@
...
+[Security Group](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_SecurityGroups.html)
+configuration that permits incoming port `9867` connections from our client's
+IP address.
+
+![Brim zapi and zqd](media/Security-Group.png)
+
+# Importing Data
+
+As mentioned in the [Limitations](#Limitations) above, it's not currently
+possible for remote Brim clients to import packet cpature data directly to a
```
jameskerr, line 201:
```suggestion
possible for remote Brim clients to import packet capture data directly to a
```

docs/Remote-zqd.md
```diff
@@ -0,0 +1,310 @@
...
+While new Spaces you create by importing logs via Brim will show up
+immediately in your app, additional Spaces created on the remote end via `zapi`
+or by other users importing to the same remote `zqd` will not appear
+automatically in your Brim app. To trigger an immediate refresh, select
+**View > Reload** from the Brim drop-down menu.
+
+![Reload to view new Spaces](media/Reload-New-Spaces.gif)
+
+A remote connection can be deleted by selecting the **Get Info** option in the
+pull-down and clicking **Remove**.
```
jameskerr, line 302: It might be good to use the word "removed" instead of "deleted" here. Or add some words saying that the connection is only removed from the app. The remote zqd is not shutdown nor the data directory deleted.

jameskerr (commented):
This is awesome Phil! Very clear to me.

docs/Remote-zqd.md
```diff
@@ -0,0 +1,310 @@
...
+
+3. The `-data` option points to the Data Directory, which is the default
+location for macOS in this case.
+
+4. The `-zeekrunner` option points to a script that is used to initiate the
+creation of Zeek logs from imported packet captures as described in the
+[Zeek Customization](https://github.com/brimsec/brim/wiki/Zeek-Customization)
+article.
+
+5. The `-brimfd=3` is an option unique to when `zqd` is launched by Brim.
```
mattnibs, line 115: I don't think we need to mention this.

mattnibs:
I would recommend using docker. Something like:

```
$ docker run -ti -p 9867:9867 ubuntu:latest /bin/bash
```

Then once inside the docker instance do as you say:
```
# wget --quiet https://github.com/brimsec/brim/releases/download/v0.20.0/brim_amd64.deb
# sudo apt update
# sudo apt install -y ./brim_amd64.deb
# mkdir -p /zqdroot
# /usr/lib/brim/resources/app/zdeps/zqd listen \
           -l :9867 \
           -data /zqdroot
           -zeekrunner /usr/lib/brim/resources/app/zdeps/zeek/zeekrunner
```

Really as a millennial that doesn't like to read, I think the document should begin with this. When I'm reading project docs I usually just want to get the quick answer. The deep dive that you do could go afterwards or perhaps in a separate document.

As an aside we should consider adding a zqd release variant that comes with the zeekrunner (and suricatarunner) prepackaged. It would be really cool to cut the steps for doing this down to:

```
# wget --quiet https://github.com/brimsec/brim/releases/download/v0.20.0/brim_amd64.deb
# sudo apt update
# sudo apt install -y ./brim_amd64.deb
# mkdir -p /zqdroot
# zqd -data /zqdroot
```

docs/Remote-zqd.md
philrz: Thanks @jameskerr. Great suggestion. I've pushed a change to clarify this.

docs/Remote-zqd.md
philrz: @mattnibs: I'd love to not have to mention it. The problem is that further down in the article I need to walk the user through creating a `zqd` command line that drops it. So once I show it being removed, it calls into question why it's there and why it had to be removed in order for the command line to work. That's why I explain what it is where it first appears.

I think you're picking up on how this kind of article is a balancing act. Many of us have been on-board for some time thinking users might benefit from starting to use remote `zqd` configs. However all along the way I've encountered bugs or usability issues that would make it difficult for all but the most experienced users. This is why I put the qualifying info in "About Cookbooks" saying the stuff is new/experimental and that exposing some of the innards is part of the point. If we had the dev/UX resources to address the things I still had to cite in "Limitations" and other usability improvements I didn't even mention, this could be more of a "For Dummies" thing where we'd not have to talk innards. Sadly, I don't think we're there quite yet, so I went this route in the interest of getting something out there rather than waiting on perfection.

philrz:
@mattnibs: I feel like a jerk because you were kind enough to provide feedback and now I'm pushing back in both the areas where you made suggestions. [😄 ]  However, there's good thoughts in here, so lemme run through it and maybe we can talk through what could be incorporated sooner/later.

> Really as a millennial that doesn't like to read, I think the document should begin with this. When I'm reading project docs I usually just want to get the quick answer.

I think this reveals that you and I see the purpose of this article quite differently. In my mind, users who want a quick answer and don't want to read are _not_ who I want following this doc. I wish I had that doc for them, but I don't think the tech is mature enough yet to produce such a doc.

What I tried to convey between the "About" and "Limitations" sections is that this functionality is potentially useful, but is really only in a state where we'd want users to try it if they're comfortable with the operational concepts & caveats that are revealed. If I just lead with a few steps that are easy to cut & paste, I fear it _won't_ "just work" due to one of the problems revealed later in the doc. At that point I fear one of two things might happen:

1. They'll come away with the impression that our tech is buggy or incomplete. Yes, we have bugs and more work left to do, so in a sense this is actually true. But I also don't want to hold back this functionality until it's perfect. Hence making them sit through disclosures.

2. They'll require Support-style hand-holding to get over the problems, and we don't have a lot of cycles for this right now. I'm certainly ready to support the users who hit snags, but I want to at least give them as much "self serve" up-front help as I can via the text.

If a user takes one look and thinks it looks too complicated for them to stomach, then I'd rather have them wait for us to deliver a SaaS service, since I expect it will be so turn-key that they don't even have to know what `zqd` is.

Maybe it would also help if I enumerate who I see as the target audience for the doc. Two come to mind:

1. The Zeek trainer that wants to offer students tools that can search logs. Some time back it was established that they'd be more comfortable including Brim if there were some central location to stage the data rather than having every user do their own imports.

2. Some of the MSSP users in the community we've talked to who currently run Brim on servers at remote customer locations. They currently Remote Desktop into such servers to run Brim on those desktops to execute queries. I felt they might like the option to maintain one local Brim app that connects to all those remotes without having to do the Remote Desktop.

I really want to enable both of these use cases and also see if other users come up with creative combos I didn't envision. But they're also going to need to know what they're signing up for and be able to debug it if there's operational problems, hence all the disclosures.

> I would recommend using docker.

This is definitely a good idea. My preference would be to deliver this via a separate document. There were similar things that came to mind that I also intentionally chose not to get into since I didn't want to boil the ocean. For instance, as soon as we describe `zqd` as this process you might always want running remotely, that makes me think "Well, I want it to always come up whenever I reboot this server, then." So that would bring us onto the topic of startup wrapper scripts, monitoring processes, which Linux distros use which approach, do we show multiple ways, etc. I'm hoping that most "ops" readers will have their favorite ways to achieve this. Therefore if we just introduce `zqd` as a fairly simple daemon process with minimal port/storage requirements, that "ops" reader will know how to fit it into their standard toolkit. Docker might very well be part of that.

All that said, you're correct that Docker is such a standard part of ops nowadays that it probably deserves coverage. I just would be nervous of making it the happy path: "You must be this container to ride this ride." The numbers are getting smaller every day, but (I won't name names 😄 ) one of our own employees recently expressed instinctive resistance to making one of our build processes dependent on Docker, and I can sympathize as I was not so different in the recent past. Given the diversity of our user community, I don't want to make too many assumptions about the skills they show up with.

In conclusion, I'd love to see some kind of doc with best practices for running our stuff in Docker, and then we could link to it from one of those "Variation" blocks like I already did with the tip about running regular `zqd` rather than the Brim package. @marktwallace already wrote his README docs that take this concept all the way to running distributed stuff in K8S, and that has Docker stuff as a prerequisite. I sense we could still benefit from a stepping stone article for an audience that maybe isn't looking at containerization to achieve big scale but instead just prefers the operational advantages as you cite.

> As an aside we should consider adding a zqd release variant that comes with the zeekrunner (and suricatarunner) prepackaged.

Would you also include the full Zeek and Suricata artifacts as well? This reminds me of some packaging discussions we had early on after starting to open source the tech. It seems like the answers can be quite different depending on the target audience.

It seems like we could learn a lot here by looking at Linux distros. For instance, you can get super big "desktop" distros that are more bulky but have all the GUI out-of-the-box and hence are well-suited for users that are looking for a turn-key alternative to running Windows or macOS on the desktop. If you handed such a user a "minimal" distro that had no GUI in it and they had to learn how to use `apt` or `yum` to install the correct additional packages a la carte, they might say "Why the heck would anyone want such a dumb distro?" Meanwhile, some other "ops" user might only ever be installing Linux on remote servers with no monitors on them, so they're only ever going to want the "minimal" distro, and if they only ever offered a giant desktop one instead with all its bulk and excess packages, they might say "Why the hack would anyone want such a dumb distro?" Hence the different desktop/minimal distros and the ability to add packages if you start in one place and then change your mind.

Much like the Linux distros, perhaps we could benefit from having packaging that makes it easy to start in one spot and easily morph to the other. For instance, make it so I can run Brim on my desktop as usual, but invoke a Preferences option to allow it to open up its `zqd` to remote connections if I'm sitting next to someone at a Capture The Flag and I want them to be able to browse my logs. Or like you imply, make it easy for someone to install the zq tooling as a minimal set or include as much of Zeek/Suricata runners or the full artifacts as they need to suit their use cases.

mattnibs:
> The numbers are getting smaller every day, but (I won't name names 😄 ) one of our own employees recently expressed instinctive resistance to making one of our build processes dependent on Docker

I can definitely sympathize with this sentiment. 

> Would you also include the full Zeek and Suricata artifacts as well?

That's what I was thinking. As you said, a minimal zqd only distribution and a large distribution with Zeek and suricata. 

OTH would be nice if zqd could just use the Zeek and suricata binaries already installed in the system- our package could even specify them as dependency. This approach might fit better with the dependency approach that package managers take. Of course we'd have to some adjustments- we currently have runners that wrap around the binaries we'd have to figure this one out. Plenty to discuss.

alfred-landrum (approved):

PR #1226, opened 2020-11-18
remove connection

### LINKED ISSUE #1191: Enable connection removal/deletion
The last missing operation on connections/clusters is to allow users to remove them.

This should include a sort of on-cascade handling of empty states for any existing search tabs or windows which are referring to the connection by notifying the user of this change, similar to how we handle space deletion.

For now, we should also prohibit the user from removing connections to `localhost:9867` since the app depends on always having it's own local zqd instance running

mason-fish:
fixes #1191 

Allows users to remove connections (not the default though). Includes a new empty state, when users have removed a connection on any existing tabs. I also found a few little bugs along the way which I'll annotate in the corresponding code

Screen Shot 2020-11-18 at 1 38 04 PM
Screen Shot 2020-11-18 at 1 38 15 PM
Screen Shot 2020-11-18 at 1 38 23 PM


src/js/components/ConnectionChooserPage.tsx
```diff
@@ -0,0 +1,99 @@
...
+  width: 100%;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  overflow: scroll;
+`
+
+const StyledHeader = styled.h1`
+  margin: 110px 0 0 0;
```
jameskerr, line 70:
```suggestion
  margin: 96px 0 0 0;
```

src/js/components/ConnectionModals/ViewConnectionModal.tsx
```diff
@@ -25,6 +30,11 @@ const StyledFooter = styled.footer`
   button {
     margin-left: 12px;
   }
+
+  a {
+    flex: 1;
+    color: var(--havelock);
```
jameskerr, line 36:
```suggestion
    margin-right: auto;
    color: var(--red);
```

flex: 1 will make all the space to the right of the "Remove" text also clickable. margin-right: auto achieves the same positioning without making that region part of the link.


I think red looks good here:

image

src/js/components/LeftPane.tsx
```diff
@@ -124,6 +125,21 @@ const DragAnchor = styled.div`
   cursor: row-resize;
 `
 
+const EmptyText = styled.div`
+  ${(p) => p.theme.typography.labelNormal}
+  color: var(--slate);
+  margin: 191px 37px 0;
```
jameskerr, line 131:
```suggestion
  margin-top: 110px;
  padding: 0 24px;
```

src/js/components/ConnectionModals/ConnectionForm.tsx
```diff
@@ -127,7 +126,8 @@ const ConnectionForm = ({onClose, conn}: Props) => {
         return obj
       }, {})
       try {
-        await dispatch(initConnection(toCluster({host, name})))
+        const id = conn && conn.id
+        await dispatch(initConnection(toCluster({id, host, name}), true))
```
jameskerr, line 130: Was this where the bug you were referring to was?

src/js/components/LeftPane.tsx
```diff
@@ -124,6 +125,21 @@ const DragAnchor = styled.div`
...
+  text-align: center;
+`
+
+const StyledPane = styled(Pane)<{isEmpty: boolean}>`
+  ${(p) =>
+    p.isEmpty &&
+    css`
+      background: var(--snow);
```
jameskerr, line 139: This pane should always have a background of snow. You mind going over into _history-pane.scss and changing the background from $white to var(--snow)? Then you could remove this bit.

src/js/components/TabContent.tsx
```diff
@@ -24,13 +26,16 @@ export default function TabContent() {
   const dispatch = useDispatch()
   const space = useSelector(Current.getSpace)
   const conn = useSelector(Current.getConnection)
-  const connStatus = useSelector(ConnectionStatuses.get(conn.id))
+  const id = get(conn, ["id"], "")
+  const connStatus = useSelector(ConnectionStatuses.get(id))
```
jameskerr, line 30: This is just an idea for the future, not a suggestion for this PR. We could make Current.getConnection also make the join with the ConnectionStatuses and add that to the object that it returns. Then we could just write conn.status.

src/js/flows/getZealot.ts
```diff
@@ -31,7 +31,7 @@ export const getZealot = (): Thunk<Zealot> => (
   getState,
   {createZealot}
 ) => {
-  const conn = Current.mustGetConnection(getState())
+  const conn = Current.getConnection(getState())
```
jameskerr, line 34: This should stay mustGetConnection I think? If not, we should check for the presence of conn.

src/js/flows/initConnection.ts
```diff
@@ -4,13 +4,14 @@ import {Cluster} from "../state/Clusters/types"
 import refreshSpaceNames from "./refreshSpaceNames"
 import {globalDispatch} from "../state/GlobalContext"
 import ConnectionStatuses from "../state/ConnectionStatuses"
+import brim from "../brim"
 
-export const initConnection = (cluster: Cluster) => (
+export const initConnection = (cluster: Cluster, isNew = false) => (
```
jameskerr, line 9: Instead of an additional conditional param, could we check the state for this cluster's id? Or check for this id in the ConnectionStatuses?

src/js/flows/removeConnection.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const removeConnection = (conn: Cluster): Thunk => (
+  dispatch,
+  _getState,
+  {globalDispatch}
+) => {
+  const {name, id, host, port} = conn
+
+  if (host === "localhost" && port === "9867")
+    throw new Error("Cannot remove the default connection")
```
jameskerr, line 18: I like having this check here. Let's also remove the remove (haha) button from the page if we're looking at the details for localhost:9867.

src/js/flows/removeConnection.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  dispatch(Current.setSpaceId(null))
+  dispatch(Current.setConnectionId(null))
+  dispatch(Investigation.clearConnectionInvestigation(id))
+  dispatch(Spaces.removeForConnection(id))
+  dispatch(ConnectionStatuses.remove(id))
+  globalDispatch(Clusters.remove(id))
+  popNotice(`Removed connection "${name}"`)
```
jameskerr, line 26: Every time I see this function I think this:

tenor

jameskerr (changes requested):
Good stuff Mason! I left a few suggestions.

src/js/components/ConnectionModals/ConnectionForm.tsx
mason-fish: Ah shoot, sorry I never saved my inline comments. Yes this is one of them. The problem here is that since I've started using generated id's for non-default connections, whenever we try to create a new connection and it fails we end up polluting connectionStatuses. As I say it out loud it occurs to me that we won't be persisting that state anyway though, so perhaps I will just remove this and let the state clean itself up on app reload

src/js/flows/removeConnection.ts
mason-fish: 😆

src/js/flows/getZealot.ts
mason-fish: Ah yes, thanks. I had changed this while debugging but forgot to switch back

src/js/components/ConnectionModals/ConnectionForm.tsx
```diff
@@ -95,15 +95,14 @@ const ConnectionForm = ({onClose, conn}: Props) => {
...
+  const toCluster = ({id, host, name}): Cluster => {
     // set defaults
     let [h, p] = host.split(":")
     if (!p) p = "9867"
-    const hostPort = `${h}:${p}`
     return {
       host: h,
       port: p,
-      id: hostPort,
+      id: id || brim.randomHash(),
```
mason-fish, line 105: using the `host:port` as the id here, as well as not using an existing id if it already exists was causing a handful of weird bugs where either ids were changing, or you might inadvertently edit a different connection if it's id matched the host:port of the current one you are creating or editing. unique ids make more sense for this anyway.

one note: to help distinguish the "default" connection/workspace we will keep the id for that as "localhost:9867", which both helps us pluck it out of the clusters object more readily and also leaving it like that saves us a migration

src/js/initializers/initConnectionParams.ts
```diff
@@ -20,14 +22,12 @@ const setupConnection = (host, port) => (dispatch, _, {globalDispatch}) => {
...
 
 export default function(store: Store) {
-  const {host, port, id} = getUrlSearchParams()
+  const {id} = getUrlSearchParams()
   global.windowId = id
 
   const existingConnection = Current.getConnection(store.getState())
 
-  if (host && port) {
-    store.dispatch(setupConnection(host, port))
```
mason-fish, line 29: here was another bug: searching from a detail window back to a search window (via context menu) was  resetting the connection's name back to it's default (host:port). Using these host and port params isn't necessary anymore though

jameskerr (approved):
Code looks good! And works as expected with my testing. It would be good to get some integration tests on this new behavior. But I won't stop this PR for that. :shipit:

PR #1228, opened 2020-11-18
Log endpoint uses multipart form

### LINKED ISSUE #1094: Update Log Post API to Multi-Part File Upload
As @mattnibs described in Slack:

The updated endpoint now expects a request in the following format. It will also be the desktop app's job to determine how to fill the progress bar, instead of relying on the response stream.

The zq branch to develop this with is in the PR: https://github.com/brimsec/zq/pull/1336

```
path: /space/:spaceid/logs
method: POST
query params: 
- stop_error <bool>: This instructs the server to abort the request if it receives a parsing error, or is unable to identify the content-type of a file. For brim I would recommend not using this param; the server will report these errors as warnings in the response body.
form data contents:
- name: "json_config"
  filename: <can be left blank>
  contents:  This should be a `json` payload of the typing configs, used to inform the zqd server how to richly encode data from an ndjson log file. This should be included as the first item in the FormData array.
- name: <can be left blank>
  filename: <name of the log file posted>
  contents: This should be the raw contents of a selected log file. You can post as many of these as you like.
response:
- content-type: application/json
- payload:
    - type <string>: Payload type. Will always be "LogPostResponse"
    - bytes_read <number>: The amount of bytes read from the form. The body of `json_type_config` will also be included in the count, so this might not match up the count of log data sent.
    - warnings <[]string>: A list of warnings encountered during log ingest.
```

jameskerr:
UPDATE:

I've added a new method to the fetcher called "upload". It uses XHR to make the request and track progress events. We still return an iterable stream of payloads which now include a message that looks like `{type: "UploadProgress", progress: number}`. This allows us to keep the progress bar unchanged as it was.

Fixes #1094 

This changes the log post method in the zealot client to make a multipart form file upload, instead of sending the paths of the files. The pcaps post endpoint has remained unchanged.

It is not easy to get a progress bar for this type of request (it's impossible ATM using the Fetch API, we'd have to rewrite that method with XHR). I think we didn't have progress for log uploads anyway.

@philrz when you verify this, upload a large log file (or set of log files) over 4 GB. I expect it to work but that would be good to check.

P.S.
@mattnibs I noticed that the key for the warnings was a "nil". That may be a bug, see below:

image

### DIFF
--- a/src/js/brim/ingest/fileList.ts
+++ b/src/js/brim/ingest/fileList.ts
@@ -1,46 +1,52 @@
 import {IngestFileType} from "./detectFileType"
 import lib from "../../lib"
 
-export type FileListData = {type: IngestFileType; path: string}[]
+export type FileListData = {type: IngestFileType; file: File}[]
 
-export default function fileList(files: FileListData) {
+export default function fileList(list: FileListData) {
   return {
     first() {
-      return files[0]
+      return list[0]
     },
 
     oneFile() {
-      return files.length === 1
+      return list.length === 1
     },
 
     multiple() {
-      return files.length > 1
+      return list.length > 1
     },
 
     paths(): string[] {
-      return files.map((f) => f.path)
+      return list.map((f) => f.file.path)
+    },
+
+    files(): File[] {
+      return list.map((f) => f.file)
     },
 
     any(type: string) {
-      return !!files.find((f) => f.type === type)
+      return !!list.find((f) => f.type === type)
     },
 
     allPcap() {
-      return files.every((f) => f.type === "pcap")
+      return list.every((f) => f.type === "pcap")
     },
 
     mixed() {
-      return !files.every((f) => f.type === files[0].type)
+      return !list.every((f) => f.type === list[0].type)
     },
 
     inSameDir() {
-      return files.every(
-        (f) => lib.file(f.path).dirName() === lib.file(files[0].path).dirName()
+      return list.every(
+        (item) =>
+          lib.file(item.file.path).dirName() ===
+          lib.file(list[0].file.path).dirName()
       )
     },
 
     dirName() {
-      return lib.file(files[0].path).dirName()
+      return lib.file(list[0].file.path).dirName()
     }
   }
 }
--- a/src/js/flows/ingestFiles.ts
+++ b/src/js/flows/ingestFiles.ts
@@ -158,12 +158,6 @@ const trackProgress = (client, gDispatch, clusterId) => {
         else return status.pcap_read_size / status.pcap_total_size || 0
       }
 
-      function logPostStatusToPercent(status): number {
-        // log_total_size may not be present
-        if (!status.log_total_size) return 1
-        else return status.log_read_size / status.log_total_size
-      }
-
       gDispatch(space.setIngestProgress(0))
       for await (const {type, ...status} of stream) {
         switch (type) {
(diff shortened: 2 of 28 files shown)


alfred-landrum:
@jameskerr : I know we can't get the same status messages that were used to implement the existing progress bar. I thought we would watch how much data has been uploaded to the backend api, and use that as the basis for a progress bar in this cae? That's how the `zapi` command is giving a progressive status for the new api.

jameskerr:
I understand. It's just that the zealot client is setup to use the modern, promise-based "Fetch" web api for making requests. The Fetch API has not created a way to hook into the upload progress. We'd have to use the older XMLHttpRequest syntax for just this one api call. Definitely possible, but not quite plug and play.

https://usefulangle.com/post/321/javascript-fetch-upload-progress

philrz:
Just to chime in with a visual, here's proof of the progress bar we've previously had for log import:

ProgressBar

Regarding large imports, I'll do a final verification when this merges to master, but I did just do a quick test with the "wrccdc-year1" data set that's 12 GB of uncompressed Zeek TSV logs and they did import without complaint.

jameskerr:
@philrz Thanks for your help! Glad to hear it worked with a big one. So that progress bar is there, but they are unable to search for anything until it completes. That's right.

src/js/flows/ingestFiles.ts
```diff
@@ -169,6 +171,13 @@ const trackProgress = (client, gDispatch, clusterId) => {
             gDispatch(space.setIngestProgress(logPostStatusToPercent(status)))
             updateSpaceDetails()
             break
+          case "LogPostResponse":
+            console.log(status)
```
mason-fish, line 175: left a straggler

src/js/flows/ingestFiles.ts
```diff
@@ -169,6 +171,13 @@ const trackProgress = (client, gDispatch, clusterId) => {
...
+            updateSpaceDetails()
+            ;(status.warnings || []).forEach((warning) => {
```
mason-fish, line 177: I think this semi-colon isn't doing anything. Also, would recommend using lodash's `forEach` for this instead

src/js/brim/ingest/fileList.ts
```diff
@@ -1,46 +1,52 @@
 import {IngestFileType} from "./detectFileType"
 import lib from "../../lib"
 
-export type FileListData = {type: IngestFileType; path: string}[]
+export type FileListData = {type: IngestFileType; file: File}[]
```
mason-fish, line 4: not sure if you considered extending the `File` interface by annotating it with the new `type` field? would save all the extra `.file.`s since it would be flattened, just a thought!

mason-fish (approved):
Looks good! I had thought we could keep track of file upload as well, thanks for pointing out that constraint. Hopefully they do add it in to the fetch api soon.

My comments are mostly superficial, but pls take a look before merging.

src/js/flows/ingestFiles.ts
jameskerr: Thanks!

src/js/flows/ingestFiles.ts
jameskerr: Yeah, prettier added that in. I guess when a line starts with parens, it adds the semi-colon.

src/js/brim/ingest/fileList.ts
jameskerr: Yeah true, I usually try to stay away from extending things like that. "Compose rather than inherit" mindset I guess. Is this something that is done conventionally in go?

mason-fish: I think the composition vs inheritance dilemma is the same in go. I do agree with the compose > inheritance mindset, but thought that in this case with just these two types, and one always depending on the other, extending would grant the small bonus of less typing as you access the imbedded File interface. I have no strong feelings in this case at all though!

zealot/src/fetcher/fetcher.ts
```diff
@@ -21,11 +23,42 @@ export function createFetcher(host: string) {
...
       const resp = await fetch(url(host, path), {method, body, signal})
       const iterator = createIterator(resp, args)
       return createStream(iterator, resp)
+    },
+    async upload(args: FetchArgs): Promise<ZReponse> {
+      return new Promise((resolve) => {
+        const iterator = createPushableIterator<ZealotPayload>()
+        const xhr = new XMLHttpRequest()
+
+        xhr.upload.addEventListener("progress", (e) => {
```
mason-fish, line 37: 🔥

PR #1235, opened 2020-11-21
Precise error message zqd pcap not found error

### LINKED ISSUE #1221: Improve error message for failed attempt to import pcap to remote zqd
We'd like to start encouraging community users to work with remote `zqd` configurations. However, the majority of users have grown accustomed to importing pcaps into Brim as part of their workflows, and therefore are likely to attempt this in their remote configurations as well. In the user docs I'm creating, I'm making all attempts to educate the user that this is not currently supported, but I would not be surprised if users don't fully absorb the docs and try it anyway (it's already happened once that I'm aware of). Therefore the error messages should be as helpful as possible.

At the moment (Brim commit `b77257b` talking to `zqd` commit `acf0718b`), if I attempt this, the error message I see isn't super helpful:

image

It seems like this might be down to how the app presents the error, since if I sniff the response coming from `zqd`, there's a bit more info in there than is bubbling up:

Request:

```
POST /space/sp_1kODRYbcJRSboGmag9Sq67o38hx/pcap HTTP/1.1
Host: 18.219.222.199:9867
Connection: keep-alive
Content-Length: 40
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_16_0) AppleWebKit/537.36 (KHTML, like Gecko) Brim/0.19.0 Chrome/85.0.4183.121 Electron/10.1.4 Safari/537.36
Content-Type: text/plain;charset=UTF-8
Accept: */*
Accept-Encoding: gzip, deflate
Accept-Language: en-US

{"path":"/Users/phil/pcap/hello.pcapng"}
```

Response:

```
HTTP/1.1 404 Not Found
X-Request-Id: 21
Date: Mon, 16 Nov 2020 21:38:57 GMT
Content-Length: 93

{"type":"Error","kind":"item does not exist","error":"file:///Users/phil/pcap/hello.pcapng"}
```

My initial proposal for how to improve would be to have some reference to the name of the connection. Something like:

> File `/Users/phil/pcap/hello.pcapng` does not exist on MyRemote

mattnibs:
If a pcap file is posted to a remote server the remote server will return 404
not found. The current error message was just name of the file. Update the
error message in this case to reflect that the server does not have to access
the file.

Related change:

- The zealot fetcher was not throwing an error if a stream request response was
a 404. Change so it does if the response is not ok.

Closes #1221

### DIFF
--- a/zealot/src/fetcher/fetcher.ts
+++ b/zealot/src/fetcher/fetcher.ts
@@ -24,6 +24,10 @@ export function createFetcher(host: string) {
     async stream(args: FetchArgs) {
       const {path, method, body, signal} = args
       const resp = await fetch(url(host, path), {method, body, signal})
+      if (!resp.ok) {
+        const content = await parseContentType(resp)
+        return Promise.reject(createError(content))
+      }
       const iterator = createIterator(resp, args)
       return createStream(iterator, resp)
     }
--- a/zealot/src/zealot_mock.ts
+++ b/zealot/src/zealot_mock.ts
@@ -4,6 +4,7 @@ import {createStream} from "./fetcher/stream"
 import * as zqd from "./zqd"
 import {zjson} from "./index"
 import {Zealot} from "./types"
+import {createError} from "./util/error"
 import {zngToZeek} from "./enhancers/mod"
 
 type StubMode = "always" | "once"
@@ -42,6 +43,7 @@ export interface ZealotMock {
     mode?: StubMode
   ) => ZealotMock
   stubPromise: (method: string, output: any, mode?: StubMode) => ZealotMock
+  stubError: (method: string, err: any, mode?: StubMode) => ZealotMock
   calls: (method: string) => {method: string; args: any}[]
   zealot: Zealot
 }
@@ -92,6 +94,10 @@ export function createZealotMock(): ZealotMock {
     },
     stubPromise(method: string, output: any, mode: StubMode = "once") {
       stub(method, output, promise, mode)
+      return this
+    },
+    stubError(method: string, err: any, mode: StubMode = "once") {
+      override(method, () => { throw createError(err) })
       return this
     },
     calls(method: string) {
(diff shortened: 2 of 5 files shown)


philrz:
Thanks @mattnibs. Definitely an improvement.

image

zealot/src/fetcher/fetcher.ts
```diff
@@ -24,6 +24,10 @@ export function createFetcher(host: string) {
     async stream(args: FetchArgs) {
       const {path, method, body, signal} = args
       const resp = await fetch(url(host, path), {method, body, signal})
+      if (!resp.ok) {
+        const content = await parseContentType(resp)
+        return Promise.reject(createError(content))
+      }
```
jameskerr, line 30: This check existed in the next function `createIterator` (https://github.com/brimsec/brim/blob/master/zealot/src/fetcher/iterator.ts#L12), but it does feel like it belongs where you put it here. I like it. Would you remove it from the `createIterator` function and copy the `isObject(error)` check?

zealot/src/zealot_mock.ts
```diff
@@ -94,6 +96,10 @@ export function createZealotMock(): ZealotMock {
       stub(method, output, promise, mode)
       return this
     },
+    stubError(method: string, err: any, mode: StubMode = "once") {
```
jameskerr, line 99: Nice!

jameskerr (changes requested):
Thanks Matt! I have one requested change, but looks good otherwise.

zealot/src/fetcher/fetcher.ts
mattnibs: Sure

mattnibs: Looks like the `createError` method already does the check so I don't think we need to include it.

jameskerr (approved):
Great! Using createError in that spot seems like a good call. Thanks

mattnibs:
> Great! Using createError in that spot seems like a good call. Thanks

Thanks @jameskerr

Issue #1199, opened 2020-11-10
CRUD ui for query lib


mason-fish:
Implement current ui designs for the query lib once the model, reducer, and generic list component has been completed. This generally should bring all of the above together, and will include the full CRUD flow for adding, viewing, adjusting, and removing queries.

Issue #1201, opened 2020-11-10
Populate standard/default query lib


mason-fish:
Once we have finished building the query lib, we intend to include some standard queries that most people would find useful. Let's build that starter lib!

philrz:
As we discussed in a recent group meeting, let's make sure one of the out-of-the-box queries highlights the Suricata events. This would give us a place to highlight the new data even if we don't yet have other fancier visualizations specific to Suricata data. We could use community feedback to adjust this or add alternatives if we don't get it right the first time.

My proposal would be:

```
event_type=alert | count() by alert.severity,alert.category | sort count
```

With the example of the our test data `all.pcap`, this gives:

![image](https://user-images.githubusercontent.com/5934157/100921024-d77a8d00-3490-11eb-89ca-86069eeafd4a.png)

orochford:
Name: Activity Overview
ZQL: count() by _path | sort -r
Description: This query shows a list of all Zeek streams  in the data set, with a count of associated records
Tags: zeek, initial_exploration

orochford:
Name: Unique DNS Queries
ZQL: _path=dns | count() by query | sort -r
Description: Shows all unique DNS queries contained in the data set with count
Tags: dns, initial_exploration

orochford:
Name: Windows Networking Activity
ZQL: _path=~smb* OR _path=dce_rpc
Description: Filters and displays smb_files, smb_mapping and DCE_RPC activity
Tags: windows, smb, malware

orochford:
Name: HTTP Requests
ZQL: _path=http | cut id.orig_h, id.resp_h, id.resp_p, method,host,  uri | uniq -c 
Description: Displays a list of the count unique HTTP requests including source and destination
Tags: http, initial_exploration, malware

orochford:
Name: Unique Network Connections
ZQL: _path=conn | cut id.orig_h, id.resp_p, id.resp_h | sort | uniq
Description: Displays a table showing all unique source:port:destination connections pairings
Tags: network, initial_exploration

orochford:
Name:
ZQL: _path=conn | put total_bytes = orig_bytes + resp_bytes | sort -r total_bytes | cut uid, id, orig_bytes, resp_bytes, total_bytes
Description: Shows the connections between hosts, sorted by data received
Tags: network

orochford:
Name: File Activity
[...]

PR #1239, opened 2020-11-24
QueryLibrary model and reducer

### LINKED ISSUE #1198: Query library data model
Design a data model which can accommodate the query library constraints we intend to implement initially, while also being flexible enough to adapt to some of the future features we have discussed.

**Initial requirements**
- [ ] model can easily serialize into (or simply is) `json` so it may be easily transported/shared.
- [ ] metadata
  * tags: many to many cardinality for tags :: queries
  * name: possibly limited in length
  * description
  * value: actual zql to be executed
- [ ] parametric search (description, name, value), though perhaps just by name initially? (cc @alfred-landrum)
- [ ] tag filtering
- [ ] ordered (and reorder-able) while not filtered

**Future features**
- [ ] query templates, or parametric queries: allow for parameters which are meant to be filled in by the user in addition to selecting the query itself
- [ ] grouping: directory style, arbitrary, user-defined/named grouping of queries
- [ ] versioning: for synchronizing with remote sources

Finally, this ticket should also include writing the corresponding reducer for the model that we come up with

mason-fish:
fixes #1198 

Adds a new QueryLibrary reducer. This code is all additive and is not yet being used, though it will be once we link together the remaining ui work tracked in #1199 

The underlying data structures we are using to represent this lib are simply an object tree and arrays. The arrays provide the order which we need while also keep the structure easily serializable into JSON, and the tree structure provides the grouping hierarchy which we will want to use later on. The issue/inefficiency of re copying elements on array resize/reorder/remove is moot since we are using redux, which depends on data being immutable and when changes occur they must be done by creating new resources anyway.


src/js/state/QueryLibrary/selectors.ts
```diff
@@ -0,0 +1,6 @@
+import {QueryLibraryState} from "./types"
+import {State} from "../types"
+
+export default {
+  getRaw: (state: State): QueryLibraryState => state.queryLibrary
+}
```
mason-fish, line 6: We may want more selectors here, and will likely return the library as a parsed tree model in one of them. For now I am only returning the raw state and the tests are all built off of that, and as I am wiring this up to the components that consume this data then I may add more

src/js/state/QueryLibrary/initial.ts
```diff
@@ -0,0 +1,9 @@
+import {QueryLibraryState} from "./types"
+
+const init = (): QueryLibraryState => ({
+  id: "root",
+  name: "root",
+  items: []
```
mason-fish, line 6: this will actually be populated by #1201

src/js/state/QueryLibrary/reducer.ts
```diff
@@ -0,0 +1,91 @@
...
+  const srcItem = get(draft, toItemPath(srcItemPath), null)
+
+  if (!srcItem) return
+  if (!get(draft, toItemPath(initial(destItemPath)), null)) return
+
+  // If the move is all in the same directory then the adjusting indices can
+  // cause an off by one issue since the destination index will be affected after
+  // removal (e.g. an item cannot be moved to the end of its current group because of this).
+  // For this situation we instead remove the item first, and then insert its copy
+  if (srcItemPath.length === destItemPath.length) {
```
jameskerr, line 74: Is this just comparing the depths? Should check to see if the two paths (minus the last item) are equal?

src/js/state/QueryLibrary/types.ts
```diff
@@ -0,0 +1,50 @@
+export type QueryLibraryState = Group
+
+export interface Query {
+  id: string
+  name: string
+  value: string
```
jameskerr, line 6: What do you think about naming this field `zql` instead of `value`?

jameskerr (approved):
Mason, this is a great foundation. I'm glad we have those thorough tests as well. I have 2 questions for ya below.

src/js/state/QueryLibrary/types.ts
mason-fish: Sure, that works

src/js/state/QueryLibrary/reducer.ts
mason-fish: Oh thank you, yes this is not enough as is. will fix

mason-fish: actually, this does work as is: removing before or after only matters if it IS the same group. But I still prefer the handling to be only during the special case so am adding an isEqual check in the next commit

src/js/state/QueryLibrary/reducer.ts
```diff
@@ -71,7 +71,7 @@ const moveItem = (
   // cause an off by one issue since the destination index will be affected after
   // removal (e.g. an item cannot be moved to the end of its current group because of this).
   // For this situation we instead remove the item first, and then insert its copy
-  if (srcItemPath.length === destItemPath.length) {
+  if (isEqual(initial(srcItemPath), initial(destItemPath))) {
```
jameskerr, line 74: I didn't know about that "initial" function. Very cool

jameskerr (commented):
I was thinking about this this week and wondered if the QLIB prefix was a bit too jargony. I think a prefix like QUERIES would be more obvious for a new user.

```
QUERIES_ADD_ITEM
QUERIES_REMOVE_ITEM
QUERIES_MOVE_ITEM
etc..
```

mason-fish:
@jameskerr I agree qlib is a very insider term. I'm only using it in code for var names and for the action strings, and never show it to users. I shortened it to save some typing and I think it's mostly used within folders that have the name expanded so the context for future devs should be clear, but if you feel strongly I can change

jameskerr:
Thanks for your comment Mason. You know, it appears I do feel strongly enough to request the change to be QUERIES.  I think the top-level folder in the state should also match and just be "queries".  Thanks!

jameskerr:
Thanks for making that name change for me!

PR #1241, opened 2020-11-25
Handle unset records in zealot

### LINKED ISSUE #1174: Crash due to "null" value for a record
Repro is with Brim commit `3ffee71` talking to `zqd` commit `481c6f7`.

The [ZNG spec](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md) states:

> Any value in ZNG can take on a null representation.

We start with the following TZNG, that includes a `null` value for a nested `record`:

```
$ cat nullrec.tzng 
#0:record[_path:string,ts:time,foo:record[one:float64]]
0:[crashy;0.0;-;]
```

When imported into Brim, we see the following crash:

image

To the degree that Brim is currently rigged up primarily to handle Zeek data, indeed, this data is a bit unusual. Since Zeek always flattens records before outputting them, we'd never see that "top-level" `null` on data that truly came from Zeek. For instance, if I make that input file take a round trip through `zq` to be reformatted into Zeek TSV before being turned back into TZNG, we see:

```
$ zq -f zeek nullrec.tzng | zq -t -
#0:record[_path:string,ts:time,foo:record[one:float64]]
0:[crashy;0;[-;]]
```

_That_ imports into Brim without complaint. However, I stumbled onto this other example while going through variations related to "generic ingest", so I'm trying to get a handle on where we're currently weak when it comes to non-Zeek data.

nwt:
Fixes #1174.

### DIFF
--- a/zealot/src/zng/types/record.ts
+++ b/zealot/src/zng/types/record.ts
@@ -25,7 +25,7 @@ export class Record implements ZngClass<Type[] | null> {
     return new Record(type.of, value)
   }
 
-  isSet(): this is {value: zjson.Value[]} {
+  isSet() {
     return this.value !== null
   }
 
@@ -53,7 +53,7 @@ export class Record implements ZngClass<Type[] | null> {
   at(index: number) {
     const col = this.type[index]
     if (!col) throw new Error(`No column at index: ${index}`)
-    const val = this.isSet() ? this.value[index] : null
+    const val = this.isSet() ? this.value![index] : null
     const type = "of" in col ? col : col.type
     return constructType(type, val)
   }
@@ -79,7 +79,7 @@ export class Record implements ZngClass<Type[] | null> {
     if (!col) {
       throw new UnknownColumnError(name, this.getColumnNames())
     } else {
-      const val = this.value[this.type.indexOf(col)]
+      const val = this.value![this.type.indexOf(col)]
       const type = "of" in col ? col : col.type
       return constructType(type, val)
     }
@@ -121,7 +121,7 @@ export class Record implements ZngClass<Type[] | null> {
       } else {
         cols.push({...column, name: prefix + column.name})
         // For an unset record, supply an unset value for each column.
-        vals.push(this.isSet() ? this.value[index] : null)
+        vals.push(this.isSet() ? this.value![index] : null)
       }
     })
     return new Record(cols, vals)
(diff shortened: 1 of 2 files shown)


jameskerr (approved):
Thanks for the tests and the changes @nwt ! I have one suggestion for ya.

zealot/src/zng/types/record.ts
```diff
@@ -49,7 +53,7 @@ export class Record implements ZngClass<Type[] | null> {
   at(index: number) {
     const col = this.type[index]
     if (!col) throw new Error(`No column at index: ${index}`)
-    const val = this.value[index]
+    const val = this.value === null ? null : this.value[index]
```
jameskerr, line 56: What do you think about calling `this.isSet()` and `!this.isSet()` in all the places you compare `this.value === null`. I think it might read a bit better.

nwt: That would definitely read better.

zealot/src/zng/types/record.ts
```diff
@@ -12,25 +12,29 @@ class UnknownColumnError extends Error {
...
+  constructor(
+    readonly type: zjson.Column[],
+    readonly value: zjson.Value[] | null
+  ) {}
 
   static deserialize({type, value}: SerializedRecord): Record {
     return new Record(type.of, value)
   }
 
   isSet() {
```
jameskerr, line 28:
```suggestion
  isSet(): this is {value: zjson.Value[]} {
```

You can fix the build errors with this type guard.

zealot/src/zng/types/record.ts
```diff
@@ -49,7 +53,7 @@ export class Record implements ZngClass<Type[] | null> {
   at(index: number) {
     const col = this.type[index]
     if (!col) throw new Error(`No column at index: ${index}`)
-    const val = this.value[index]
+    const val = this.isSet() ? this.value![index] : null
```
nwt, line 56: @jameskerr: The isSet type guard resulted in `Error: src/js/electron/menu/actions/searchActions.ts(207,31): error TS2352: Conversion of type 'Type[]' to type 'string' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`, so I tried these non-null assertion. Not sure if they're kosher but they seem to do the job.

PR #1244, opened 2020-11-27
add autoupdate restrictions

### LINKED ISSUE #1219: Error checking for linux updates: Invalid latest version format:
On an Ubuntu Linux VM, I happened to start my Brim GA `v0.19.0` app by executing `brim` in a shell. It started fine, but it showed the error message:

```
12:23:18.726 > Error checking for linux updates: Invalid latest version format:
```

I'm suspecting this might actually be a benign side effect of there being no newer version available. However, it'd be good to know for sure.

I've marked this as a `v0.20.0` target just to make sure we give it a look. If someone looks at the code and decides it's benign, we can surely put it off addressing it until later.

mason-fish:
fixes #1219 

Adds the same version check mechanism that we do for linux: if the latest version is *ahead* of the current version then don't update. This is not currently the default behavior of MacOS and Windows, and it is useful for building and sharing beta releases.

The other fix in here has to do with the autoupdate server: if the current version, which we encode in the server url, is the *same* as the latest version that we have in our github releases, then the server returns a 204 status. In my initial probing of the server I had mistakenly thought that the server still returned a body in this case (and contradicted its status) but the server actually does behaves restfully and does not return anything with that 204 code. This is why we were seeing the empty string in the error for the fixes ticket referenced above. 

### DIFF
--- a/src/js/electron/autoUpdater.ts
+++ b/src/js/electron/autoUpdater.ts
@@ -9,47 +9,53 @@ const getFeedURLForPlatform = (platform) => {
   return `https://update.electronjs.org/brimsec/brim/${platform}/${app.getVersion()}`
 }
 
-const autoUpdateLinux = async () => {
-  // Check for updates for MacOS and if there are then we assume there is also one for linux
+const getLatestVersion = async (): Promise<string> => {
+  // Check for updates for MacOS and if there are then we assume there is also one for our other supported OSs
   const url = getFeedURLForPlatform("darwin-x64")
-  try {
-    const currentVersion = app.getVersion()
-    log.info({currentVersion})
-    if (!semver.valid(currentVersion))
-      throw "Invalid current version format: " + currentVersion
+  const resp = await got(url)
 
-    const resp = await got(url).json()
-    log.info({resp})
+  // the update server responds with a 204 and no body if the current version is the same as the
+  // latest version, but will otherwise return json naming the latest version published on github
+  // (even if it is behind the current version)
+  if (resp.statusCode === 204) return app.getVersion()
 
-    const latestVersion = get(resp, "name", "")
-    if (!semver.valid(latestVersion))
-      throw "Invalid latest version format: " + latestVersion
+  const body = JSON.parse(resp.body)
+  const latestVersion = get(body, "name", "")
+  if (!semver.valid(latestVersion))
+    log.error(new Error(`Invalid latest version format: ${latestVersion}`))
 
-    if (semver.gte(currentVersion, latestVersion)) {
-      // up to date
-      return
-    }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Brim Desktop App",
   "repository": "https://github.com/brimsec/brim",
   "license": "BSD-3-Clause",
-  "version": "0.19.0",
```
mason-fish, line 6: I'm using CI to build the releases for me just so I can test this out. Once done I'll switch this back to 19.

jameskerr (commented):
This was not part of your diff, but while reviewing the file I noticed that the line comment and the line below it are not in accord:

https://github.com/brimsec/brim/blob/54484d05718c041a1603519346977341c812fd2f/src/js/electron/autoUpdater.ts#L72-L73

src/js/electron/main.ts
```diff
@@ -47,7 +47,7 @@ async function main() {
   // autoUpdater should not run in dev, and will fail if the code has not been signed
   if (!electronIsDev) {
     try {
-      setupAutoUpdater()
+      await setupAutoUpdater()
```
jameskerr, line 50: Is the await is necessary. I don't think we want to hold up this main function to wait for the autoUpdater code to resolve. The `.catch()` syntax will let the main function continue running.

jameskerr (changes requested):
It seems the code below does the trick at fixing the ticket, so I won't hold it back. But I do notice some duplication in checking the version and returning early. Also some duplication in the timeout checks. 

This code seems so important for users, it makes me nervous that we don't have tests around it. Because this is the intersection of so many systems, it would actually be hard to test. We'd have to mock the api server, the electron app api, the electron dialog api, and the autoupdater api.

Not in this PR, but next time we need to touch this code, I think we should invest in some tests like that.

My only requested change would be to remove that await from the main.ts function.

src/js/electron/autoUpdater.ts
```diff
@@ -84,19 +81,20 @@ export async function setupAutoUpdater() {
...
+  // check for updates 30s after startup
+  setTimeout(updateCb, 30 * 1000)
   // then check for updates once a day
-  setInterval(async () => {
-    const latestVersion = await getLatestVersion()
-    if (semver.gte(app.getVersion(), latestVersion)) return
-
-    autoUpdater.checkForUpdates()
-  }, 24 * 60 * 60 * 1000)
+  setInterval(updateCb, 24 * 60 * 60 * 1000)
```
jameskerr: Nice!

jameskerr (approved):
Thanks for the changes!

PR #1247, opened 2020-12-01
Sidebar Sectional 

jameskerr:
In this PR we've got some new patterns.

I've introduced the `/src/pkg` directory. This is where we can put complete packages that might as well be their own node module. They do not have any relative imports other than within their own folder. Keeping them in the src directory keeps the overhead down.

Our first package is `sectional`.

Sectional is a component that allows a view to be split into 2 or more sections. Each section can be resized and toggled. Here's how it's used.

```jsx
const [sections, setSections] = useState([
  {id: "a", min: 100, closedSize: 24},
  {id: "b", min: 200, max: 800, closedSize: 24}
])

<Sectional sections={sections} onChange={setSections}>
  {(data, provided) => (
    <div key={data.id} style={provided.style}>
      <hr {...provided.resizeProps} />
      <button {...provided.toggleProps}>Toggle</button>
    </div>
  )}
</Sectional>
```

The Sectional props:

```ts
type Props = {
  sections: SectionData[]
  children: (data: SectionData, provided: Provided) => React.ReactNode
  onChange: (views: SectionData[]) => void
}
```

The SectionData interface looks like so:

```ts
interface SectionData {
  id: string
  size?: number
  min?: number
  max?: number
  isOpen?: boolean
  closedSize?: number
}
```

The function it expects as its only child is passed the `SectionData` as the first argument, and the `Provided` object as the second.

```ts
type Provided = {
  resizeProps: {
    onMouseDown: (e: MouseEvent) => void
    style: object
  }
  toggleProps: {
    onClick: (e: MouseEvent) => void
  }
  style: object
}
```

* `resizeProps` should be given to the element the user should drag when resizing. 
* `toggleProps` should be given to the element the user will click to show/hide the section. 
* `style` should be put on the section container element

In practice, you would switch on the section id to know which section to render.

```jsx
<Sectional sections={sections} onChange={setSections}>
  {(data, provided) => {
    if (data.id === "spaces")
      return (
        <SpacesSection
          isOpen={data.isOpen}
          key={data.id}
          {...provided}
        />
      )
    if (data.id === "history")
      return (
        <HistorySection
          isOpen={data.isOpen}
          key={data.id}
          {...provided}
        />
      )
    return null
  }}
</Sectional>
```

Other notable features:
* The look of each section is up to the caller. This package only provides the dimensions and behavior.
* The sectional component will intelligently grow and shrink the sections to fit their container when the window is resized.
* Animations for opening/closing a section

In the future, it seems possible to allow the user to drag and sort the sections

Here is the app using the new sectional component.

ZQDGsnHKde


src/js/state/migrations/202012011232_sidebarSections.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  // @ts-ignore
+  for (const {state} of windows) {
+    for (const tab of state.tabs.data) {
+      expect(tab.layout.sidebarSections).toEqual([
+        {id: "spaces", isOpen: true},
+        {id: "history", isOpen: true}
+      ])
+      expect(tab.layout.spacesIsOpen).toBe(undefined)
+      expect(tab.layout.historyIsOpen).toBe(undefined)
+      // I decided the height of each section isn't worth saving during the migration
```
mason-fish, line 18: 👍

src/pkg/sectional/models/Controller.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  serialize() {
+    return {
+      size: this.size,
+      sections: this.map((c) => c.serialize())
+    }
+  }
+
+  private distribute(extra: number) {
```
mason-fish, line 56: yay access modifiers!

mason-fish (approved):
This is great, James! I like the move towards class-defined objects too, I was hoping we'd want to do that more after moving to typescript.

I'm a little hesitant about the `pkg` dir, since it will end up spreading our code over even more territory/directories, but I'm willing to go forward on faith that organizing non-stylistic (those would go in `common`) AND independent components/code could eventually be useful to us if we want to reuse things on a different property or if we ever made something so useful to the general community that we'd want to publish it separately.

Issue #1218, opened 2020-11-15
Intermittent problems launching Brim v0.19.0+ on a Windows VM over Remote Desktop


philrz:
As part of some test work, @henridf happened to try installing GA Brim tagged `v0.19.0` onto a Windows 2019 Server VM on Google Cloud and found it would not start. I reproduced this on a Windows Server 2019 Datacenter VM and saw the same. It comes up with a white blank screen. The DevTools console shows no error messages.

Other details:
* It is intermittent. Sometimes it _does_ come up ok.
* While I initially spotted it on the first app launch that happens at the tail end of installation, I've also reproduced it by manually killing all processes from the hung app via the Task Manager and relaunching it via the icon on the desktop.
* Thinking the problem might be unique to the Windows builds on Google Cloud, on AWS EC2 I created a Microsoft Windows Server 2019 Base VM, but it also failed to start `v0.19.0`, showing the same blank white screen.
* I've since reproduced the same intermittent failure on a Windows Server 2016 Datacenter VM on Google Cloud and a Windows 10 VM in Azure as well.
* I _can't_ seem to reproduce the problem with GA Brim `v0.18.0`. I've tested this on both Windows Server 2016 & 2019 Datacenter VMs on Google Cloud, 
* As part of Suricata testing, I've had reason to test with other release artifacts based on Brim `master` commits since GA v0.19.0 was released and the symptom reproduces in those as well.
* I've _never_ seen this problem on my Windows 10 laptop that I use for most Brim testing, leading me to wonder if triggering the problem is somehow specific to when working on a VM and/or over Remote Desktop.

This particular "blank white screen" is a little different from others we've seen in the past, since it _does_ have the menu bar present. Example:

![image](https://user-images.githubusercontent.com/5934157/100555081-14a30d00-324e-11eb-94af-38f174b284a8.png)

Here's an example of how the logging looks when it's in this state:

```
C:\Users\phil\AppData\Roaming\Brim\logs>dir
 Volume in drive C has no label.
 Volume Serial Number is DEA0-1188

 Directory of C:\Users\phil\AppData\Roaming\Brim\logs

11/29/2020  10:17 PM    <DIR>          .
11/29/2020  10:17 PM    <DIR>          ..
11/29/2020  10:17 PM             1,059 main.log
11/29/2020  10:17 PM                 0 zqd-core.log
               2 File(s)          1,059 bytes
               2 Dir(s)  33,007,079,424 bytes free
```

[...]

PR #1248, opened 2020-12-01
Disable gpu hardware acceleration

### LINKED ISSUE #1242: Brim release v0.19.0 fails to render in Parallels Ubuntu 18.04
Using a fresh install of `Ubuntu 18.04.5` with Parallels 16 for Mac (Version 16.1.1) and doing a fresh install of the Brim release `v0.19.0` there appears to be a rendering issue. This does not appear on the `v0.18.0` release, and it seems that this only occurs when using Parallels to manage the virtualization of Ubuntu. Here is the result I saw:

The only error in any of the logs is the same issue described in https://github.com/brimsec/brim/issues/1219 but the handling of that error is just to log and proceed, so it should not be related to this.

@alfred-landrum confirmed the same on Parallels with a fresh install of `Ubuntu 18.04.4`: 

While @philrz saw no issues while using VirtualBox to manage his Ubuntu VM.

@alfred-landrum found a workaround and possible clue of this behavior by opening the app with the cli and passing in the `--disable-gpu` flag (`/usr/lib/brim/Brim --disable-gpu`), which allows the app to load normally

jameskerr:
After disabling the hardware acceleration, the only issue revealed was the backdrop-filter css property. All the modals have a solid color background instead of the freakin-sweet background blur they used to have. Don't worry background blur, I will come back for you one day.

Fixes #1242 and #1218

🦜🦜 + 🪨 = ☠️☠️

### DIFF
--- a/src/js/components/ModalDialog/ModalDialog.tsx
+++ b/src/js/components/ModalDialog/ModalDialog.tsx
@@ -21,9 +21,6 @@ const Overlay = styled.div`
 `
 
 const Background = styled.div`
-  --blur-color: hsla(32, 5%, 85%, 0.5);
-  --blur-shadow: hsla(32, 5%, 20%, 0.5);
-
   pointer-events: all;
   display: flex;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
Dang, we just added that too! Oh well

PR #1257, opened 2020-12-05
[experiment] investigate alert.pcap integration test failures

henridf:
experiment to see if disabling suricata-update helps with CI consistency.


jameskerr:
Hey @henridf, which tests are being inconsistent? The alert.pcap ingest test?

henridf:
> Hey @henridf, which tests are being inconsistent? The alert.pcap ingest test?

@jameskerr yes, that's the one.

jameskerr:
Got it. Does the suricate updater only run when suricate is called for the first time? Or does it run on app startup?

henridf:
It runs on app startup (at zqd startup to be specific). But given that disabling it didn't appear to resolve the issues, I've moved on to timeout...

alfred-landrum:
@henridf : FYI that this bug I introduced (and fixed) might be related to this CI flakiness:
https://github.com/brimsec/zq/issues/1797
https://github.com/brimsec/zq/pull/1798

That bug causes the time range for a newly created space to be truncated to whatever is reported from an initial 'get info' operation on the space. If that 'get info' happens before all of the records have been processed & stored in the space's zng file, I don't think the tests would ever see them.

jameskerr:
I'm going to close for now until we have some more cycles to spend on getting this test to pass. Even with #1274 going in, the suricata test still is flakey. The next step would be to burden the CPU locally to re-created this locally and see where it is failing. It appears to be a different issue than the one that #1274 was targeting.

PR #1260, opened 2020-12-07
Mention install guide in README

philrz:
Now that we have the install guide in the wiki, we might as well link to it from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@ Brim is built from open source components, including:
 
 ## Installing Brim
 
-The easiest way to install Brim is to download the installation package for
-your platform from the latest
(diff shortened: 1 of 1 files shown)


philrz:
The failure of the markdown link checker didn't make sense to me because I could easily `wget` the same URL and see it working. Digging revealed some breakage in the some downstream tooling that the Actions Workflow depends (https://github.com/tcort/markdown-link-check/issues/142). Rather than put in an exception that we might forget to unwind, I'm inclined to let the Markdown link checker fail on this one URL for the moment on the assumption this will get fixed pretty fast. There's already an issue in the Actions repo for it as well, so people are quickly catching on.

mason-fish (approved):

PR #1262, opened 2020-12-08
Alerts in UID correlation

jameskerr:
This PR adds suricate alerts to the UID correlation panel. It also adds a red background to the values in an "event_type" field. There sort order of the suricate records will be resolved in a different PR.

At the moment, there is no correlation when clicking the suricate alert. For this we'll do a grouping on the community_id and display the items in a list. Expect that soon in a follow up PR.

Screen Shot 2020-12-07 at 8 58 37 PM 

Screen Shot 2020-12-07 at 8 58 31 PM

Screen Shot 2020-12-07 at 8 58 47 PM


philrz:
@jameskerr: Looking good so far! One thing I noticed is that the Log Detail pane is now pre-opened on a fresh app  whereas it wasn't in the most recent GA release. Is that intentional?

image

jameskerr:
@philrz thanks! Yes that is intentional. I feel like the log detail pane is so important, but so hidden by default. You have to know to click "View => Show Right Pane". What do you think?

philrz:
@jameskerr: Yes, I'm fine with it being there by default. It's probably worth floating to the rest of the team, though. I seem to remember we've had debates at times about which of the panels should be open by default, how the pop-out Log Detail window relates to that, etc., and I don't recall all the opinions that led to our current approach.

src/css/_field-cell.scss
```diff
@@ -15,7 +15,8 @@
     }
   }
 
-  &._path {
+  &._path,
+  &.event_type {
```
jameskerr, line 19: This is how we give the event_type field a red background in the various places in the app.

src/js/components/LogDetails/UidPanel.tsx
```diff
@@ -1,45 +1,26 @@
...
-
   return (
     <div
       className="correlated-logs-panel detail-panel"
       {...reactElementProps("correlationPanel")}
     >
-      <PanelHeading status={status}>Uid Correlation</PanelHeading>
-      {logs.length === 0 && (
+      <PanelHeading status={status}>Correlation</PanelHeading>
+      {logs.length === 0 && status !== "SUCCESS" && (
```
jameskerr, line 20: Sometimes if the search returns no logs, it would show to loading screen forever. Checking that the status is not success helps avoid that.

src/js/components/LogDetails/index.tsx
```diff
@@ -8,22 +8,21 @@ import LogDetails from "../../state/LogDetails"
...
 
 export default function LogDetailComponent() {
-  const log = useSelector(LogDetails.build)
-  const zeek = createZeekLog(log)
-  console.log(log)
-  if (!log) return <NoSelection />
+  const record = useSelector(LogDetails.build)
+  if (!record) return <NoSelection />
 
+  const correlation = new Correlation(record)
```
jameskerr, line 17: I may or may not do away with this class in the future. It just basically knows to to find a uid and community_id from a record.

In the future we may use the ZeekEvent and SuricataEvent classes (which we store in the ppl directory), that wrap a record and have knowledge of the Zeek and Suricata fields.

src/js/correlation/components/ZeekCorrelation.tsx
```diff
@@ -1,27 +1,25 @@
...
-type OwnProps = {
+type Props = {
   logs: zng.Record[]
   log: zng.Record
 }
 
-type Props = DispatchProps & OwnProps
-
-export default function UidTimeline({logs, log, dispatch}: Props) {
+export default function ZeekCorrelation({logs, log}: Props) {
```
jameskerr, line 21: I've moved the component for the UID Correlation into the `src/js/correlation` folder. I've also moved all supporting components/classes/hooks into this folder. This folder will be a ppl folder soon as well.

src/js/correlation/flows/fetch.ts
```diff
@@ -0,0 +1,52 @@
...
+    const records = await dispatch(getFromUid(uid))
+    const conn = findConn(records)
+    if (conn && conn.has("community_id")) return dispatch(getFromConn(conn))
+    else return records
+  } else if (cid) {
+    return dispatch(getFromCommunityId(cid))
+  } else {
+    return []
+  }
+}
```
jameskerr, line 22: This is the core logic of which searches we run to find correlated events.

src/js/correlation/flows/stubs.ts
```diff
@@ -0,0 +1,1869 @@
+import {ZealotPayload} from "../../../../zealot/dist/types"
+
+export const uidAndCommunityResult: ZealotPayload[] = [
```
jameskerr, line 3: This is a stub for zealot to use in the tests. I just copied and pasted the zjson responses from the network tab in the app after making a query. I'm not sure if this is best place for stubs to live. Open to ideas. It's nice that since this is just a stub for these tests that it lives right next to it. I just don't like that it will be included in the final build. In fact, I don't like like that all our tests are included in the final build.

As an aside, I've been thinking a lot about folder structure. It would be great to have a more scalable structure with a known place for test support files like this (files that are used for tests that don't have a ".test.ts" extension). Then in a build step we could strip out all the tests in the build.

src/js/correlation/models/BrimEvent.ts
```diff
@@ -0,0 +1,21 @@
+import {zng} from "../../../../zealot/dist"
+import {UnknownEvent} from "./UnknownEvent"
+import {SuricataEvent} from "./SuricataEvent"
+import {ZeekEvent} from "./ZeekEvent"
+
+export interface BrimEventInterface {
+  getType: () => string
+  getTime: () => Date
+}
```
jameskerr, line 9: This is my first step towards some classes that "know" about Zeek and Suricata fields.

jameskerr: I don't know if these classes should contain the code that's in the Correlation class. More will be revealed.

src/js/flows/search/mod.ts
```diff
@@ -5,14 +5,17 @@ import {handle} from "./handler"
...
-  from: Date
-  to: Date
-  spaceId: string
-  id: string
-  target: SearchTarget
+  from?: Date
+  to?: Date
+  spaceId?: string
+  id?: string
+  target?: SearchTarget
```
jameskerr, line 18: I just added sensible defaults for all these fields so that we don't have to pass them in every time. 

from => whatever the start time of the current tab is
to => the end time of the current tab
spaceId => the space id of the current tab
id => a random hash (this is used to cancel a request if needed)
target => "events" instead of "archive".

src/js/flows/viewLogDetail.ts
```diff
@@ -1,16 +1,27 @@
...
-    dispatch(LogDetails.push(log))
+  if (!isEqual(record, current)) {
+    dispatch(LogDetails.push(record))
+
+    if (!Current.getSpaceId(getState())) return // It is possible for this code to be called without a space
+    dispatch(LogDetails.updateUidStatus("FETCHING"))
+    dispatch(fetchCorrelation(record))
+      .then((records) => {
+        dispatch(LogDetails.updateUidLogs(records))
+        dispatch(LogDetails.updateUidStatus("SUCCESS"))
```
jameskerr, line 23: The name "uidLogs" is wrong now since it could be uid or community_id logs. I think "relatedLogs" would be better. A future PR...

src/js/searches/programs.ts
```diff
@@ -14,7 +17,47 @@ export function filenameCorrelation(md5: string) {
...
+}
+
+export function connCorrelation(conn: zng.Record) {
+  const uid = conn.get("uid") as zng.Primitive
+  const cid = conn.get("community_id") as zng.Primitive
+  const ts = (conn.get("ts") as zng.Primitive).toDate()
+  const dur = (conn.get("duration") as zng.Primitive).toFloat() + 90 // Add a 1.5 minute buffer for events that get logged late
+  const endTs = new Date(new Date(ts).getTime() + dur * 1000)
+
+  const cidFilter = zql`community_id = ${cid} and ts >= ${ts} and ts < ${endTs}`
```
jameskerr, line 60: Ok, so I'm pretty stoked about this. I'm using a template literal function for the first time. It allows us to interpolate complex objects into the zql string and smartly translate them into valid zql. 

For example, we can pass in a JS Date object and it will convert it to time.

```js
const ts = new Date(123)
const query = zql`ts >=${ts}`

console.log(query)
// "ts >= 00000123.0"
```

We could even pass whole zng.DataTypes and it could convert them to their equivalent zql strings.

```js
const uid = conn.get("uid") // => zng.Primitive class
zql`uid={uid}` => 'uid="abc123k48%@4kdfsf"'
```

src/js/state/Layout/reducer.ts
```diff
@@ -1,8 +1,8 @@
 import {LayoutState, LayoutAction} from "./types"
 
 const init: LayoutState = {
-  rightSidebarIsOpen: false,
-  rightSidebarWidth: 450,
+  rightSidebarIsOpen: true,
+  rightSidebarWidth: 300,
```
jameskerr, line 5: I'll revert this change now that we've talked about it today.

src/js/zql/toZql.ts
```diff
@@ -0,0 +1,34 @@
+import {zng} from "../../../zealot/dist"
+
+export function toZql(object: unknown): string {
```
jameskerr, line 3: This will grow with time as we add support for more and more types of objects.

zealot/src/zjson.ts
```diff
@@ -1,7 +1,7 @@
 export interface Item {
   id: number
   schema?: Schema
-  aliases?: string[]
+  aliases?: PrimitiveColumn[]
```
jameskerr, line 4: The change and the rest were just incomplete typings of the zjson responses from zqd. The stubs that I copied and pasted found these.

src/js/searches/programs.ts
mason-fish: 🔥

zealot/src/zqd.ts
```diff
@@ -22,9 +23,18 @@ export type SearchWarnings = {
...
-  start_time: number
+  update_time: Ts
+  start_time: Ts
   bytes_read: number
+  bytes_matched: number
+  records_read: number
+  records_matched: number
+}
+
+export type SearchEnd = {
```
mason-fish, line 34: I saw this in the stub you included, was the api sending these payload types before? or is this new?

mason-fish (approved):
Thanks for annotating the PR, looks good! I think my only concerns relate to structuring code in a way that can accommodate more generic data sources, which I can see from your comments is on your mind as well. I'm down to wait and plan that refactor later though! 👍

src/js/correlation/models/BrimEvent.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class BrimEvent {
+  static build(r: zng.Record) {
+    if (r.has("_path")) {
```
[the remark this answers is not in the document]
jameskerr: I suppose there is. It is possible to ingest generic data with both. But with our use cases of Zeek and Zuricata, it appears they don't overlap.

zealot/src/zqd.ts
jameskerr: Yes it was sending these before, it's just the types that were incomplete.

PR #1272, opened 2020-12-10
Query Library UI

### LINKED ISSUE #1081: Query Library frontend
@orochford had some awesome suggestions for a query library in his "Initial Brim Query Use Cases" document. Let's implement them!

Some of the constraints we had discussed for this:
* The query lib would be a third section within the left side panel
* It would organize queries by type/category, each of which are collapsible
* Query libs can be imported/exported as .json through user prefs, and we will provide an initial lib as a sort of "getting started" template with common queries in it
* Queries will be annotated with a display name, which is what is shown in the library, and that name along with it's underlying query can also be edited/deleted
* Queries are space-agnostic: by selecting a given query, it will be applied to whatever current space you are on, or not at all if no space is open.

mason-fish:
fixes #1081 

Includes the remaining UI elements for the Query Library. For now I have left out a couple things from the designs, specifically: the tag coloring (seen while editing/creating a query), search box, and hover tooltip. The PR was already getting a bit large and none of those pieces seemed trivial or immediately necessary so I thought I'd planned to put them in a future iteration.

I ended up re-modelling the access patterns to our data structure, from `indexPath: number[]` to simply search for a given element by it's id in the tree. And this is only done for deletion, insertion, and moving items. The performance cost to this is okay because on those operations, we would have to create a new copy of the whole data structure and give it to redux anyway. By going this route, we can also now use the `tree-model` lib to assist with some of the internal tree manipulating work involved in those operations.

image

image

image

image

image

image


jameskerr (commented):
* The list items should be in a disabled state if no space is selected, with title text saying  "You must have a space selected to run queries". 
* The option for "Run Query" should be disabled if no space is selected.
image

This is tight!

src/js/components/LeftPane.tsx
```diff
@@ -284,21 +295,138 @@ function HistorySection({isOpen, style, resizeProps, toggleProps}) {
...
+              else dispatch(Queries.removeItems([item]))
+            }
+          })
+      }
+    }
+  ]
 
-  function onItemClick(e, item) {
-    console.log("clicked", item)
+  const menu = usePopupMenu(template)
```
jameskerr, line 390: I know I'm the one who wrote this, but I think this is pattern for context menus is making our app slow. Anytime data needed only in the context menu gets updated, this whole component has to re-render.

A better pattern is to wait until the user has right clicked, then dispatch a thunk. Then in the thunk we gather all the state needed for the context menu and show it.

src/js/components/LeftPane.tsx
```diff
@@ -321,3 +453,31 @@ function QueriesSection({isOpen, style, resizeProps, toggleProps}) {
...
+    ...[
+      {
+        label: "Filter by tag",
+        enabled: false
+      },
+      {type: "separator"}
+    ]
+  )
+
+  const menu = usePopupMenu(template)
```
jameskerr, line 475: Same goes for this pop menu.

src/js/state/Queries/initial.ts
```diff
@@ -6,24 +6,104 @@ const init = (): QueriesState => ({
...
       id: "1",
-      name: "Count by Path",
-      zql: "* | count() by _path",
-      description: "",
-      tags: []
+      name: "Activity Overview",
+      value: "count() by _path | sort -r",
+      description:
+        "This query shows a list of all Zeek streams in the data set, with a count of associated records",
+      tags: ["zeek", "initial exploration"]
```
jameskerr, line 13: hmm, I wonder if we should allow tags with spaces in them.

src/js/state/Queries/selectors.ts
```diff
@@ -1,6 +1,37 @@
-import {QueriesState} from "./types"
+import {QueriesState, Query} from "./types"
 import {State} from "../types"
+import TreeModel from "tree-model"
+import {includes} from "lodash"
 
 export default {
-  getRaw: (state: State): QueriesState => state.queries
+  getRaw: (state: State): QueriesState => state.queries,
+  getTags: (state: State): string[] => {
```
jameskerr, line 8: I'm not sure if you're using this or not in the code above. But if you are, use createSelector from releselect here to derive the state so that the reference of the return value is the same when the args are the same and components don't re-renderer when they don't need to.

jameskerr: Same goes for getQueriesByTag below.

jameskerr (changes requested):
Good stuff man! A few comments below. It also might be time to start breaking up the LeftPane file. We could at least make each section it's own file. But feel free to push that into another PR.

src/js/state/Queries/initial.ts
mason-fish: Good question, if I am the user then I'd prefer to be able to use multiple words with spaces. Looks cleaner. But it does complicate how we detect start and end when we go to edit/create

mason-fish:
@jameskerr thanks for the great comments, I've addressed them all but the first: creating a query *does* work it turns out, but for some reason the inner list component starts cutting off the last item after a certain length. I'm still looking into that one

src/js/state/Queries/selectors.ts
```diff
@@ -1,37 +1,24 @@
...
-        if (!n.model.tags) return true
-        n.model.tags.forEach((t) => {
-          tagMap[t] = true
-        })
+export const getRaw = (state: State): QueriesState => state.queries
 
-        return true
+export const getTags = createSelector<State, QueriesState, string[]>(
+  getRaw,
+  (queries): string[] => {
```
jameskerr, line 10: Nice!

jameskerr (approved):
Good stuff! 

For another PR: After creating a new query, scroll to it in the queries list.

PR #1275, opened 2020-12-11
Round the pixels before distributing them

### LINKED ISSUE #1256: Zoom-in freezes Brim
I happened to be giving a demo today on my Mac of a recent Brim commit to show off Suricata and tried zooming in via Command+ and it froze the app. I just did a binary search to see where it broke and I can confirm that thankfully the last GA release `v0.20.0` is healthy, but the more recent commit where it started freezing is `c7d34e8`, which is associated with #1247 (cc: @jameskerr).

jameskerr:
Fixes #1256 

The "sectional" component distributes pixels evenly to all the sections on start up. It does this by adding 1 pixel to each section in a while loop. It stops when each section is at it's max, or when there is are more pixels left to distribute. `if (left === 0) break`

To check if each section is at it's max, it keep the leftover pixels from the previous iteration. If the current amount of pixels to distribute equals the amount from the previous loop, that means each section is at it's max, so break the loop. `while(prev !== left)`

What I didn't realize is that when you zoom in a window, you can have elements with a fractional height (421.365px). Because the distributor adds one whole pixel each loop, the remaining space variable never was exactly zero. Since the sections do not have a "max" size by default, this caused an infinite loop and the screen froze.

I now round the number to the nearest whole number before starting the loop. I've added a test to test for fractional pixels as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     ],
     "testURL": "http://localhost",
     "roots": [
-      "./src/js"
+      "./src"
     ],
     "collectCoverageFrom": [
       "**/*.js",
--- a/src/pkg/sectional/models/Controller.test.ts
+++ b/src/pkg/sectional/models/Controller.test.ts
@@ -33,6 +33,24 @@ test("distribute positive beyond max", () => {
   expect(views.at(3).size).toBe(700)
 })
 
+test("distribute a fractional amount", () => {
+  const data = {
+    size: 1000.1,
+    sections: [
+      {id: "1", min: 24, max: 100},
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     ],
     "testURL": "http://localhost",
     "roots": [
-      "./src/js"
```
jameskerr, line 48: Jest was not running the tests in `src/pgk`. It does now with this change.

mason-fish (approved):
Nice catch! Also, thanks for the clear explanation in the description

Issue #1281, opened 2020-12-15
Missing space list and open space details on Windows at startup


alfred-landrum:
I saw this issue with the Brim v0.21.0 release candidate on Windows: Occasionally when I opened Brim, I would see that 'spaces' left side menu was empty, and expanding/collapsing it didn't make a difference, and the details under the space name were "nan undefined 0ms".

I wondered if this was some type of zqd issue, so I made sure that I could use zapi to get the expected space details. I also reproduced this when I already had zqd running when Brim started.

<img width="412" alt="image" src="https://user-images.githubusercontent.com/704945/102161591-75306d80-3e3c-11eb-98ce-88fa5be8ef54.png">

jameskerr:
@alfred-landrum was there a series of steps to reproduce this. Or at least have a good chance of reproducing this?

alfred-landrum:
@jameskerr : I saw it by selecting a space (ensuring the default search view was visible), then repeatedly closing the app, waiting a second & re-opening it, and it would eventually manifest.

philrz:
One thought: The way the update logic works today, despite the RC being versioned as v0.21.0, the app would be trying to check for a newer version and auto-update itself in the background. At one point on my Mac I'd though I'd repro'ed this issue but I'd forgotten that I'd already seen & acknowledged the "Later" button to restart at the auto-updated v0.20.0. It seems like downgrading to v0.20.0 _definitely_ makes the Spaces disappear, as I've witnessed this on my Mac. Maybe you were not being downgraded and witnessed a different issue (I didn't witness any "nan undefined 0ms", just empty Spaces), but just wanted to put it out there.

philrz:
I've attached a video of a repro that's quite consistent on Windows using yesterday's Brim `v0.21.0` release candidate that was based on commit `6bb51cb`. The attached [hello.zip](https://github.com/brimsec/brim/files/5699523/hello.zip) is imported (must be unzipped first), once the "Import complete" pop-up is acknowledged, select File->Exit from the drop-down menu and immediately relaunch Brim. The Spaces menu is empty and you can see the "NAN UNDEFINED" as shown in the original repro screenshot.

[Repro.zip](https://github.com/brimsec/brim/files/5699522/Repro.zip)

I did also successfully repro earlier on macOS using the same steps.

philrz:
[...]

PR #1283, opened 2020-12-15
Set max to infinity if its null

jameskerr:
If you pass `undefined` to a function with a default parameter, the default parameter will be used. However if you pass `null` to it, it will use `null` as the value.

The constructor of the Section model has a max parameter with a default value of `Infinity`. But when you serialize an object that contains `Infinity`, it gets converted to null.

```
> let myObject = {max: Infinity}
undefined
> JSON.stringify(myObject)
'{"max":null}'
>
```

So the next time we passed these serialized args to the constructor, the max value was set to `null` instead of `Infinity`. That caused an issue when "clamping" the size of the section, resulting in this bug.

I've added some tests.

fixes: #1281

### DIFF
--- a/src/js/components/AddSpaceButton.tsx
+++ b/src/js/components/AddSpaceButton.tsx
@@ -1,8 +1,7 @@
 import {useDispatch} from "react-redux"
 import React from "react"
 import styled from "styled-components"
-
-import {resetTab} from "../flows/initNewTab"
+import {setSpaceId} from "../state/Current/actions"
 
 const StyledAnchor = styled.a`
   margin-left: auto;
@@ -21,7 +20,7 @@ const StyledAnchor = styled.a`
 
 export default function AddSpaceButton() {
   const dispatch = useDispatch()
-  const onClick = () => dispatch(resetTab())
+  const onClick = () => dispatch(setSpaceId(null))
 
   return (
     <StyledAnchor className="add-space" onClick={onClick}>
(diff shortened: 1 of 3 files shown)


src/js/components/AddSpaceButton.tsx
```diff
@@ -21,7 +20,7 @@ const StyledAnchor = styled.a`
 
 export default function AddSpaceButton() {
   const dispatch = useDispatch()
-  const onClick = () => dispatch(resetTab())
```
jameskerr, line 24: A bug with the "isOpen" arrow made be realize that this was overkill for what this button needs to do. It just needs to set current space to null.

src/pkg/sectional/models/Section.ts
```diff
@@ -20,6 +22,7 @@ export default class Section {
     public isOpen: boolean = true,
     public closedSize: number = min
   ) {
+    if (!isNumber(max)) this.max = Infinity
```
jameskerr, line 25: Fixed it when it comes in.

src/pkg/sectional/models/Section.ts
```diff
@@ -66,7 +69,7 @@ export default class Section {
       id: this.id,
       size: this.size,
       min: this.min,
-      max: this.max,
+      max: this.max === Infinity ? undefined : this.max,
```
jameskerr, line 72: Fix it when it goes out.

mason-fish (approved):
Nice! thanks for adding tests too 👍

PR #1287, opened 2020-12-16
Electron on windows, instanceof bugs

### LINKED ISSUE #1280: Error "Data.getValue(...).map is not a function" on Windows when clicking records for imported pcap
@alfred-landrum and I both saw this issue when doing smoke tests on Windows while prepping the Brim v0.21.0 release candidate (Brim commit `6bb51cb` talking to GA `zqd` tagged `v0.24.0`). We couldn't seem to trigger the same issue on macOS or Linux.

I've triggered the issue with a wrccdc pcap as well, but this small one triggers it easily (unzip before importing):

[hello.pcapng.zip](https://github.com/brimsec/brim/files/5692814/hello.pcapng.zip)

To repro:

1. Drag the pcap into Brim
2. Wait for the "Import Complete" pop up
3. Click on one of the Zeek events such as `conn` or `http`

At this point I was hoping to see the Packets button activate, but instead I get this error pop-up:

image

If I click the `files` event instead, I get the different error:

image

I first reproduced this on my personal Windows 10 laptop, but have now also reproduced it on a scratch Windows Server 2019 VM on Google Cloud.

mason-fish:
fixes #1280 

The way we type check using `instanceof` has different results on windows electron than it does in other platforms. The cause of this is still something to chase down, and could be related to the way our import paths get transpiled by babel, but for now this is a fix that will get us by. Here are some of the related tickets I found on the topic:

https://github.com/xOPERATIONS/maestro/issues/121
https://github.com/electron/electron/issues/1586
https://github.com/electron/electron/issues/1289

### DIFF
--- a/src/js/correlation/models/Correlation.ts
+++ b/src/js/correlation/models/Correlation.ts
@@ -24,18 +24,20 @@ export class Correlation {
   }
 
   getUid() {
+    console.log("r is: ", this.r)
     if (this.r.has("_path")) {
       const path = this.r.get("_path").toString()
       const name = get(specialUids, path, "uid")
+      console.log("name is: ", this.r)
       if (this.r.has(name)) {
         const data = this.r.get(name)
+        console.log("data is: ", data)
         if (data instanceof zng.Primitive) {
           return data.toString()
         } else {
-          return data
-            .getValue()
-            .map((v) => v.toString())
-            .join(" ")
+          const value = data.getValue()
+          if (!value) return null
+          return value.map((v) => v.toString()).join(" ")
         }
       }
     }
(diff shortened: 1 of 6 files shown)


src/js/correlation/models/Correlation.ts
```diff
@@ -32,10 +32,10 @@ export class Correlation {
         if (data instanceof zng.Primitive) {
           return data.toString()
         } else {
-          return data
-            .getValue()
-            .map((v) => v.toString())
-            .join(" ")
+          const value = data.getValue()
+          if (!value) return null
+          if (isString(value)) return value
```
mason-fish, line 37: this wasn't actually the main issue, but it's safer code now so I'm including it

jameskerr (approved):
Oh great! I'm so glad that fixed it.

PR #1297, opened 2020-12-21
Better Scripts

jameskerr:
Adds the following scripts:

* ./scripts/clean.js
* ./scripts/post-install.js
* ./scripts/build.js
* ./scripts/start.js

Scripts are ignored by typescript so that we can easily run them without a compile step. They should all be in vanilla node .js files with `module.exports` and `require`.

**Build**

image

**Post-Install**

It is now possible to skip downloading the zeds during an npm install. This is helpful if you've already got good enough zdeps for your feature, or if you are on a slow network. Like so: `ZDEPS=false npm install`. Note that unlike other npm scripts, install cannot receive any arguments in the form of -- --arg1 because it has its own argument logic in order to install specific packages. Hence, the env var workaround.

image

Or the normal install:

image

**Start**

You can now skip the initial build when running npm start. It will still watch for changes and recompile, but you don't have to pay the up-front cost if you've already compiled recently.

The command for this is: `npm start -- --no-build`

image

Or the normal npm start

image


mason-fish (approved):
Nice, that `package.json` is looking much tidier now!

scripts/start.js
```diff
@@ -0,0 +1,22 @@
+const {bold} = require("chalk")
+const run = require("./util/run")
+const flags = require("./util/flags")
+const {JS, SCSS, STATIC} = require("./util/commands")
+
+async function start() {
+  if (!flags.noBuild) {
+    await run("node", "scripts/build")
+  } else {
+    console.log(bold("Skipping build step"))
```
mason-fish, line 10: This is useful, good call!

PR #1302, opened 2020-12-23
Export ndjson csv

### LINKED ISSUE #1090: Add NDJSON/CSV export options
Right now the "Export" button only generates ZNG output. It's not too difficult for users to run that ZNG through `zq` to generate any number of other supported output formats. However, "vanilla" endpoints were recently added to `zqd` that return NDJSON and CSV respectively (https://github.com/brimsec/zq/issues/1275, https://github.com/brimsec/zq/issues/1276). As a convenience to the user, it would be great if the Export button offered additional options for NDJSON and CSV that returned the data from these endpoints.

Currently with "Export", the user chooses a filesystem location for output as a file. An additional nice-to-have might be if the user were also offered a single-click button that would instead copy the output directly into the user's paste buffer. This would facilitate quick pivots of CSV data into Excel or Google sheets, or NDJSON data into web-based tools that accept it.

In a group discussion we reached consensus that we'll only want to add this in Brim once we've implemented https://github.com/brimsec/zq/issues/1270 and https://github.com/brimsec/zq/issues/1271 since this will ensure users always get successful CSV output regardless of how much data is returned by their query or if the output cross heterogeneous record types.

mason-fish:
fixes #1090 

Adds an export modal with, for now, a simple format choice to be made before we send the user along to the existing save-file-to flow. This enables them to specify the csv and ndjson formats which zqd now offers.

image

image
image

image

image


philrz:
@mason-fish: It's cool that you put the functionality in here to "gray out" CSV export if the records are non-uniform. But if you're hitting a `zqd` endpoint to get the data for export, @nwt's pending work on https://github.com/brimsec/zq/issues/1271 should render that unnecessary, as I understand it.

src/js/components/ExportModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </RadioItem>
+      </RadioButtons>
+      <Footer>
+        <ButtonGroup>
+          <ToolbarButton text="Close" onClick={onClose} />
+          <ToolbarButton
+            {...defaultModalButton.props}
+            isPrimary
+            text={isExporting ? "" : "Export"}
+            icon={isExporting ? <MacSpinner light /> : null}
```
jameskerr, line 115: It think we should allow the user to continue to use the app while the data is being exported. If I'm reading this right, this waits until the export is finished before dismissing?

jameskerr: I do appreciate the desire to be consistent with the popNotices. I think we need a universal place to show "downloading" statuses. Right now we have:

1. Space import progress
2. Pcap download progress
3. Export file progress

They each have their own progress indicators. I'll think of a design for this.

jameskerr (changes requested):
Nice Mason! Looks great. I have a few stylistic edits for you:

1. Make each radio label a click target for the label. Sometimes this is done by wrapping the <input> element with the <label> element. Other times you can use <label for={name}/> I think...
2. Make the cursor for each label `cursor: default`. I think this looks more app like.
3. Make sure the label is grayed out as well as the radio if it's disabled. `var(--slate)` is a good grayed out color.
4. If CSV is disabled, place the help text right under the label instead of in a tooltip. I think it's important enough to not hide behind hover. There is plenty of available space as well. Although it looks like that code path might be going away due to Noah's work in zqd.

I think these visual changes look good. Let me know if you feel strongly oppose any of them. 

Screen Shot 2020-12-23 at 12 55 20 PM

1. Make the title "Export Results"
2. Left align the form
3. Indent the radio buttons by 12px
4. Lowercase the options
5. Remove the footer background color (I think it only needs to be a darker color if there is some scrollable content above it.

jameskerr:
Hey Mason, here is another round of feedback. I thought the red text was too jarring for the warning, so I  brainstormed in sketch and ended up with a design for the modal that I think looks good. The link is at the bottom with the dimensions and spacing of everything.

1. Export complete pops up before it actually completes, looks the promise needs to be returned in `exportResults.ts`. 
2. There is no indication progress indicator in this version. I'd suggest putting the old one (the notice that looks like a pop notice but is different) back in for now until we have a solid replacement.
3. I think it would be good to use the default radio buttons for the OS that brim is being run on. They look great and feel "native". This can be done by simply following the semantic markup for radio buttons as shown here. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio. This will also allow the user to click on the text of the radio button to select it, rather than having to click on the small dot.
4. Here is a link to a design for the spacing and position of the items. https://www.sketch.com/s/4b2957a2-910e-403c-9fda-617de055c061/a/zx7EjKy#Inspector The disabled state should have the `cursor: disabled` style as well.

If you need more clarification on any of the items above, let me know.

jameskerr (approved):
This is great stuff Mason. Thanks for your patience during our back and forth discussions. I think the react toast library has a nice api and look forward to using it other places. In another ticket, I'll take a look at how I can theme it to be more consistent with the Brim App.

PR #1310, opened 2020-12-29
Log Detail Re-Style and Alert Correlations

### LINKED ISSUE #1250: Cross-ref between Suricata alerts and related Zeek context events
In the next Brim release, we'll have a field `community_id` present/populated in all Zeek events and Suricata alerts generated from pcaps that users import. These create a logical point to bi-directionally link these two data sources together.

For example, using the `all.pcap` test data, let's say I found my way to this Suricata alert.

image

Today, by right-clicking on the `community_id` value and selecting "New search with this value", my results show me both the Suricata alert and any other records that share that value, which might include other Suricata alerts as well as Zeek `conn` events that were all generated from the same flow.

image

However, we could improve on this by having the Log Detail view automatically show the records returned from a query `community_id` query. This would show both the Suricata+Zeek events together similar to what we already do for the "uid correlation" view in Zeek events.

Note that on the Zeek side, only the `conn` event has the `community_id` field. Therefore, to make this most useful we'd ideally want to include in the same visualization all the Zeek events that share the same `uid` as the Zeek `conn` event that had the matching `community_id`. For example, following on what I showed above, if I right-click on the `uid` field in that `conn` record, I see:

image

In other words, something I could have been shown in the first Log Detail view for the Suricata alert could effectively be a top-down waterfall of all these records:

image

As we discussed recently as a team, this "chaining" of queries is something we already do in the app, i.e. when a user clicks on an app-level Zeek event like `dns` or `http`, the app automatically triggers a query for the `uid` of that event to find the corresponding `conn` event, since if we find it we know we have the 5-tuple and time information to extract a flow from the pcap, and hence we activate the "Packets" button.

Whatever visualization we add here should be "bi-directional". That is, above I described clicking on a Suricata alert and seeing the waterfall that includes all the Zeek events, but we'd want to make sure if a user clicks on any Zeek event, the same type of triggered query shows them any corresponding Suricata alerts. This will be essential to allow users to quickly pivot in Log Details between the alerts and the context about why they fired.

Assuming we end up with a single united view that shows both Zeek+Suricata events in the same waterfall and it takes the place of what we today call "uid correlation", it seems a name change is in order. Perhaps just call it "Correlations"? I'm open to suggestions here (cc: @orochford).

### LINKED ISSUE #1292: "proto" field appearing to the left of the "alert" tile
A couple of us have noticed that in GA Brim `v0.21.0`, on views that show a mix of multiple record types, the `proto` field sometimes appears to the left of the `alert` tile. The ordering in the record definition in the ZNG is such that `alert` is still leftward of `proto`, so it seems this ordering is somehow being caused by the app.

image

jameskerr:
This PR closes #1250 and closes #1292 

### Color code the event_type in the viewer to be consistent with the detail pane. Also, changes color based on the severity field. 
image

### Adds related alerts and related connections to the detail pane and the detail window.

image

image

### Makes the size of the connection record to scale in the visuals.

image

### Fixes the ordering of the "event_type" field when the columns are in "auto" mode.

image

There are a bunch of line comments below to help clarify specific parts of the code.


app/core/hooks/useContentRect.ts
```diff
@@ -0,0 +1,21 @@
+import {useState, useLayoutEffect} from "react"
+
+export default function useContentRect(): [
```
jameskerr, line 3: A hook to get the size of an html element.

app/core/hooks/useResizeCallback.ts
```diff
@@ -0,0 +1,19 @@
+import {useState, useLayoutEffect} from "react"
+
+type Callback = (DOMRectReadOnly) => void
+
+export default function useResizeCallback(
```
jameskerr, line 5: A hook to run a function whenever an element is resized.

app/core/hooks/useSearch.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  useEffect(() => {
+    const {response, abort} = dispatch(search({query}))
+    response.chan(0, (records) => setRecords(records))
+    response.status((status) => setIsFetching(status === "FETCHING"))
+    return abort
+  }, deps)
+
+  return [records, isFetching]
+}
```
jameskerr, line 22: I think this will be the future of data fetching in the dashboard widgets. It's a simple hook that returns the records as they come in, and a boolean isFetching. 

```js
const [records, isFetching] = useSearch(* | count() by _path", [])
```

app/detail/Fields.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <PanelHeading>Fields</PanelHeading>
+      <DataPanel
+        fields={fields}
+        onRightClick={onRightClick}
+        onHover={onHover}
+      />
+      <Tooltip field={hovered} record={record} />
+    </section>
+  )
+})
```
jameskerr, line 106: The above code is for this part of the app: 
image

app/detail/Pane.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const {uid, cid} = new Correlation(record).getIds()
+  const isConn = isZeek && record.try("_path").toString() === "conn"
+  const hasMd5 = isZeek && record.has("md5")
+
+  return (
+    <BG className="detail-pane-content">
+      <div className="column">
+        <Fields record={record} />
+      </div>
+      <div className="column">
```
jameskerr, line 43: These column classes do nothing in the search window, but get styled in the detail window to provide that two column look.

image

app/detail/flows/contextMenu.ts
```diff
@@ -0,0 +1,22 @@
...
+  const columns = tableColumns.getColumns().map((c) => c.name)
+  const builder =
+    global.windowName === "detail"
+      ? detailFieldContextMenu
+      : searchFieldContextMenu
+
+  showContextMenu(builder(program, columns, space)(field, record, false))
+}
+
+export default contextMenu
```
jameskerr, line 22: This is the thunk that will be dispatched whenever any field is right clicked. It checks the global variable "windowName" to determine which type of menu to show, since the detail window and the search window provided similar but different options.

jameskerr: This is the pattern I'd like to continue for all right clicking. The usePopMenu hook is worse for performance.

package.json
```diff
@@ -109,6 +109,7 @@
     "chrono-node": "^1.4.8",
     "classnames": "^2.2.6",
     "d3": "^5.16.0",
+    "date-fns": "^2.16.1",
```
jameskerr, line 112: Adds a bunch of date functions. I think this should be used in favor of moment from now on. Since moment has been EOL'd.

ppl/detail/Correlation.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <Data>
+            <Name>Duration</Name>
+            <Value>{formatDur(conn?.getTime(), conn?.getEndTime())}</Value>
+          </Data>
+        </TableWrap>
+      </Panel>
+      <Caption>Populated by uid & community_id</Caption>
+    </section>
+  )
+})
```
jameskerr, line 54: This is the good old uid correlation panel.

ppl/detail/EventLimit.tsx
```diff
@@ -0,0 +1,29 @@
...
+  }, [query])
+
+  if (count < UID_CORRELATION_LIMIT) return null
+  return (
+    <Caption>
+      Limited to {UID_CORRELATION_LIMIT} events.
+      <Link onClick={onClick}>Query for All</Link>
+    </Caption>
+  )
+})
```
jameskerr, line 29: If there are more than 100 events returned for any correlation, we get this link where they can query for all of the events without the `| head 100` tacked on the end.

ppl/detail/EventTag.tsx
```diff
@@ -0,0 +1,12 @@
...
+import {ZeekEvent} from "./models/ZeekEvent"
+import SuricataTag from "./SuricataTag"
+import ZeekTag from "./ZeekTag"
+
+export default function EventTag({event, ...rest}) {
+  if (event instanceof ZeekEvent) return <ZeekTag event={event} {...rest} />
+  if (event instanceof SuricataEvent)
+    return <SuricataTag event={event} {...rest} />
+  return null
+}
```
jameskerr, line 12: These are the "path", "http", "alert" tags that get put on the event timelines.

ppl/detail/EventTimeline.tsx
```diff
@@ -0,0 +1,151 @@
...
+                style={getWidth(e)}
+                className={current === i ? "current" : undefined}
+              />
+            </div>
+          </Layer>
+        </Lane>
+      ))}
+    </div>
+  )
+})
```
jameskerr, line 151: This is the shared component for rendering a timeline of events. It's used in Related Alerts, Related Connections, and UID Correlation.

jameskerr: I expect it can be used in the dashboard if we wanted as well. Anytime we need to display timeline of events.

ppl/detail/RelatedAlerts.tsx
```diff
@@ -0,0 +1,63 @@
...
+              <Name>{name}</Name>
+              <Value>{value}</Value>
+            </Data>
+          ))}
+        </TableWrap>
+      </Panel>
+      <Caption>Populated by community_id</Caption>
+    </section>
+  )
+})
```
jameskerr, line 63: This is the component that renders this piece. It's only shown when viewing the details of a suricata record.

image

ppl/detail/RelatedConns.tsx
```diff
@@ -0,0 +1,57 @@
...
+              <Name>{name}</Name>
+              <Value>{value}</Value>
+            </Data>
+          ))}
+        </TableWrap>
+      </Panel>
+      <Caption>Populated by community_id</Caption>
+    </section>
+  )
+})
```
jameskerr, line 57: This component render's this piece.

image

ppl/detail/flows/fetch.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (!uid && !cid) return []
+  if (cid && uid) return run()
+  if (cid) return run()
+
+  // If there is only a uid and not a cid
+  const records = await run()
+  const conn = findConn(records)
+  if (conn && conn.has("community_id")) return dispatch(fetchCorrelation(conn))
+  else return records
+}
```
jameskerr, line 33: Going forward, I'd like to move the data fetching logic into the component itself. However, currently the LogDetails.uidLogs bit of state is used in the PacketsButton and in the UID Correlation view. It would be more maintainable to have each of those components issue their own queries to the backend to get what they need.

Alas, this is the logic that tries to find the uid related logs, and the community_id related logs.

ppl/detail/flows/getCorrelationQuery.ts
```diff
@@ -0,0 +1,21 @@
...
+  if (cid && uid) {
+    return connCorrelation(record)
+  } else if (uid) {
+    return uidCorrelation(uid)
+  } else if (cid) {
+    return cidCorrelation(cid)
+  } else {
+    return null
+  }
+}
```
jameskerr, line 21: I needed to split this out so that, if there are more than 100 events returned, we can offer a link that says, "Query for all". In which case it calls this function to get the query and put it in the main search input.

ppl/detail/models/ZeekEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+    return (this.r.get("ts") as zng.Primitive).toDate()
+  }
+
+  getEndTime() {
+    if (this.r.get("_path").toString() !== "conn") return null
+    const dur = (this.r.get("duration") as zng.Primitive).toFloat()
+    if (!dur) return
+    const ts = (this.r.get("ts") as zng.Primitive).toDate()
+    return new Date(ts.getTime() + dur * 1000)
+  }
```
jameskerr, line 21: This endTime is used only on the Zeek connection event. Now we can have the length of the connection be rendered to scale like so:

image

ppl/detail/util/formatDur.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+function firstTwo(array) {
+  const first = []
+  for (let i = 0; i < array.length; ++i) {
+    if (i === 2) break
+    first.push(array[i])
+  }
+  return first
+}
```
jameskerr, line 38: This function allows for nice, human readable durations. All the other date formatting functions in date-fns and momentjs were not human friendly enough for our case. 

Instead of choosing one unit, this function takes the two most significant units and displays them. For example:

```
2 years, 3 months
2 months, 5 days
1 day, 3 hours
2 hours, 59 minutes
1 minute, 3 seconds
and finally
54.2 seconds
0.34 seconds
```

src/css/_drag-anchor.scss
```diff
@@ -6,6 +6,7 @@
   background: transparent;
   pointer-events: all !important;
   cursor: col-resize;
+  z-index: 99;
```
jameskerr, line 9: Had a layering problem on the resize handle in the right pane.

src/css/settings/_colors.scss
```diff
@@ -95,4 +95,16 @@ $chrome-border: $white-5;
...
+  --ip: #4a90e2;
+  --port: #e3971c;
+  --interval: #544aa6;
+  --count: #e65835;
+  --bool: #3eaef4;
+
+  // alert severities
+  --alert-3: #fbc00e;
+  --alert-2: #f4912f;
+  --alert-1: #d0250b;
```
jameskerr, line 109: Got organized with where the type colors and alert severity colors live.

src/js/brim/time.ts
```diff
@@ -92,8 +92,8 @@ time.setDefaultFormat = function(format = "") {
     moment.defaultFormat = format
     moment.defaultFormatUtc = format
   } else {
-    moment.defaultFormat = "YYYY-MM-DDTHH:mm:ssZ"
-    moment.defaultFormatUtc = "YYYY-MM-DDTHH:mm:ss[Z]"
+    moment.defaultFormat = "YYYY-MM-DDTHH:mm:ss.SSS"
+    moment.defaultFormatUtc = "YYYY-MM-DDTHH:mm:ss.SSS"
```
jameskerr, line 96: Added milliseconds to the default date formatting. The data we tend to see often share the same second. Having milliseconds will help differentiate events.

src/js/components/App.tsx
```diff
@@ -38,6 +39,7 @@ export default function App() {
       <SpaceModal />
       <HTMLContextMenu />
       <PopNotice />
+      <BrimTooltip />
```
jameskerr, line 42: Made a shared tooltip for anywhere we just need a simple "data-tip".

src/js/components/ConnVersation.tsx
```diff
@@ -32,25 +28,23 @@ function filter(record: zng.Record, names: string[]) {
   return new zng.Record(cols, vals)
 }
 
-const ConnVersation = ({log, contextMenu}: Props) => {
```
jameskerr, line 35: Here and in subsequent components, I've moved the contextMenu builder into a thunk to eliminate extra renders. And renamed the var "log" to "record".

src/js/components/FieldCell.tsx
```diff
@@ -4,17 +4,31 @@ import classNames from "classnames"
 import {createCell} from "../brim/cell"
 import {zng} from "zealot"
 
-type Props = {field: zng.Field}
+type Props = {field: zng.Field; record: zng.Record}
 
-export default function FieldCell({field}: Props) {
+function getBackground(field, record) {
```
jameskerr, line 9: This function adds the `alert-1-bg-color` class to event_type=alert cells in the viewer.

jameskerr: It allows us to color the different alert severities with different colors: 
image

src/js/components/SearchResults/ResultsTable.tsx
```diff
@@ -82,14 +81,10 @@ export default function ResultsTable(props: Props) {
...
-      if (selection.isEmpty()) return
-      dispatch(viewLogDetail(logs[selection.currentRange[0]]))
-    },
-    400,
-    [selection]
-  )
+  useEffect(() => {
+    if (selection.isEmpty()) return
+    dispatch(viewLogDetail(logs[selection.currentRange[0]]))
+  }, [selection])
```
jameskerr, line 87: Removed the delay here. Things feel much snappier now and it doesn't look like there was a performance hit other places.

src/js/lib/columnOrder.ts
```diff
@@ -1,13 +1,13 @@
...
+  let orderedCols = cols.filter(({name}) => !EXCLUDED.includes(name))
   const ts = cols.find(({name}) => name === "ts")
-  if (ts) {
-    return [ts, ...orderedCols]
-  } else {
-    return orderedCols
-  }
+  const eventType = cols.find(({name}) => name === "event_type")
+
+  if (eventType) orderedCols = [eventType, ...orderedCols]
```
jameskerr, line 10: Moved event_type to the front of the columns after ts. Unfortunately, the way we handle "auto" columns is by creating 1 huge set of columns, then asking each record to display their fields in that order.

It seems like what we should do is, when in "auto" mode, simply render all the fields of each log in the order they were given from the server. This is a larger task than what was feasible here.

This is a little hack to move the event_path to the front just like we do with "_path" and "ts".

src/pkg/tree-list/hooks/useKeyBindings.ts
```diff
@@ -27,11 +29,11 @@ export function useKeyBindings(
       setSelector(selector.dup())
     }
 
-    const bindings = new Mousetrap(ref.current)
+    const bindings = new Mousetrap(node)
```
jameskerr, line 32: This was a bug where the first time the page loaded, these key bindings got registered globally instead of only local to their parent Dom node.

tsconfig.json
```diff
@@ -72,5 +72,5 @@
     "skipLibCheck": true /* Skip type checking of declaration files. */,
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
   },
-  "exclude": ["scripts", "test/api"]
+  "exclude": ["scripts", "test/api", "dist", "run"]
```
jameskerr, line 75: Excluding these really speeds up `npm run tsc -- --watch`

ppl/detail/ZeekTag.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {BrimEventInterface} from "./models/BrimEvent"
+
+const BG = styled.div`
+  height: 18px;
+  border-radius: 3px;
+  display: inline-flex;
+  color: white;
+  cursor: default;
+  user-select: none;
+  justify-content: flex-end;
```
mason-fish, line 13: Just gonna throw out that I personally prefer the justification to be centered here, looks a little neater that way imo

ppl/detail/Correlation.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <EventLimit query={query} count={events.length} />
+        </ChartWrap>
+        <TableWrap>
+          <Data>
+            <Name>Duration</Name>
+            <Value>{formatDur(conn?.getTime(), conn?.getEndTime())}</Value>
+          </Data>
+        </TableWrap>
+      </Panel>
+      <Caption>Populated by uid & community_id</Caption>
```
mason-fish, line 51: This should just be "Populated by uid", no?

app/detail/Fields.tsx
mason-fish: Might want to set a minimum width on this so it doesn't squish together in the detail window (I think the pane has this covered by enforcing it's own min-width as is)

mason-fish (commented):
This looks really great! I really like several of the patterns you're introducing here (esp. those new hooks and the context menu flow 👌 ). In addition to my code comments, I found a couple other small things while playing around with the branch.

This `conn` padding issue, mentioned offline:
image

and then it looks like the timestamp tooltip works fine in the detail pane, but not the detail window:
image

In the `Query for All` case (more than 100 events in correlation) case, and if we are in the detail window, we should bring the last search window into focus like we do for the context menu actions

ppl/detail/ZeekTag.tsx
jameskerr: You got it.

ppl/detail/Correlation.tsx
jameskerr: This one uses both, uid to get the Zeek records, and then if there is a conn in that first set, use it to get any associated alerts during that time:  Here's the function that gets those records: https://github.com/brimsec/brim/blob/3dadc9226574de87bfa8cef2251569061217588d/ppl/detail/flows/fetch.ts#L18

jameskerr:
I settled the min width issue, by setting a min width for the whole detail window of 500 px. Looks like this now at its smallest size: 
image

jameskerr:
@mason-fish thank you for finding that padding bug. Since it's a little more complicated to fix, I've created another ticket to track it with a screenshot and description. It does not seem like a show-stopping bug so we'll live with it for now and fix later: https://github.com/brimsec/brim/issues/1319

jameskerr:
@mason-fish thanks for your review! The path tags are now centered, the detail window has a min width, there's the issue for tracking the timeline padding, the tooltip works in the detail window, and the query for all function opens a new tab like the context menu actions.

🙏

mason-fish (approved):
Awesome work! Thanks for looking into those pieces!

PR #1328, opened 2021-01-05
remove csv guard

mason-fish:
now that https://github.com/brimsec/zq/issues/1271 closed, we don't need this!


jameskerr (approved):
Excellent use of the color red in this diff.

PR #1330, opened 2021-01-05
Queries Migration

jameskerr:
It appears the we had a few queries added to the state in the v0.21.1 release:

```js
> state.data.windows.cb42087bb5.state.queries.items
[
  {
    id: '1',
    name: 'Count by Path',
    zql: '* | count() by _path',
    description: '',
    tags: []
  },
  {
    id: '2',
    name: 'Suricata',
    zql: 'event_type=alert',
    description: '',
    tags: []
  },
  {
    id: '3',
    name: 'Long connections',
    zql: '_path=conn duration > 10',
    description: '',
    tags: []
  }
]

```

The awesome Query Library work that mason did was not being populated because some queries state already existed. 

_The queries section after migrating to master_
Screen Shot 2021-01-05 at 3 21 16 PM

Also, clicking on any of these did not work because one of the Query object properties changed. `query.zql` became `query.value`.

To address both of these problems, I've added a migration that resets the queries state to be undefined. That way, the reducer will use the initial value, containing all the cool queries we've prepared.

_The queries section after migrating to this branch_
image


mason-fish (approved):
Thanks for catching this!

PR #1334, opened 2021-01-06
System Test Hooks

### LINKED ISSUE #1274: change how tests detect import completion
In a slack discussion about some flakiness we've seen with new suricata verification tests, @jameskerr said:

> Some more info, the pop notice is displayed for 4 seconds, then it disappears. Checking for the presence of it seems to be working well enough to determine if an import is complete, but I’ve always wondered if sometimes spectron waits too long to start checking for the existence of it. If that happened the test would fail with “couldn’t find the pop notice”. Anyway, there’s a little upload of thoughts I’ve had in the past.
> alfred  2 hours ago
Is there a different way the tests could check for pcap import being done that doesn't rely on catching the popup window? Some variable on the DOM, ... ?
> Kerr  1 hour ago
Yeah, that’s an idea. Set a persistent attribute on the dom.

jameskerr:
Closes #1274 

If the env var BRIM_ITEST="true" is set, then a small log will appear in the right side of the app. 

image

I'm calling these messages, "system test hooks". They are backed by a piece of the redux store called "systemTest" which contains an array of stings (hooks).

To add a hook to the log, simply `dispatch(SystemTest.hook("import-complete")` somewhere in the app code. Then in the system test call `waitForHook(app, "import-complete")`

This should help us identify states of the app more reliably when it is hard to do so otherwise.


jameskerr:
When there's a bunch of events:

image

mason-fish (approved):
Love it!

PR #1338, opened 2021-01-07
Reset state

### LINKED ISSUE #740: Wire up "Reset State"
We now have [an article](https://github.com/brimsec/brim/wiki/Troubleshooting#brim-seems-unable-to-restart-normally-such-as-after-a-bad-crash) that guides users through manually deleting `appState.json` to recover from a bad crash. Per a comment from @jameskerr that was captured in #697, we can probably do even better if we actually wire up the "Reset State" pull-down option in Electron to do a similar kind of clearing, so that way the user doesn't need to drop to the filesystem to get unstuck.

jameskerr:
Closes #740 

The "Reset State" button is now wired up to do just that. It will:

1. Ask the user if they are sure and inform them of what will happen
2. Delete the appState.json file
2. Relaunch the app without persisting state

image

image

In order to get this code path under a unit test, I've moved more code into the `Brim` class. A `new Brim()` is now composed of 4 objects. 

1. windows (the window manager)
2. session (the tron.session)
3. store (the redux global store)
4. zqd (the zqd manager)

With those 4 items, you can do just about anything you need to within "Brim". For this primary fix, the final code was simply, `brim.resetState()`. Instead of needing to pass 2,3 or 4 args into some of our handler functions, we simply pass in the `brim` instance. 

The brim instance is now much easier to unit test, thanks to the fact that all 4 of those dependencies can be injected with mocks.


mason-fish (approved):
Awesome, glad we have this wired back up. Nice to have an escape hatch!

PR #1339, opened 2021-01-08
Column order:  ts, _path, event_type

### LINKED ISSUE #1292: "proto" field appearing to the left of the "alert" tile

jameskerr:
Actually fixes #1292 

Eventually, I'd like to display the events in the same order they come from the sever, then have a per-schema user preference to move the ordering around in the app. For now, we'll just continue with the less-optimal solution.

The logic is:

1. Gather all the unique columns from the schemas present in the results into an array
2. Grab the event_type column and force it to the front of the array
3. Do the same with the _path column
4. And finally do the same with the ts column

Unit tests have been added that show this is working now.

image

### DIFF
--- a/src/js/lib/columnOrder.ts
+++ b/src/js/lib/columnOrder.ts
@@ -1,13 +1,15 @@
 import {$Column} from "../state/Columns/models/column"
 
-const EXCLUDED = ["ts", "_td", "event_type"]
+const EXCLUDED = ["ts", "event_type", "_path"]
 
 export default (cols: $Column[]) => {
   let orderedCols = cols.filter(({name}) => !EXCLUDED.includes(name))
   const ts = cols.find(({name}) => name === "ts")
   const eventType = cols.find(({name}) => name === "event_type")
+  const path = cols.find(({name}) => name === "_path")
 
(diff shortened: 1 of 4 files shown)


src/js/lib/columnOrder.ts
```diff
@@ -1,13 +1,15 @@
...
+const EXCLUDED = ["ts", "event_type", "_path"]
 
 export default (cols: $Column[]) => {
   let orderedCols = cols.filter(({name}) => !EXCLUDED.includes(name))
   const ts = cols.find(({name}) => name === "ts")
   const eventType = cols.find(({name}) => name === "event_type")
+  const path = cols.find(({name}) => name === "_path")
 
   if (eventType) orderedCols = [eventType, ...orderedCols]
+  if (path) orderedCols = [path, ...orderedCols]
```
mason-fish, line 12: 👍

mason-fish (approved):

PR #1343, opened 2021-01-08
The proper appRoot in dev mode

jameskerr:
Since we changed the directory structure, the user tasks on windows were broken in dev mode. However they worked just fine on release builds.

### DIFF
--- a/src/js/electron/userTasks.ts
+++ b/src/js/electron/userTasks.ts
@@ -26,7 +26,7 @@ export default function(app: any) {
 
 function getArguments(arg) {
   if (electronIsDev) {
-    // This is not exactly relyable. Each time the directory changes, this will change.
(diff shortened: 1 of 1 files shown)


mason-fish (approved):

src/js/electron/userTasks.ts
```diff
@@ -26,7 +26,9 @@ export default function(app: any) {
 
 function getArguments(arg) {
   if (electronIsDev) {
-    const appRoot = path.join(__dirname, "..", "..", "..")
+    // This is not exactly relyable. Each time the directory changes, this will change.
```
jameskerr, line 29:
```suggestion
    // This is not exactly reliable. Each time the directory changes, this will change.
```

Issue #1347, opened 2021-01-09
Infinite scroll appears to be broken


philrz:
Repro is with Brim commit `f6c7ba3`.

The attached video starts after test pcap https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz (uncompressed) has been imported. As shown, it contains a count of over 670k+ records.

Staying on the main events panel, the default search is in effect such that I expect that I'm being shown the first of all 670k+ records. At the default window size shown here, there's 20 records visible at a time in this main events panel. As a user, when clicking the "blank space" above or below the draggable control in a scrollbar, it's my expectation that I'll be shown the next "screen's worth" of records (the "next 20" in this case). However, I manage to do this click only 27 times before I'm shown "end of results", giving me the impression I've seen every record. Clearly that's way fewer clicks than you'd expect to pan through 670+ records.

https://user-images.githubusercontent.com/5934157/104105775-61263300-5265-11eb-8b2b-e98024778170.mp4

I know from experience that we only ask the backend for 500 records at a time. In the past there was "infinite scroll" functionality when we reached the bottom of the screen that would trigger the pull of the next 500 records. From the symptoms shown here, I get the impression that may have stopped working.

philrz:
Verified in Brim commit `ce4d71f`.

The attached video picks up just short of the "27 clicks" described in the original repro. Now when we click such that the scroll control gets to the very bottom, the scroll control jumps up as an indication that another batch of records has been fetched from `zqd`. Another set of clicks brings the scroll control to the bottom again and the cycle repeats. The same is true if the scroll control is clicked and dragged immediately to the bottom.

https://user-images.githubusercontent.com/5934157/104771244-d773db00-5726-11eb-99a8-cec9480ade90.mp4

Infinite scroll is back! Thanks @jameskerr!

PR #1348, opened 2021-01-11
Infinite fix

jameskerr:
Fixes: #1347

Thanks for catching this Phil. Geez, I think was with us since the big zjson update 3 months ago: https://github.com/brimsec/brim/pull/1108. The infinite scroll was mistakenly unfinished when the PR went in. Automated testing for this is a bit unclear to me at the moment. We could execute javascript within the spectron driver and set the scrollTop of the element to the max. Then wait until the fetch next page hook fires, then check the contents of the viewer. I want to write these tests now, but I'm feeling pulled to work on the security dashboard.

Thankfully, one of the bug was something typescript could have caught. I've restructured the SearchResponse class in away that typescript understands. After doing that, the compiler revealed the bug:

Screen Shot 2021-01-11 at 10 10 38 AM

I think some better unit tests might be easier.

### DIFF
--- a/src/js/flows/search/handler.ts
+++ b/src/js/flows/search/handler.ts
@@ -1,4 +1,4 @@
-import {createResponse} from "./response"
+import {SearchResponse} from "./response"
 import whenIdle from "../../lib/whenIdle"
 
 function abortError(e) {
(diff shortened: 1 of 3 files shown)


mason-fish (approved):
lgtm! Your mention of typescript reminds me that we should consider adding in some more transpiler options to encourage us to type things more explicitly

PR #1349, opened 2021-01-11
Fix pin bugs

### LINKED ISSUE #1166: Inconsistent pins in tree view
I noticed this while running through the common steps we use to demo Brim using the `all.pcap` data. Repro is with Brim commit `ae6ef08`.

As shown in the attached video, the following steps are used to repro:

1. Enter search term `192.168.0.51`, then pin
2. Add search term `_path=http`, then pin
3. Add search term `GET` and execute (_not_ pinned)
4. Click on the pinned search term `192.168.0.51` and change it to `192.168.0.53` to create a new branch in the tree view

[Repro.mov.zip](https://github.com/brimsec/brim/files/5440588/Repro.mov.zip)

At this point we see two branches in the tree view that have the same hierarchy, since each consists of two separately-pinned search terms followed by a non-pinned one. However, if I click on the `_path=http` in each tree branch, they are presented differently in my search bar & the pinned terms shown below the search bar (one brings in the `GET`, one does not).

I'm honestly not even sure which of those two behaviors is the more desirable one, I just sense they should be consistent. In fact, thinking about it some more, both seem kind of wrong based on what I just clicked on. Since in both cases I'm clicking on a pinned search term that includes another pin below it, I think maybe this is what would have been most appropriate:

image

...but I'll defer to the UX guys, for sure.

mason-fish:
fixes #1166 

There were a few inconsistencies in the tree view, and this PR aims to fix them with these changes:

* When clicking on any part of the investigation tree, the value you click on will go to the search bar and any of it's parental lineage will be be used as pins. Previously the underlying query was being used, even if you had only intended to use a fragment of that original query.
* When deleting an entry from the tree, we use the underlying values as well, but now we inform the user about that behavior.
* Replaces our own Tree class with the `tree-model` package we are using in the query lib reducer.


jameskerr:
This behavior is much more strait forward now. Thanks for swapping in the tree-model too. 

I noticed that we provide an "are you sure" popup now in the tree view, but we don't in the linear view. Was that your intent? I think it might be cumbersome to click through that each time I wanted to delete something from the tree. What do you think about changing the label from "Delete" to "Delete underlying entry"? That might provide enough of a warning to the user without making them click through a popup. It would also match the linear view. What do you think?

src/js/components/FilterTree.tsx
```diff
@@ -16,22 +15,80 @@ import Search from "../state/Search"
...
+
+const getPins = (node?: InvestigationNode): string[] => {
+  const result = reduce(
+    node?.getPath(),
+    (res, n) => {
+      res.push(n.model.filter)
+      return res
+    },
+    []
+  )
```
jameskerr, line 28: I think might be simpler as a map function? `node.getPath().map(n => n.model.filter)`

src/js/components/FilterTree.tsx
```diff
@@ -16,22 +15,80 @@ import Search from "../state/Search"
...
+          .showMessageBox({
+            type: "warning",
+            title: "Delete History Entry",
+            message: `Are you sure you want to remove this entry and it's underlying query?`,
+            buttons: ["OK", "Cancel"]
+          })
+          .then(({response}) => {
+            if (response === 0) {
+              const multiTs = node
+                .all(() => true)
```
jameskerr, line 85: Is this the same as saying `node.children`?

mason-fish: As mentioned offline: no, `all()` is similar to a `descendants()` or function which recursively returns all children, grand-children, etc. It just takes a predicate function which I'm not really utilizing any sort of filter here, hence `() => true` (return all nodes)

mason-fish:
@jameskerr I agree the dialog was a bit too much, here is the new behavior:
image

jameskerr (approved):
Thanks! Looks good

PR #1352, opened 2021-01-13
ensure contextMenu itest uses sorted data

alfred-landrum:
When using archive store by default, this test failed as it the data file it uses as the same timestamp for all its records.


mason-fish (approved):
Oh nice catch!

PR #1358, opened 2021-01-15
Enable creation of ad hoc builds (on-demand and nightly scheduled)

philrz:
## tl;dr

This PR allows for creation of "ad hoc builds" based on the current state of Brim `master`. These are signed and made available for all supported platforms just like our regular GA builds.

## Triggers for build creation

There's two ways these are currently triggered:

1. On-demand, by clicking the **Run workflow** pull-down on Actions (aka "workflow_dispatch" mode)

image

2. Automatically every night at 12:05 AM Pacific time (aka "schedule" mode)

## Build availability

Once the builds are created and available, a message is sent to a **#releases** channel on Slack where users are notified of the availability. Link text offered is of the format `gs://` to reflect that they're on a Google Cloud Storage bucket, but they're hyperlinked to point to the `https://` path which publicly-expose buckets.

image

I've tested this all out on a personal fork of the Brim repo at https://github.com/philrz/brim, where I changed the necessary secret to point to a temporary public bucket `gs://phil-releases`. If this PR is approved, I'll plan to open up our other "release" bucket publicly as well, since its contents do not seem to be sensitive, and it should be convenient for us to download these directly via the Slack hyperlinks and also offer them to community users in Support contexts as needed, e.g. to test bug fixes or get early feedback on new features. If we should ever want to, we could extend the workflow to also publish the nightly build links on the community Slack system.

# Functional description

A part of the functionality I've carefully tested is the handling of version strings, specifically this portion:

```
ADHOC_VERSION_STRING=$(npm version preminor --no-git-tag-version --preid master-${GITHUB_SHA:0:8})
```

This has the effect of advancing the base version number to the next minor one, and then appending an additional string that includes a reference to the `master` commit. I rigged it up this way so the user will always see a meaningful version string they can share with us in Support contexts (the commit being the most important part in this case) while the base version string tries to best accommodate expected auto-update behaviors. Here's some specific examples that illustrate this.

In the first case, the `version` string in `package.json` and `package-lock.json` is set to `0.21.0`. An ad hoc build was created when the tip of `master` was commit `676b15ee`, so the above `npm` command then has the effect of advancing the version string to `0.22.0-master-676b15ee.0`. Right now the newest GA Brim release has version string `0.22.0`. Therefore, when running this particular ad hoc release, the auto-update _does_ initiate the download of the GA `0.22.0` release.

Brim-v0 22 0-master-676b15ee 0

In the next case, the `version` string in `package.json` and `package-lock.json` is set to `0.22.0`, which happens to be the current state of official Brim `master` right now. An ad hoc build was created when the tip of `master` was at commit `7657b80a`, so the above `npm` command then has the effect of advancing the version string to `0.23.0-master-7657b80a.0`. Since the newest GA Brim release of `0.22.0` is "older" than that base number, I was able to have that ad hoc version open and running on my desktop indefinitely _without_ any auto-update firing. I think this is key behavior that we want to have. Let's say that we'd given this release to a community user to fix a bug or be an early tester/user of upcoming new functionality. We'd want them to be able to keep running this ad hoc build without interruption until the next minor GA release is published (`0.23.0` in this example) which would be expected to deliver the "official" bug fix or enhancement they've already been using.

Brim-v0 23 0-master-7657b80a 0

Finally, the previous functionality for handling "release candidates" is unaffected. Branch names starting with `rc-v*` still have significance in that they trigger the creation of a build while preserving whatever `version` string is in `package.json` and `package-lock.json`. The creation of a "Release" on GitHub is still what triggers the "promotion" of such RC builds to the Releases page. These RC builds are still announced via the Slack channel in the same way, though, which should be convenient for those who want to participate in smoke testing as GA releases are being finalized.

Brim-rc-v0 23 0

image

### DIFF
--- a/.github/workflows/linux-release-candidate.yml
+++ b/.github/workflows/linux-release-candidate.yml
@@ -4,12 +4,18 @@ on:
   push:
     branches:
       - rc-v*
+  schedule:
+    - cron: '5 8 * * *'
+  workflow_dispatch:
 
 jobs:
   build:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
+    - name: Extract branch name
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      id: extract_branch
     - name: install rpm
       run: sudo apt-get install -y rpm
     - uses: actions/setup-go@v2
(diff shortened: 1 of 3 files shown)


jameskerr (approved):
Cool, Phil. Makes sense to me. Thanks for your explanation.

PR #1366, opened 2021-01-20
move query lib to global state

### LINKED ISSUE #1365: Query Library purged by exiting Brim via "X" on Linux
This issue was originally [reported by a community user](https://brimsec.slack.com/archives/C010DR0HHMF/p1611073606017600?thread_ts=1610799195.005500&cid=C010DR0HHMF) on Solus Linux, though here I've reproduced it on Ubuntu 18.04. Repro is with GA Brim tagged `v0.22.0`.

As shown in the attached video, I can create a saved query, quit the app via **File > Exit**, and when I relaunch the app I'm immediately dropped into the same Space and my saved query is present. However, if I quit via the "X" in the upper-right corner of the window, when I relaunch I'm _not_ placed into any Space, and when I click the Space, I find my saved query is gone.

https://user-images.githubusercontent.com/5934157/105197430-76b31c80-5af1-11eb-84c5-610d8942250f.mp4

mason-fish:
fixes #1365 

Moves query lib to global state instead of per window state.


src/js/state/migrations/202101201109_moveQueriesStateToGlobal.test.ts
```diff
@@ -0,0 +1,29 @@
+import getTestState from "../../test/helpers/getTestState"
+import migrate from "./202101201109_moveQueriesStateToGlobal"
+import init from "../../../../ppl/queries/initial"
```
jameskerr, line 3: Since we have to keep these migration files around a long time, I think it's best not to depend on code that we might want to change in the future. If we change the init function, this migration might fail. So I think it's best to copy past the whole expected queries array right into this test case.

jameskerr (changes requested):
Thanks for getting to the bottom of this. Looks good to me. One request on the test.

src/js/state/migrations/202101201109_moveQueriesStateToGlobal.test.ts
mason-fish: Oh, good point. Yes I'll change.

jameskerr (approved):
I just tested this out. I checked out master and opened two windows. In the first I saved a query called "A" and in the second I saved a query called "B". Each window also had the default queries. Since the queries state in master is localized to each window, the new queries did not populate to the other windows. I then quit the app which saved my state.

I then checked out this branch and started the app again. Both of my previous windows were opened and the queries A and B were in both windows as we expected. Adding a third query C added a new entry to both windows.

PR #1367, opened 2021-01-20
Add right click "copy" to search and detail fields

### LINKED ISSUE #1304: Right-click to copy to paste buffer
A community user asked:

> ...if you [...] can just add highlight right click copy, I'll be really happy.
> It's so basic you'll laugh. In Word, when you highlight something and right click, you have the option to copy what is highlighted to the clipboard. Brim has a lot of options when I right click, but not that basic one

This is a fair point! We'll plan to add this. 😄

jameskerr:
Fixes #1304 

A simple but missing feature. 

image

image

image

Also added a "Copy" menu option to all the values in the detail pane.

image

There may be a better way to set a single context menu handler for all generic text, but this will do for now.

### DIFF
--- a/app/core/Data.tsx
+++ b/app/core/Data.tsx
@@ -35,16 +35,18 @@ export const Data = styled.dl`
 
 export const Name = styled.dt`
   margin: 0;
-  margin-right: auto;
   white-space: nowrap;
 `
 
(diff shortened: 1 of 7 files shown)


mason-fish (approved):
Right on 👍

PR #1368, opened 2021-01-20
"Restart Required" text fix

### LINKED ISSUE #1364: "Restart required" message is obscured when entering runner paths
Repro is with a Brim build based on commit `50788557`.

As shown in the attached video, when entering values for the Suricata Runner or Zeek Runner, the "Restart required" message is partially obscured. If I click away to another field it becomes visible, but considering the importance of the message, we should probably make sure it's fully visible.

https://user-images.githubusercontent.com/5934157/104980599-cabbe500-59bb-11eb-887c-a2dc119ecd20.mov

jameskerr:
fixes #1364

The restart required text was previously obscured. It now is a little smaller, positioned better, and always above it's file input.

image

### DIFF
--- a/src/css/_settings-modal.scss
+++ b/src/css/_settings-modal.scss
@@ -52,13 +52,16 @@
 
   .feedback {
     position: absolute;
-    right: 24px;
-    top: -11px;
+    right: 12px;
+    top: 1px;
+    margin: 0;
     @include label-small;
+    line-height: 14px;
     background: $blue;
     color: white;
     padding: 0 6px;
-    border-radius: 3px 3px 0 0;
+    border-radius: 4px;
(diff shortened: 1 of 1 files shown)


src/css/_settings-modal.scss
```diff
@@ -52,13 +52,16 @@
...
+    margin: 0;
     @include label-small;
+    line-height: 14px;
     background: $blue;
     color: white;
     padding: 0 6px;
-    border-radius: 3px 3px 0 0;
+    border-radius: 4px;
     animation: fadein 300ms;
+    z-index: 1;
```
mason-fish, line 65: 👌

mason-fish (approved):

PR #1372, opened 2021-01-21
Rename workspace

mason-fish:
renames cluster/connection to `workspace`! Also removes `password` and `username` from the new workspace model in prep for some incoming auth related changes


src/js/state/migrations/202101210823_renameToWorkspace.test.ts
```diff
@@ -0,0 +1,25 @@
+import getTestState, {getAllStates} from "../../test/helpers/getTestState"
+import migrate from "./202101210823_renameToWorkspace"
+
+test("migrating 202101210823_renameToWorkspace", () => {
+  let {data} = getTestState("v0.21.1")
+
+  let next = migrate(data)
+
+  if (next) {
```
jameskerr, line 9: Need this if here?

src/js/state/migrations/202101210823_renameToWorkspace.ts
```diff
@@ -0,0 +1,29 @@
...
+    delete state.globalState.clusters
+  }
+  for (const s of getAllStates(state)) {
+    delete s.clusters
+  }
+
+  // window: current.connectionId -> current.workspaceId
+  for (const s of getAllStates(state)) {
+    if (!s.tabs) continue
+    for (const tab of s.tabs.data) {
```
jameskerr, line 22: Just so you're aware in the future, I created a `getAllTabs(state)` function as well.

jameskerr (approved):
Awesome! Much more clear. I did a git grep for the word cluster and looks like there is only one more method which uses the word.

```
src/js/state/Tab/selectors.ts:const clusterUrl = createSelector<State, Workspace | null, string>(
```

Happy to merge this as is though.

src/js/state/migrations/202101210823_renameToWorkspace.test.ts
```diff
@@ -0,0 +1,25 @@
+import getTestState, {getAllStates} from "../../test/helpers/getTestState"
+import migrate from "./202101210823_renameToWorkspace"
+
+test("migrating 202101210823_renameToWorkspace", () => {
```
jameskerr, line 4: I think it would be good to add `expect.assertions(12)` (or whatever the number of expect calls we "expect"). Putting the expect calls in a loop is a little dangerous because if the array is empty, the expect calls won't get run, and the test will still pass.

PR #1373, opened 2021-01-21
Boards State with Redux/Toolkit

jameskerr:
Boards and Tiles will be the two new slices of state in the app. This PR adds the CRUD actions and reducer cases for creating a board, removing a board, getting all boards, getting one board, and appending a tile to a board.

It adds the redux/toolkit package and makes use of its functions. This is a nice guide to the functions and the api. https://redux-toolkit.js.org/usage/usage-guide

No UI is in the PR

Related to #1058

We will start seeing warnings in our development environment whenever we save something to the store that is not serializable and if we mutate the store outside of a dispatch.


src/js/initializers/initStore.ts
```diff
@@ -19,16 +16,14 @@ function getInitialState(windowId) {
...
-    composeWithDevTools(
-      applyMiddleware(
-        reduxThunk.withExtraArgument({
-          globalDispatch,
-          createZealot
-        })
-      )
-    )
-  )
+  return configureStore({
```
jameskerr, line 19: Docs for configure store: https://redux-toolkit.js.org/usage/usage-guide#simplifying-store-setup-with-configurestore

src/js/state/Boards/index.ts
```diff
@@ -0,0 +1,30 @@
+import {createEntityAdapter, createSlice, PayloadAction} from "@reduxjs/toolkit"
+import {State} from "../types"
+
+export type Board = {id: string; title: string; tiles: string[]}
+export type BoardsState = ReturnType<typeof slice.reducer>
+
+const adapter = createEntityAdapter<Board>()
```
jameskerr, line 7: It was useful for me to first understand [Create Action](https://redux-toolkit.js.org/api/createAction), [Create Reducer](https://redux-toolkit.js.org/api/createReducer), and [Create Slice](https://redux-toolkit.js.org/api/createSlice) as they each build on one another.

Docs for create entity adapter: https://redux-toolkit.js.org/api/createEntityAdapter

src/js/state/Boards/index.ts
```diff
@@ -0,0 +1,30 @@
...
+const selectors = adapter.getSelectors((state: State) => state.boards)
+const slice = createSlice({
+  name: "boards",
+  initialState: adapter.getInitialState(),
```
jameskerr, line 11: The "EntitiyState<Board>" type looks like this:

```
{
  ids: ["a", "b," "c"]
  entities: { "a": Board, "b": Board, "c": Board}
  // Board being a board object.
}
```

mason-fish (approved):
Nice! It looks like with this toolkit keeping all the of a store subtree/slice (types, actions, reducers, etc.) in one file is actually quite doable, though I imagine we will still want to keep the larger ones separated by file. 👍

PR #1380, opened 2021-01-24
Improve docs for pcap open failures

philrz:
A community user on public Slack [reported a problem](https://brimsec.slack.com/archives/C010DR0HHMF/p1611371926002900) with the **Packets** button not working as expected on Debian 10. While they figured out the problem themselves, they made the helpful suggestion that we could use their experience to improve the Troubleshooting article in the wiki. I've drafted those improvements here.

Since we now bundle the `pcap` tool with Brim (we didn't in the past) I was able to improve that part of the doc as well to save users the hassle of separately downloading the `zq` toolset.


jameskerr (approved):
Reads well to me.

PR #1383, opened 2021-01-25
Upgrade Migration Testing

jameskerr:
While reviewing the diff between the last release and master, I found a migration test that should have failed but was passing. The reason was because this particular migration was dependent on the previous migration and tests only exercise a single migration function at a time, so they did not catch the bug. 

In reality, a "batch" of migrations get applied at a time. So I've added a helper function in the migration tests.

```js
const nextState = await migrate({state: "v0.22.0", to: "20210101000"})
```

This function will read the test state file called "v0.22.0.json", look at the version attached that that state, then run all migrations between the version and the "to:" parameter, returning the migrated state.

This is much closer to the way migrations are run in production and should add to our migration confidence.

I updated all the tests from version v0.7.0 onward. Tests older than that rely on a different versioning system and it seems not worth the cost of accommodating it.

### DIFF
--- a/src/js/state/migrations/202008191031_changeZqdClusterIdToHostPort.test.ts
+++ b/src/js/state/migrations/202008191031_changeZqdClusterIdToHostPort.test.ts
@@ -1,12 +1,9 @@
-import getTestState from "../../test/helpers/getTestState"
-import migrate from "./202008191031_changeZqdClusterIdToHostPort"
+import {getAllStates, getAllTabs} from "src/js/test/helpers/getTestState"
+import {migrate} from "src/js/test/helpers/migrate"
 
-test("migrating 202008191031_changeZqdClusterIdToHostPort", () => {
-  const {data} = getTestState("v0.15.1")
-
-  const next = migrate(data)
-
-  const windows = Object.values(next.windows)
+test("migrating 202008191031_changeZqdClusterIdToHostPort", async () => {
+  const next = await migrate({state: "v0.15.1", to: "202008191031"})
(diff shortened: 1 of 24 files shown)


src/js/state/migrations/202008191031_changeZqdClusterIdToHostPort.test.ts
```diff
@@ -17,16 +14,13 @@ test("migrating 202008191031_changeZqdClusterIdToHostPort", () => {
...
     expect(state.clusters[newId]).toEqual(newValue)
     expect(Object.keys(state.clusters)).not.toContain(oldId)
   }
 
-  // @ts-ignore
-  const tabsData = windows.flatMap((win) => win.state.tabs.data)
-
-  for (const td of tabsData) {
-    if (!td.current) continue
-    expect(td.current.workspaceId).toBe(newId)
```
jameskerr, line 30: @mason-fish this was what tipped me off to the need for this PR. I'm guessing the migration file was updated as the result of a  find/replace of clusterId to workspaceId. Since this is an old migration file, I expected this test to fail. But it was never getting to that expect call because of the if clause above it. This state has "search.clusterId" instead of "current.clusterId". It was not being caught because the migration before this moves clusterId to the current slice.

mason-fish: I see, good catch!

mason-fish (approved):
Makes sense to me 👍

PR #1387, opened 2021-01-26
Account for No Windows when Migrating Queries

### LINKED ISSUE #1386: Query Library lost upon upgrade due to migration issue
This bug was found during smoke testing of a `v0.23.0` release candidate based on Brim commit `1404bdf`. The repro is shown in the attached video. @jameskerr has already been able to explain the problem:

> If you quit v0.22.0 with at least one window open, the query library is fine again in v0.23.0. But if you close all your windows in v0.22.0, then open v0.23.0, the queries will be lost.
>
> It’s a migration issue. When there are no windows to restore, the migration thinks that there are no queries, when it should in fact restore the initial state.

https://user-images.githubusercontent.com/5934157/105781551-a0bc7280-5f27-11eb-88fc-421f3d34d1f9.mp4

jameskerr:
fixes #1386

This edit's an existing migration. Since migrations don't go backwards, you'll need to clear you app state, open v0.22.0, then save the state with no windows, then upgrade to v0.23.0 and see that the query library is populated.

### DIFF
--- a/src/js/electron/tron/migrations.ts
+++ b/src/js/electron/tron/migrations.ts
@@ -42,7 +42,7 @@ export default async function migrations(
 
     run(state: VersionedData, migrations: Migration[]) {
       for (const {version, migrate} of migrations) {
-        state.data = migrate(state.data)
+        state.data = state.data ? migrate(state.data) : undefined
         state.version = version
       }
       return state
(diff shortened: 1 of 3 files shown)


mason-fish (approved):
Thanks for catching this!

philrz (approved):
I created a local `.dmg` installer and repeated the repro steps from #1386 and the Query Library was still there after relaunch. 👍

PR #1388, opened 2021-01-26
Auth0 integration

mason-fish:
This is the bulk of the auth0 integration! Tests are still to come, but this is ready for review. 

This includes an auth0 client which, ultimately, I ended up writing our own after finding that the SDKs that auth0 offered were not focused so much on storing multiple credentials in a single client, and generally focused more on browser based clients rather than electron. It also wasn't that much work and is pretty straight forward.

Speaking of credentials, we persist those in the default OS's secret storage (i.e. for Mac, it is Keychain). We only pull out the refresh token when the access token doesn't exist, is expired or malformed, or fails in any other way as it's used in requests. We pull the access token from the secrets storage and place it in the global state the first time we need to use it, and then continue to reference it there (in this case, the global state acts as a cache to the secrets storage), but upon app close we do not persist the token. This way our tokens are only persisted in the OS's native secure storage.

I also refactored a bit of our workspace initiation to accommodate these changes. I had originally extended the `initWorkspace` flow to handle all of this but I found that there were several issues about notifying it's callers (esp the workspace form, but also the new login page(see below)) of different success and failure states. And there are a lot of those states. For that reason I also found that it was just unwieldy to put everything in there, so now it is broken up into 3 main steps:

* `buildWorkspace`: Should only be done on creation or edit of workspace: it builds the most basic, required pieces of a workspace that we cannot get from the form (version, authType and authData if necessary).
* `getAuthCredential`: Attempts to grab access token from it's potential locations, and will grab refresh token to try and generate a new access token if necessary. For new connections we expect this to return empty handed.
* `login`: Redirects the user to browser, sets up an ipc listener for handling the re-re-direct (return from browser back to Brim) and accepts a callback parameter so the caller can decide how to handle the login return behavior. I had thought to promisify this at first, but since we want both a callback AND a cancel button I thought this pattern was the simplest.

Below are some screenshots of the different user flows:

New workspace form, upon hitting "connect", if the designated connection requires auth, will inform user:
image

While user is redirected from form, fields and primary button are disabled but cancel is available:
image

Login page:
image

If login failed:
image

"Login" page, for existing connections which fail to refresh their access tokens:
image

When user is redirected to browser, show spinner and cancel button:
image

refresh token and access token in keychain
image


zealot/fetcher/fetcher.ts
```diff
@@ -11,28 +11,30 @@ export type FetchArgs = {
...
+      const {path, method, body, signal, headers} = args
+      const resp = await fetch(url(host, path), {method, body, signal, headers})
       if (!resp.ok) {
         const content = await parseContentType(resp)
         return Promise.reject(createError(content))
       }
       const iterator = createIterator(resp, args)
       return createStream(iterator, resp)
     },
+    // TODO: send auth in xml upload, when present
```
jameskerr, line 37: Does this mean that importing a log file won't be authenticated atm?

src/js/components/TabContent.tsx
```diff
@@ -45,6 +46,7 @@ export default function TabContent() {
     )
 
   if (wsStatus === "disconnected") return <ConnectionError workspace={ws} />
+  if (wsStatus === "login-required") return <Login ws={ws.serialize()} />
```
jameskerr, line 49: Why serialize the workspace here?

jameskerr (changes requested):
Thanks Mason! The app feels a lot more grown up with all this functionality in it. I left quite a few comments. Feel free to take the ones that make sense to you. All this new logic would be grateful for some tests as you mentioned you planned on doing. Let me know if you'd like any help stubbing things or structuring tests.

Thanks again for this work. Awethsome stuff :)

zealot/fetcher/fetcher.ts
mason-fish: Shoot, yes, I didn't immediately know where to include headers and meant to circle back on this. I'll add this in, thanks for catching

src/js/flows/workspace/initWorkspace.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const initWorkspace = (ws: Workspace, status: WorkspaceStatus) => (
+  dispatch
+): void => {
+  dispatch(Workspaces.add(ws))
+  dispatch(WorkspaceStatuses.set(ws.id, status))
+  globalDispatch(Workspaces.add(ws)).then(() => {
+    dispatch(Current.setWorkspaceId(ws.id))
+    dispatch(refreshSpaceNames())
+  })
```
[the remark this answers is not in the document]
mason-fish: `initWorkspace` commits the given workspace to state, updates its status, and sets it to current so the user is taken to it. It is called by the workspace form to create/update a workspace after the form has already pinged, assembled, and potentially authenticated the provided workspace.

`activateWorkspace` is always used for an existing workspace that we want to switch to, but may be stale in some ways (such as version/availability, or it's accessToken has not been loaded into state from the secret storage yet), in which case we refresh that data if possible. We call this whenever we are loading a workspace we've created before (The workspace chooser menu, the page version of that chooser, login page, and loading a tab that already has a workspace open)

Calling `initWorkspace` instead of `activateWorkspace` would result in stale workspace data, and calling `activateWorkspace` instead of `initWorkspace` would make redundant/excessive api calls (since the form will have already done that part).

src/js/components/TabContent.tsx
```diff
@@ -33,7 +34,7 @@ export default function TabContent() {
     if (ws && !wsStatus) {
       dispatch(initCurrentTab())
     }
-  }, [ws, wsStatus])
```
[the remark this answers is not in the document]
mason-fish: I believe so, though I discovered this in a really fun infinite login loop which revealed that auth0 *does* indeed have rate limits 😅

src/js/components/SavedSpacesList.tsx
```diff
@@ -116,7 +116,9 @@ const SpaceListItem = ({space}: {space: Space}) => {
 
 export default function SavedSpacesList({spaces, workspaceStatus}: Props) {
   if (workspaceStatus === "disconnected")
-    return <EmptySection message="Unable to connect to service" />
+    return <EmptySection message="Unable to connect to service." />
+  if (workspaceStatus === "login-required")
+    return <EmptySection message="Login required to view spaces." />
```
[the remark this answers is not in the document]
jameskerr: Correct! That was my hope.

src/js/components/TabContent.tsx
jameskerr: I just checked the contents of `<Login />` and it doesn't look like there is any global dispatch calls that pass in the `ws` object. It looks like this is the only place `<Login />` is used as well. Changing the Login props to:

```diff
 type Props = {
-  ws: Workspace
+  ws: BrimWorkspace
 }
```

and then passing in `ws` seems to work fine with the compiler.
```diff
-  if (wsStatus === "login-required") return <Login ws={ws.serialize()} />
+  if (wsStatus === "login-required") return <Login ws={ws} />
```

If you were to need to serialize the object, it'd be best to do it at the last minute it's required. Not calling serialize also allows the ws={ws} prop to keep the same object references if it doesn't change. If later you made <Login /> a memo'd component, that would help avoid re-renders as well. Calling .serialize() returns a new object every time.

src/js/flows/workspace/initWorkspace.ts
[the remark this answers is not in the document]
jameskerr: Thanks for the explanation Mason. That makes sense. I think the words init and activate are too related. Maybe we have a `saveWorkspace` does everything that "activateWorkspace" doesn't.

If so, then when a user first creates a workspace the flow is: 1) save workspace 2) activate workspace

And when they click on an existing workspace: 1) activate workspace

I'd argue that it's better to have 2 or 3 redundant API calls if the code is more strait forward.

src/js/components/TabContent.tsx
mason-fish: Ah my bad, I had originally written `Workspaces.setWorkspaceToken` to take the whole workspace and hit the problem I described, began serializing workspace to fix, and then realized we really didn't need the whole workspace in that action and just passed id in. Thanks for pointing that out and I agree about the serialization strategy suggest.

src/js/auth0/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    loginUrl.searchParams.append("state", state)
+
+    return shell.openExternal(loginUrl.toString())
+  }
+
+  private getTokenURL(): string {
+    return new URL("oauth/token", this.auth0Domain).toString()
+  }
+
+  async refreshAccessToken(refreshToken: string): Promise<string> {
```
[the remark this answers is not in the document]
mason-fish: I thought refresh was a more meaningful verb, and gives more of a hint as to why the function accepts a `refreshToken` as input. Similar for `exchangeCode`, though perhaps `exchangeCodeForTokens` is more informative?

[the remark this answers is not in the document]
jameskerr: Cool, yeah keep the word "refresh" there then. And I like "exchangeCodeForTokens".

src/js/components/Login.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  useEffect(() => () => ctlRef.current.abort(), [])
+
+  const onClick = async () => {
+    setIsFetching(true)
+    try {
+      ctlRef.current = new AbortController()
+      const accessToken = await dispatch(
+        login(workspace, ctlRef.current.signal)
+      )
```
jameskerr, line 56: You maybe only want to wrap this in the try catch. If the lines below fail, you'd want to know about it?

src/js/components/WorkspaceModals/WorkspaceForm.tsx
```diff
@@ -95,51 +112,69 @@ const WorkspaceForm = ({onClose, workspace}: Props) => {
...
+      // user elected to cancel at dialog or mid-login
+      if (cancelled) return
+
+      // success
+      onClose()
+    } catch (e) {
+      console.error(e)
+      setIsSubmitting(false)
+      setErrors([{message: "Failed to add Workspace"}])
+    }
```
jameskerr, line 177: 🔥 Easy to read! Thanks for seeing it through.

src/js/components/WorkspacePicker.tsx
```diff
@@ -53,17 +53,15 @@ export default function WorkspacePicker() {
       checked: isCurrent,
       click: () => {
         if (isCurrent) return
-        dispatch(initWorkspace(w)).catch(() => {
-          dispatch(Current.setWorkspaceId(w.id))
-        })
+        dispatch(activateWorkspace(w.id))
```
jameskerr, line 56: Love it. Activate workspaces implies setting it to "current". Makes sense

src/js/electron/main.ts
```diff
@@ -1,32 +1,34 @@
...
+import secretsMainHandler from "./ipc/secrets/mainHandler"
+import electronIsDev from "./isDev"
+import menu from "./menu"
+import {handleQuit} from "./quitter"
+
+import {handleSquirrelEvent} from "./squirrel"
 import userTasks from "./userTasks"
 
 // app path and log setup should happen before other imports.
 appPathSetup()
```
jameskerr, line 18: Shoot, I think this appPathSetup() does need to be called above the other imports. Maybe your editor auto hoisted them up to the top?

mason-fish: Ah yes it did, I added a plugin to organize imports and auto-remove unused, I can change. Though manual and itests didn't throw any issues 🤔

src/js/flows/workspace/buildAndAuthenticateWorkspace.test.ts
```diff
@@ -0,0 +1,326 @@
...
+    removeListener: jest.fn()
+  },
+  remote: {
+    dialog: {
+      showMessageBox: jest.fn()
+    }
+  }
+}))
+
+const fxt = {
```
jameskerr, line 31: Not sure what this is based on the name. Is this the mock zqd response?

mason-fish: Oh sorry, it is just short for `fixtures` but since I use it so much I shortened the name. Shame on me

src/js/flows/workspace/buildAndAuthenticateWorkspace.test.ts
```diff
@@ -0,0 +1,326 @@
...
+      {
+        ...fxt.newWorkspace,
+        ...fxt.version,
+        authType: "none"
+      },
+      "connected"
+    )
+  })
+
+  test("existing public workspace, updated version", async () => {
```
jameskerr, line 100: Maybe this can also test that the total count of workspaces is unchanged?

src/js/flows/workspace/buildAndAuthenticateWorkspace.test.ts
```diff
@@ -0,0 +1,326 @@
...
+
+    const [cancelled, error] = await store.dispatch(
+      buildAndAuthenticateWorkspace(fxt.newWorkspace, ctl.signal)
+    )
+
+    expect(cancelled).toEqual(false)
+    expect(error).toBeInstanceOf(LoginError)
+    expect(Workspaces.all(store.getState())).toHaveLength(0)
+  })
+})
```
jameskerr, line 326: These tests are great!

src/js/flows/workspace/buildAndAuthenticateWorkspace.test.ts
jameskerr: Aww, maybe it means fixtures?

src/js/flows/workspace/buildAndAuthenticateWorkspace.ts
```diff
@@ -0,0 +1,80 @@
...
+        )
+      )
+      return [false, null]
+    } catch (e) {
+      return [false, new LoginError(e)]
+    }
+  } catch (e) {
+    return [false, new ConnectionError(e)]
+  }
+}
```
jameskerr, line 80: Right on!

src/js/flows/workspace/login.ts
```diff
@@ -0,0 +1,47 @@
...
+  )
+  return new Promise<string>((res, rej) => {
+    const handleAuthCb = handleAuth(res, rej)
+    ipcRenderer.once("windows:authCallback", handleAuthCb)
+    abortSignal.addEventListener("abort", () => {
+      ipcRenderer.removeListener("windows:authCallback", handleAuthCb)
+      res()
+    })
+  })
+}
```
jameskerr, line 47: 👍

jameskerr (approved):
Looks great!

PR #1395, opened 2021-01-27
Notify slack on master branch ci failure

mattnibs:
Add job that will notify HQ Slack should a failure occur on brim master.


mattnibs:
This is almost the same job in zq: https://github.com/brimsec/zq/blob/master/.github/workflows/notify-master-failure.yml

philrz (approved):
👍 

Having seen this, I just made the org-level secret `SLACK_WEBHOOK_BRIMLABS_TEST` visible to the Brim repo.

PR #1405, opened 2021-02-01
Color consolidation

jameskerr:
Removed all the old colors we wanted to migrate away from. In the process, found tons of dead css from components that don't exist anymore and removed them.

Also, changed some of the color values to have a bit more blue hue to them.

The only color vars that should be present are defined in colors.scss with css variables at the :root level. Usage should be `var(--aqua)` in all cases (Styled Components, JS, SCSS)


src/css/_conn-versation.scss
```diff
@@ -36,15 +36,15 @@
...
   }
 
   .responder {
     .ip {
-      background: $green;
+      background: var(--green);
     }
     .port {
-      background: $green-light;
+      background: #5ec4a8;
```
mason-fish, line 47: Do we not want to capture this (and all of the zeek log type colors for that matter) in more centralized variables as well?

mason-fish (approved):
lgtm! just one qq about the zeek colors

src/css/_conn-versation.scss
jameskerr: Good point. The Zeek log type colors are mostly centralized in this file: https://github.com/brimsec/brim/pull/1405/files#diff-6764dbaa5ca14068a22c350c9eea02ca7e3848ccdcfee72e8b1348a48ac5333d

The reason I didn't make this light green a variable is because I'm planning on re-doing that ConnVersation component at some point soon. It doesn't really fit the new detail pane width anymore.

PR #1415, opened 2021-02-04
Query Library transfer script and cookbook article

philrz:
A community user asked about being able to transfer Query Library contents. While we have future plans to make this more seamless, we discussed the topic internally and agreed that it would be reasonable to educate power users on how to safely modify the JSON themselves.

In this PR I've included a Python script that attempts to make the transfer as easy/safe as possible along with a "cookbook" article for the wiki that describes its use and limitations.

Since we use Actions to move docs content from its location in the repo to its final wiki destination, I've put up a copy of the article in a wiki in a personal repo at https://github.com/philrz/scratchwiki/wiki/Query-Library-Transfer so you can see that the links work as intended. The location of the Python script as referenced in the `curl` command lines in the article won't work until this PR is merged, but you can check it out locally for now if you want to test it.


jameskerr (commented):
Looks good Phil! I just tested the python script out in this branch and it all worked as expected.

jameskerr (approved):

PR #1416, opened 2021-02-04
Responsive Toolbar

### LINKED ISSUE #1393: Re-Org Toolbar for Summary View

jameskerr:
fixes #1393 

1. Makes the existing toolbar responsive (buttons hide when the space get's small)
2. Removes irrelevant buttons when in "summary" view.

I spent some time organizing the commits into logical chunks. Stepping through those might be a good way to review.

3iOuTPsRc6

Bxxq3a2zUm


app/toolbar/hooks/usePackets.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  useDebouncedEffect(() => setEnabled(!!conn && space.pcap_support), 100, [
+    conn,
+    space.pcap_support
+  ])
+
+  function getTitle() {
+    if (!space.pcap_support) return "This space has no pcap support."
+    if (conn) return "Open packets from this connection."
+    else return "No connection record found."
```
mason-fish, line 24: nit: don't need this `else`

app/columns/columns-modal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return ReactDOM.createPortal(
+    <CSSTransition
+      in={true}
+      classNames="dim-portal-overlay"
+      timeout={{enter: 200}}
+      onClick={onClose}
+      appear
+    >
+      <div className="portal-overlay">
+        <div className="portal-item" style={style}>
```
mason-fish, line 29: Is there any way we can style this all with just styled-components?

app/toolbar/actions.tsx
```diff
@@ -0,0 +1,26 @@
...
+import useColumns from "./hooks/useColumns"
+import useExport from "./hooks/useExport"
+import usePackets from "./hooks/usePackets"
+import useView from "./hooks/useView"
+import ResponsiveActions from "./responsive-actions"
+
+export default function Actions() {
+  const mainView = useSelector(Layout.getMainView)
+  const showSummary = useSelector(Feature.show("summary"))
+  const view = useView()
```
mason-fish, line 15: Nice, separating those hooks out really makes this clean 👌

app/toolbar/responsive-actions.tsx
```diff
@@ -0,0 +1,98 @@
...
+      (c) => c.getBoundingClientRect().width
+    )
+  }, [parent])
+
+  const onResize = useCallback(
+    ({width}) => {
+      let total = 0
+      let index = 0
+      for (let childWidth of widths.current) {
+        total += childWidth + 8
```
mason-fish, line 66: mind labeling this magic number `8`?

mason-fish (approved):
Checked this out and it looks awesome!! 🔥 

I left a couple small comments, also I think you have an empty `useMainView.ts` file that you don't need ;)

app/columns/columns-modal.tsx
jameskerr: I don't want this Portal Overlay to live much longer. This is the only place it's used, so I've added a deprecation comment. Next time we need to add features to the column picker, I'd like to delete this "portal" component. And at that time with Styled Components. Just didn't seem worth it to convert it.

app/toolbar/responsive-actions.tsx
jameskerr: Yeah totally, thanks for pointing that out.

PR #1417, opened 2021-02-04
disable target select for now

### LINKED ISSUE #1355: Hide archive index search dropdown
We won't be ready to expose index searches for archive backed spaces in the short term. Before we flip our default from file to archive storage, we need to hide/remove the search dropdown that allows index searches:

mason-fish:
fixes #1355 


jameskerr (approved):
Red is the color of love.

PR #1424, opened 2021-02-04
Don't burn the toast 🔥 🍞 

### LINKED ISSUE #1346: Export continues long after "Exporting..." pop-up goes away
Repro is with Brim commit `f6c7ba3`.

The attached video shows the initiation of a fairly large CSV export (specifically, the log contents after the import of https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz, uncompressed). After the "Exporting..." pop-up fades away, you can see the CSV file continues to increase in size, indicating the export operation is still in-progress. As a user, a pop-up that communicates an in-progress operation (e.g. a verb ending in "ing"), my assumption is that when the pop-up goes away, the operation described has completed.

https://user-images.githubusercontent.com/5934157/104105536-b19c9100-5263-11eb-828a-2e54325daff0.mp4

mason-fish:
fixes #1346 

The `react-hot-toast` component we use to indicate the loading state during export has a default timeout of 30s, after which it will dismiss itself. This comes off as an indication to the user that the export is complete even though it may not be (and then when it finally does complete the "Export Complete" toast happens but the user would not expect it since the loading was already dismissed). Allowing the toast to persist indefinitely is not yet a feature of the toast lib, though this happened to have been requested a couple of days ago and there is already a PR to enable it in their pipeline (see https://github.com/timolins/react-hot-toast/pull/37 and https://github.com/timolins/react-hot-toast/issues/36). In the meantime, we can set the duration to int32 max which is the highest value that setTimeout will accept. `Number.MAX_VALUE`, and `Number.MAX_SAFE_INTEGER` could both be too large and overflow, so we define it explicitly now.

I also added another second to each of the default values for the success and error cases, since I thought the defaults were a tad short for my taste. 

### DIFF
--- a/src/js/components/ExportModal.tsx
+++ b/src/js/components/ExportModal.tsx
@@ -14,6 +14,7 @@ import {SearchFormat} from "../../../zealot"
 import InputLabel from "./common/forms/InputLabel"
 import {defaultModalButton} from "../test/locators"
 import {toast} from "react-hot-toast"
+import {AppDispatch} from "../state/types"
 
 const RadioButtons = styled.div`
   display: flex;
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Great! Thanks for do the research on this.

PR #1426, opened 2021-02-05
Upgrade eletron to 11.2.1

jameskerr:
The React DevTools are broken in version 8 - 11. The fix is in electron master. This helps us get up to speed when the fix is in.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9298,9 +9298,9 @@
       "dev": true
     },
     "electron": {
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
I'd love to get those dev tools back! lgtm, assuming those failing iTests aren't related to the bump

jameskerr:
Oh shoot, we need to upgrade Spectron as well.

PR #1430, opened 2021-02-06
Join cookbook

philrz:
In the time after we first developed a new/experimental `join` processor, we had two separate community users express an interest in making use of it. In each case I explained its use to them via Slack and was able to guide them through some rough edges to make it ultimately work for their use cases. It felt like a good learning experience for all involved.

More recently we had yet another community user express interest in `join`. Rather than having to explain over Slack the caveats and limitations we'd learned from the initial users, I figured it's time to put together one of our "cookbook" articles that walks through its use more formally. In addition to hopefully addressing the needs of this most recent user, this also gives us the opportunity to advertise the processor's availability to the wider community and solicit more feedback on its use.

As our Brim docs are not pushed to the wiki until they're merged to master, I've put a copy of this markdown in a scratch repo-with-a-wiki at https://github.com/philrz/scratchwiki/wiki/Joining-Data so reviewers can see it formatted and with links intact.

This article references the new `split` / `=>` syntax that is not yet available in GA zq/Brim releases. If approved, I intend to hold off on merging this PR until right before we publish the next release.


mccanne (approved):
Awesome work!

PR #1433, opened 2021-02-08
Remove Security Summary From State Until Shipped

jameskerr:
Before we release, I'm backing out some changes we made to the state for the unfinished security summary feature. I'd rather avoid a migration when we ship the security summary.

This is all hidden behind a feature flag which still works. The tiles are now effectively hard-coded for the time being.

When the security summary ships:

- [ ] Remove the @ts-ignore comments in the Boards/test.ts and the Tiles/test.ts
- [ ] Remove the skip prefix from all the tests
- [ ] Add the reducers to the rootReducer
- [ ] Add the reducers to the globalReducer


mason-fish (approved):
👍

PR #1440, opened 2021-02-09
handle zq AST change wrt field references with RootRecord

mccanne:
This commit changes the ast-to-field-name logic to handle
Identifier nodes that do not appear in a dotted expression
with a RootRecord as was recently introduced in the Z grammar
and AST format.


jameskerr (approved):
Thanks for the PR @mccanne

PR #1443, opened 2021-02-09
Create Default Security Tiles

### LINKED ISSUE #1392: Create Default Security Summary Tiles

jameskerr:
image

closes #1392 

This PR adds the default set of queries we want to run for version 0 of the security summary.

Included features:
* A skeleton table animation plays when the tile is loading
* A single menu action to "Search for all"
* Bars added to the table rows to signify value
* Created an `app/tile` folder to hold all the components needed to render a tile
* Made the grid only have 4 columns instead of 12. (the current tiles only use two columns)

Todo in other PRs:
* Don't issue searches until the tile is in view
* Open right-click pivots in new tabs (or provide option to)
* Do the same with the tile menu
* Incorporate the "mainView" into the tab history


app/core/hooks/useSearch.ts
```diff
@@ -9,9 +9,10 @@ type R = [zng.Record[], boolean]
 export default function useSearch(query: string, deps?: any[]): R {
   const dispatch = useDispatch<AppDispatch>()
   const [records, setRecords] = useState<zng.Record[]>([])
-  const [isFetching, setIsFetching] = useState<boolean>(false)
+  const [isFetching, setIsFetching] = useState<boolean>(true)
 
   useEffect(() => {
+    setIsFetching(true)
```
mason-fish, line 15: if we default `isFetching` to `true` in the useState above, can we not rely on the response.status handler below to solely maintain the fetching state?

app/tile/menu-button.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type Props = {
+  query: string
+}
+
+export default function MenuButton({query}: Props) {
+  const dispatch = useDispatch()
+
+  return (
+    <BG onClick={() => dispatch(showMenu(query))} className="menu-button">
```
mason-fish, line 58: darn, wish we could just use styled components here instead of both 💅 and the css from ".menu-button", makes it a bit difficult to know what the reconciled style will be after the latter overwrites the former

app/tile/table-skeleton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  &:nth-child(even) {
+    background: none;
+  }
+  position: relative;
+`
+
+export default function TableSkeleton() {
+  return (
+    <BG>
+      {Array(10)
```
mason-fish, line 35: I was gonna suggest to make the number of rows be a prop, with a default of 10 so we could easily adjust based on amount of records from a query with `... | head n`, but I see below that we don't actually know how many records we will get until search is complete and then this component is replace.

So for now, is it the case that we have a constraint to keep our summary queries to 10 results or risk seeing a change in size to the tiles on completion?

app/tile/viz.tsx
```diff
@@ -0,0 +1,66 @@
...
+      return null
+  }
+}
+
+type Props = {
+  format: TileFormat
+  records: zng.Record[]
+}
+
+function Viz({format, records}: Props) {
```
mason-fish, line 62: what does `Viz` stand for? I'm not sure what a better name would be, `<TileContent />` maybe?

ppl/summary/summary.tsx
```diff
@@ -35,12 +36,14 @@ type Tile = {
 }
 
 function getGridLayout(tiles) {
-  return tiles.map((t) => ({i: t.id, ...t.layout}))
+  return tiles.map((t) => ({i: t.id, ...t.layout, static: true}))
```
mason-fish, line 39: Are we disabling moving of tiles bc we don't want to also support storing that in state right now? or are there also bugs with the actual act of moving them around. I'm also just assuming that's what `static: true` means here.

mason-fish (approved):
Looks great! I left a few non-blocking comments and questions

mason-fish:
I just noticed there still is no empty state when there are no results for tables, which I recall you observing during 🎸 session. Were you planning on addressing this in a different ticket?

image

app/core/hooks/useSearch.ts
jameskerr: The status handler takes a few milliseconds to update the response which resulted in a flash of empty state before isFetching ui kicked in.

Setting it here early fixed it.

app/tile/menu-button.tsx
jameskerr: You know, the only reason I needed the class name is because I couldn't figure out how to reference this menu-button component inside of the tile.tsx. 

I tried:

const StyledMenuButton = styled(MenuButton)

const BG /* tile bg */ = `
  ${StyledMenuButton} {
     // my styles to show/hide
  }
` 

But that didn't work. Was I doing something wrong?

app/tile/table-skeleton.tsx
jameskerr: The tiles will not change size every. They have a fixed width/height. If a table gets too large, the overflow is hidden.

I did add head 10 for all the queries on the board. It seems like a good constraint for now. But yes, I do think that the skeleton should have a prop for the number of rows. I'll add it in a later PR most likely.

app/tile/viz.tsx
jameskerr: Viz is for Visualization.   I can change it to visualization...

Content is used in another place. Body?

ppl/summary/summary.tsx
jameskerr: Yes, your assumption is correct on both fronts. We're going to release the first version with no moving or resizing. Then upgrade later.

jameskerr:
Thanks for finding those bugs in your last comment @mason-fish . I'm going to open new tickets to fix those.

https://app.zenhub.com/workspaces/brim-5e84a867c8ec919560459033/issues/brimsec/brim/1457
https://app.zenhub.com/workspaces/brim-5e84a867c8ec919560459033/issues/brimsec/brim/1458

app/tile/menu-button.tsx
mason-fish: Oh, yeah I think its bc the `styled` template requires the back ticks to work, so:

const StyledMenuButton = styled(MenuButton)``

and then styled components require that the component it wraps takes and passes in a `className` prop. So perhaps more simply, you can just export the menu-button as a styled component like this

PR #1447, opened 2021-02-10
Fix History Back, Push Bugs

### LINKED ISSUE #1331: "Back" button is not returning to previously-viewed record
Repro is with Brim commit `803b151` and with test pcap https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz (uncompress before import). It does look like this is not a new bug, though, as I was able to repro in Brim GA tagged `v0.21.1` as well.

Per the attached video, repro steps:

1. Start by searching for `"C1md9p1UxupPOvCs9f"` and bring up the `conn` record in the Log Detail window
2. Click the tile for the first `http` event and then click the Back button. I'm brought back to the `conn` event as expected.
3. Click the the tile for the first `files` event and then click the Back button. At that point I expected to be brought back to the `conn` event again since it's what was just displayed. Instead the Back button brings me back to the `http` event, so I need to click Back yet again to get back to the original `conn` event.
4. If I click to other tiles, it seems to only increase the "stack" of additional events I have to click through via Back to return all the way to the original `conn` I'd started at.

https://user-images.githubusercontent.com/5934157/103717127-88d17e80-4f79-11eb-828f-6e202c86fbb7.mov

jameskerr:
Fixes #1331 

There were 3 implementations of a History class in the app:

1. LogDetailsHistory.ts
2. InputHistory.ts
3. brim/entries.ts

This PR creates a single `app/core/models/history.ts` which is used in all the places the above 3 classes were used.

The bug was ins the "push" method of both `LogDetailHistory.ts` and `InputHistory.ts`. It did not properly "splice" the new record at the correct position, then delete all following entries. It simply "pushed" a new entry onto the array. The `brim/entries.ts` implementation did have it correct.

In the first commit I create a failing test that reproduces the bug.

To Do: 
- [x] Fix the Input History. I actually needs to behave a little more simply than the others.


src/js/components/SearchBar/Input.tsx
```diff
@@ -25,10 +25,11 @@ const StyledInput = styled.input`
     outline: none;
   }
 `
+// create a tiny little history closure
```
mason-fish, line 28: Is this comment referring to this component or just a leftover note to self?

mason-fish (approved):
This is great, thanks for refactoring everything so neatly, love all that red. Just left a "tiny little" comment.

src/js/components/SearchBar/Input.tsx
jameskerr: A left over note to self!

jameskerr, line 28:
```suggestion
```

PR #1456, opened 2021-02-11
Fix advance zq workflow

mattnibs:


philrz (approved):
I tested this out successfully in a local fork of Brim with a simplified Workflow https://github.com/philrz/brim/blob/master/.github/workflows/advance-zq.yml which I could make fail or succeed  by toggling between `run: true` and `run: false`. So, indeed, it looks like going to single quotes is going to do the trick here. 👍

PR #1468, opened 2021-02-17
clear current spaceId if partial deleting

### LINKED ISSUE #1337: "Item does not exist" pop-up when relaunching Brim after aborted import
Repro is with an artifact based on Brim commit `2b45a97` which talks to GA `zqd` tagged `v0.27.0`. I was able to repro this with older releases as well, though.

Per the attached video, if a user goes to quit the app while a pcap import is ongoing, they're prompted with a pop-up that advises them that the partial generated data will be deleted if they quit. If I accept this option, when I relaunch the app, the Space is not present (as I would expect) but there's also a pop-up that says "Item does not exist" (that I did not expect).

https://user-images.githubusercontent.com/5934157/103954033-2bb00700-50f8-11eb-8904-b63910f764a3.mov

Sniffing in Wireshark, I can see that this message is coming from `zqd` in reaction to the app asking for info about a Space. Example:

### Request

```
GET /space/sp_1mlDwGJwyGS1kR6cfjNyGTjMMFC HTTP/1.1
Host: localhost:9867
Connection: keep-alive
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_16_0) AppleWebKit/537.36 (KHTML, like Gecko) Brim/0.21.1 Chrome/85.0.4183.121 Electron/10.1.4 Safari/537.36
Accept: */*
Sec-Fetch-Site: cross-site
Sec-Fetch-Mode: cors
Sec-Fetch-Dest: empty
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US
```

### Response

```
HTTP/1.1 404 Not Found
X-Request-Id: 3
Date: Thu, 07 Jan 2021 22:45:34 GMT
Content-Length: 76

{"type":"Error","kind":"item does not exist","error":"item does not exist"}
```

I understand that if this were a _catastrophic_ exit (e.g. send the app a `kill -9` instead of the friendly Command-Q-to-quit I did in the video) then there might understandably be clutter left behind that might trigger an error like this. However, since the app gave all appearances of a clean exit, if there was partial data to be cleaned up as the first pop-up implied, I'd have expected the app to have completed that as part of the quit such that the user sees no error message on relaunch.

mason-fish:
fixes #1337 (th3 m05t l337 t1ck3t 3v3r)

When closing brim and deleting partial spaces, check if the current spaceId is among those being cleaned up, if so then clear it's current spaceId.

### DIFF
--- a/src/js/flows/deletePartialSpaces.test.ts
+++ b/src/js/flows/deletePartialSpaces.test.ts
@@ -1,0 +1,50 @@
+import {createZealotMock} from "zealot"
+
+import Workspaces from "../state/Workspaces"
+import Current from "../state/Current"
+import fixtures from "../test/fixtures"
+import deletePartialSpaces from "./deletePartialSpaces"
+import initTestStore from "../test/initTestStore"
+import Handlers from "../state/Handlers"
+
+const testSpaceId1 = "testSpaceId1"
+const testSpaceId2 = "testSpaceId2"
+
+let store, zealot
+beforeEach(() => {
+  zealot = createZealotMock()
+  const ws = fixtures("workspace1")
+  store = initTestStore(zealot.zealot)
+  store.dispatch(Workspaces.add(ws))
+  store.dispatch(Current.setWorkspaceId(ws.id))
+})
+
+test("reset current space id if mid-ingest", async () => {
+  zealot.stubPromise("spaces.delete", true)
+  store.dispatch(
+    Handlers.register("id-1", {type: "INGEST", spaceId: testSpaceId1})
+  )
+  zealot.stubPromise("spaces.delete", true)
+  store.dispatch(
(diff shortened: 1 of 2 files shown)


jameskerr (changes requested):
Looks correct to me. Would you put a unit test right next to this file that tests this functionality? Something like:

```
// setup
  mock = createZealotMock()
    .stubPromise("spaces.delete")
  store = initTestStore(mock.zealot)

// two of these
dispatch(Handlers.register("id-1", {type: "INGEST"spaceId: string}))
dispatch(Handlers.register("id-2", {type: "INGEST"spaceId: string}))

// set the current space to one of those
// run the function
// expect(zealot.calls("spaces.delete")).toHaveLength(2)
// make sure current spaceId is null (and another test to make sure current is not null)
```

jameskerr (approved):
♥️ thanks

PR #1469, opened 2021-02-17
remove azng related itest

alfred-landrum:
Removes test for azng, related to https://github.com/brimsec/zq/pull/2158 .


mason-fish (approved):
👍

PR #1473, opened 2021-02-18
Brim on URLs

### LINKED ISSUE #1317: Edit Pin Upgrades
As a result of chatting about some existing pin bugs, Mason and I discussed a brighter future for the pins UI. This was the result:

When clicking a pin to edit it, turn that pin into a small input in place. Do not reuse the search box to edit the pin. Display a checkbox, or hit enter to submit your changes to the pin. When done, move focus back to the search box, but do not issue a search. 

This will allow us to remove the "last search box" special pin at the end.

This will also allow us to remove the editing index from the search bar reducer. 

See if the searchBar.previous piece of state can be removed too.

### LINKED ISSUE #1446: Incorporate Summary View into Tab History
If you pivot from a summary view into the search view, then click the back button. We need to take you back to the summary view. It might also be a good idea to keep the summary view rendered for a tab, but just hidden. That way if they click back, the page loads quickly again.

jameskerr:
This PR introduces routing and tab history using URLs. It uses `react-router` to accomplish the routing and a custom `Histories` class to store the histories for each tab.

### Key Files

This PR is big, so I'll point out a few key files to review:

1. The list of all the routes: [app/router/routes.ts](https://github.com/brimsec/brim/blob/router/app/router/routes.ts)
2. The top level App component using the routes: [App.tsx](https://github.com/brimsec/brim/blob/router/src/js/components/App.tsx)
3. [Workspace Routes](https://github.com/brimsec/brim/tree/router/app/workspaces)
4. [Lake Routes](https://github.com/brimsec/brim/tree/router/app/lakes)
5. The Histories class: [histories.tsx](https://github.com/brimsec/brim/blob/router/app/core/models/histories.ts)
6. The TabsRouter class: [tabs-router.tsx](https://github.com/brimsec/brim/blob/router/app/router/tabs-router.tsx)

### New Assumptions

In the past, we would change pages by changing certain key parts of the window-redux-state like (current.spaceId,  current.workspaceId, layout.mainView, SearchBar.restore,...etc). We now simply push to the history like so: `dispatch(tabHistory.push("/workspaces")). When this happens...

1. The router will only render the component that matches that path.
2. All components that subscribe to the location changes will read the url and update the window-redux-state to match it (for example, [search/home.tsx](https://github.com/brimsec/brim/blob/router/app/search/home.tsx#L23).

For example, the new code for the often used `submitSearch()` now [looks like this](https://github.com/brimsec/brim/blob/router/src/js/flows/submitSearch/mod.ts):

1. User changes the search bar inputs which get saved to redux.
2. User clicks the blue search button.
3. the `submitSearch()` thunk will grab the relevant pieces of redux state (input, pins, timespan, lakeId, workspaceId), [encode it into a url](https://github.com/brimsec/brim/blob/router/app/search/utils/search-params.ts#L3), the push it to the history.

4. The [viewer](https://github.com/brimsec/brim/blob/router/src/js/components/SearchResults/SearchResults.tsx) and the [histogram](https://github.com/brimsec/brim/blob/router/src/js/components/SearchHeaderChart.tsx) both subscribe to location changes and will run their respective backend queries using the params in the url.  

### Drive Bys

* Anywhere I saw "subspace" code or `target: "events"`, I deleted it. This was a relic of the subspace era which we are moving on from.

* I deleted some code related to the time span. We had these "steppers" that would appear when you hovered over a date part. They have not been used in a while so they're gone now.

### Questions

Do we need to migrate the user's current tab history (not the investigation history in the left sidebar)? Another way to ask this is, if a user used the app for a while, quit with a window open. Then relaunched it, would we want their back and forward buttons to keep their history state? I'm going to say yes we do. Why not. In which case I"ll need to write a migration.

### Things we can delete later
UPDATE: These have been deleted

1. The Current reducer
2. The Last reducer
3. The History reducer

To do:

- [x] Change `submitSearch` to only save to investigation, and navigate to the next search url
- [x] Create routes for "/search" and "/summary"
- [x] Have those routes listen for location changes and fetch new data when it changes using the UrlSearchParams
- [x] Create a new function that does the fetching for the main viewer and histogram. Copy the code from the old `submitSearch` function.
- [x] Write function to dump the history entries into the redux state each time the tab changes or the app is quitting
- [x] Write a function load up the history entries from the redux state when a tab is activated
- [x] Update the History reducer function to store history entries instead of SearchRecords
- [x] Remove the "Last" state slice since that data will be in the url?
- [x] Write a migration to convert all the persisted search records into history entries

Update:

I've also fixed #1317 by making the pins directly editable. 

Also closes #1446.

### DIFF
--- a/app/search/flows/histogram-search.test.ts
+++ b/app/search/flows/histogram-search.test.ts
@@ -40,7 +40,7 @@ test("zealot gets the request", async () => {
   await submit()
   const calls = zealot.calls("search")
   expect(calls.length).toBe(1)
-  expect(calls[0].args).toEqual("* | every 12hr count() by _path")
+  expect(calls[0].args).toEqual("* | every 12h count() by _path")
 })
 
 test("the chart status updates", async () => {


app/core/icons/reload.svg
```diff
@@ -0,0 +1,3 @@
...
+</svg>
```
jameskerr, line 3: image

jameskerr: An icon to refresh the summary data.

app/lakes/show.tsx
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    if (lakeId) {
+      setFetching(true)
+      dispatch(initSpace(lakeId)).finally(() => setFetching(false))
+    }
+  }, [lakeId])
+
+  if (fetching) return null
+  return children
+}
```
jameskerr, line 44: This pattern is inspired by something I learned reading this this week: https://overreacted.io/before-you-memo/

src/js/components/Span/TimeButton.tsx
```diff
@@ -87,25 +42,12 @@ export default function TimeButton({
...
   return (
-    <div
-      className={classNames("time-picker-button", {
-        hovering: fuzzy.hovering
-      })}
-      onMouseLeave={fuzzy.mouseLeave}
-      onClick={onClick}
-    >
-      <div className="hover-zone" />
-      <TimeSteppers
```
jameskerr, line 98: These time steppers were not being used anymore. Along with all the mouse handling above.

src/js/flows/initSpace.ts
```diff
@@ -1,33 +1,23 @@
...
-      const space = brim.space(data)
       dispatch(Spaces.setDetail(workspaceId, data))
-      dispatch(Current.setSpaceId(space.id))
-      dispatch(Search.setSpanArgs(space.everythingSpan()))
-      dispatch(SearchBar.removeAllSearchBarPins())
-      dispatch(SearchBar.changeSearchBarInput(""))
-      if (!space.hasIndex()) {
-        dispatch(SearchBar.setTarget("events"))
-      }
-      dispatch(submitSearch({history: true, investigation: false}))
```
jameskerr, line 30: All this is done is LakeSearch

src/js/state/Current/selectors.ts
```diff
@@ -1,20 +1,43 @@
...
 import Workspaces from "../Workspaces"
-import Spaces from "../Spaces"
-import activeTabSelect from "../Tab/activeTabSelect"
-import brim, {BrimWorkspace, BrimSpace} from "../../brim"
+import {WorkspacesState} from "../Workspaces/types"
 
 type Id = string | null
 
-export const getSpaceId = activeTabSelect((state) => state.current.spaceId)
+export const getHistory = (state, windowName = global.windowName) => {
```
jameskerr, line 13: This is how you get the history from a thunk.

src/js/state/Search/selectors.ts
```diff
@@ -1,13 +1,38 @@
...
+import {addEveryCountProc} from "../../searches/histogramSearch"
+import {SearchRecord} from "../../types"
 import Current from "../Current"
 import SearchBar from "../SearchBar"
 import Tab from "../Tab"
-import brim from "../../brim"
+import {State} from "../types"
+import {SearchArgs, SearchType} from "./types"
+
+const getCurrentRecord = (state: State): SearchRecord => {
```
jameskerr, line 13: Can't wait to get rid of this once we fix the search bar pins.

src/js/state/Tabs/actions.ts
```diff
@@ -9,10 +8,10 @@ import {
 } from "./types"
 
 export default {
-  add: (id: string, data?: AddTabData): TABS_ADD => ({
+  add: (id: string, url: string): TABS_ADD => ({
```
jameskerr, line 11: 🥳

philrz:
I was asked to start some early testing on this branch, so I'll post my findings here. I'm testing on a build [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-master-b40b95d9.0/macos/Brim.dmg](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-master-b40b95d9.0/macos/Brim.dmg) (so, based on Brim commit b40b95d, which is the current tip of the `router` branch) and comparing behaviors with the app running in Dev mode at commit 6bfb29a. For test data, I've imported wrccdc.2018-03-23.010014000000000.pcap.gz (after uncompressing).

In this first issue, I happen to have a Space selected when I close the app. When I relaunch, I'm not brought up into the Space automatically like I am on the current tip of `master`. Here's the repro:

https://user-images.githubusercontent.com/5934157/109985685-e0f6d980-7cb9-11eb-8b8f-5e0fe74334f1.mov

Incidentally, I also happened to have a particular Z query in effect when I quit, and when I clicked the Space after relaunch, I saw that query was automatically placed in the Search bar on the Security Summary, which I assume was not intended. I know that testing Security Summary is not the goal of this effort and that it's likely to be hidden away behind a feature flag before we release, so maybe both problems will magically disappear.

philrz:
In this next one, I do the following in order:

*   Enter search term `10.0.0.227`, then pin
*   Add search term `_path=http`, then pin
*   Right-click to "Count by field"
*   Click the Back button

In the build based on this branch, the rightmost term after hitting Back is the `count() by`, even though what's shown is the query result that would have gone with the expected search term `_path=http`.

https://user-images.githubusercontent.com/5934157/109988440-611e3e80-7cbc-11eb-9a8a-85d3c038914c.mov

My app running in Dev mode based on commit `6bfb29a` is not showing the problem.

https://user-images.githubusercontent.com/5934157/109988666-9dea3580-7cbc-11eb-9f9f-965955b2e3c7.mov

philrz:
Here's another involving pinning. I'm starting from having already done:

*   Enter search term `10.0.0.227`, then pin
*   Add search term `_path=http`, then pin
*   Right-click to "Count by field"

Now I tack on `,id.resp_h` to my "Count by", then hit Back. As shown in the attached video, the search output updates, but the rightmost pinned term term shown below the Search bar remains the same. I don't see this behavior on the app at commit `6bfb29a` when I run it in Dev mode.

https://user-images.githubusercontent.com/5934157/109990357-2b7a5500-7cbe-11eb-810d-5abfde8de6ed.mov

jameskerr:
Thanks for your early testing @philrz ! I've fixed the following:

✅  Start up issue fixed
✅ Right most pin issue fixed. I decided to fix it by implementing another ticket in the ice box: #1317  The right most pin is removed and the pins are directly editable. The only reason we had that special pin at the end was to get you out of pin editing mode. Here's a video of pins being directly editable now.

NGne2jx3Kq

app/core/icons/reload.svg
jameskerr: This is for manually refreshing the summary dashboard. There is no "submit search" button to accomplish this, so I figured a refresh button like this would work for now.

app/core/models/histories.ts
```diff
@@ -0,0 +1,71 @@
...
+    for (const {id, entries, index} of data) this.create(id, entries, index)
+  }
+
+  listen(listener: LocationListener) {
+    this.listener = listener
+    this.histories.forEach((h, id) => this.listenTo(id, h))
+    return () => this.unlistens.forEach((fn) => fn())
+  }
+
+  create(id: string, initialEntries?, initialIndex?) {
```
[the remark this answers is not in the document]
jameskerr: It means they are optional arguments to this function.

app/router/utils/paths.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function workspacesPath() {
+  return "/workspaces"
+}
+
+export function workspacePath(id: string) {
+  return `/workspaces/${id}`
+}
+
+export function lakeImportPath(parentId: string) {
```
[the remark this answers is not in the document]
jameskerr: I agree. I'll change.

app/toolbar/toolbar.tsx
```diff
@@ -28,16 +29,22 @@ const Row = styled.div`
...
   return (
     <Wrap>
       <SearchPageTitle />
       <Row>
         <Group>
           <MainViewSwitch />
         </Group>
-        <Actions />
-        <SpanControls />
+        <ResponsiveActions actions={actions} locationKey={location.key} />
```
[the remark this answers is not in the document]
jameskerr: Great point.

itest/lib/appStep/api/search.ts
```diff
@@ -13,7 +14,8 @@ export const runSearch = async (app: Application, searchText: string) => {
     await input.waitForDisplayed()
     return await input.setValue(searchText)
   })
-  await logStep("click the search button", () => app.client.keys("Enter"))
+  const button = await app.client.$(submitButton.css)
```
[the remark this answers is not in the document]
jameskerr: Yeah, for some reason enter was not cutting it in some tests. Now that we default to the security summary view.

PR #1476, opened 2021-02-19
Global Dispatch Design Update

jameskerr:
Instead of calling globalDispatch manually, mark which actions are intended to be global with a `$` prefix. Then have redux middleware look for the `$` prefix and make the ipc call.

I also changed the order in which the originating window receives the action. First the local store will get the action, _then_ the ipc call will be made to update the other stores. This fixes the annoying issue of dispatching an update and not being able to read that update until it makes the pic round trip.

The commits are organized in a helpful way for reviewing.


mason-fish (approved):
Great, so global actions all begin with `$` from now on and we will use a single dispatcher to demux to the other window stores when it sees that

PR #1478, opened 2021-02-19
Replace install animation & icons on Windows

### LINKED ISSUE #1037: Replace confusing animation shown during Windows install
When @orochford was giving his first impressions of the install experience on Windows, he was briefly thrown by the graphic that's shown during install, joking that "initially my thought was a graphic driver issue". It look like this, with animation effects that make it look like a moving cursor and blocks of code being added, perhaps simulating an IDE?

Screenshot

My understanding is that this is a by-product of whatever Electron installer we currently use. I admit that I was thrown by this back when I first saw it also, since most installers pop up some kind of text-based guidance saying "Installing"... or some other animated graphic, such as Slack's which shows an animation of their logo.

While we've not received outside complaints to date about the installer experience, as the app becomes accessible to more diverse users, it'd probably be good to figure out if we can improve on this.

### LINKED ISSUE #1237: Use a branded icon on Windows
@henridf noted that we're using a generic icon for our `Brim-Setup.exe`:

image

I usually take my cues from Slack in these situations since they're an Electron app that's been around much longer, so I assume if there's build tools that make it possible to produce something more polished, they're taking advantage of it. Indeed, their installer icon is branded, so at some point we'd surely want to do something similar.

image

philrz:
A recent user issue from the community boosted the priority of doing something about the confusing default animation in the Windows installer (#1037). Once @jameskerr pointed out the likely way to address it, we could see that the icons could use the same treatment, so I've knocked those out at the same time.

The video below shows the install process including the install animation and both new icons. It's a little choppy because I had to lower the frame rate to fit within GitHub's size limits, so I recommend trying the build that has these changes ([link](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-master-85b1fb39.0/windows/Brim-Setup.exe)) so you can see the install animation is smoother than it looks here.

https://user-images.githubusercontent.com/5934157/108541869-d854db00-7297-11eb-8cab-b9908d2f25d6.mp4

I learned a couple things along the way. First, my initial attempts at replacing the animation failed and the error messages weren't very helpful in figuring out why. Thankfully I found https://github.com/electron/windows-installer/issues/254 and indeed just shrinking my animated GIF to be the same width as the default one allowed it to install. I didn't spend time fiddling with seeing if there were sizes between the default one & my initial giant one that could be workable. If my animation is deemed too ugly and someone wants to spend time replacing it with something more fancy, you might want to fiddle more with that. But considering our top priority is to be "less confusing" moreso than "perfect", this seems to get the job done.

It may look odd to you that the icon that shows up in the "Programs" Settings & Control Panel is a URL and not a local file. This is discussed at https://github.com/electron-archive/grunt-electron-installer/issues/73. Therefore, I 🤷‍♂️  and go with the flow.

Just as a reminder of what we're moving away from, here was the former icon as it used to look in Explorer:

image

And the one that used to show up in Programs contexts:

image

Admittedly, our icon kinda blends in camoflauge-style with the blueness of the icon backgrounds in that default Windows theme. But the word "Brim" is still in there, it's consistent with our brand, and people probably aren't staring at this particular icon too often, so I'm still fine going this way rather than the default Atom icon that would likely be met with a a "WTF is that?" from our non-developer users. Once again, if someone wants to design something that would stand out better here, we can get it onto the website so we could link to that instead, but I don't wanna make this more work than it has to be.

image

Closes #1237
Closes #1037


jameskerr (approved):
This is so much better than what we had before. I am happy to merge this. We can polish the animations and icons later. Thanks for fixing this.

PR #1484, opened 2021-02-22
Global TabHistory & TabHistories

jameskerr:
Needed for: https://github.com/brimsec/brim/pull/1473
Relates to: #1446 

### New Globals

In this PR, I introduce two new global variables to the search page.

1. tabHistory: this a [memory history instance](https://github.com/ReactTraining/history/blob/master/docs/api-reference.md#creatememoryhistory) and will always be associated with the active tab
2. tabHistories: a map of all the tab ids to their history instance

image

### Lifecycle

In the `Tabs.reducer`, I update the values of those two new globals whenever a new tab is added, activated, or removed. This is breaking the rules since side effects are not recommend in reducers, but also seems like the best place for it. I am open to other ideas. 

The tabHistory object doesn't do anything to app at the moment, that's in another branch.

### Persistance

1. During initialization, select all the persisted entries from the store and pass them to the constructor of the `Histories` class
2. During shutdown, serialize the `tabHistories` instance and save it to redux

While the app is being used, redux will not be synced with the entries. This is what [react-redux recommends](https://reactrouter.com/web/guides/deep-redux-integration).

### DIFF
--- a/app/core/models/histories.test.ts
+++ b/app/core/models/histories.test.ts
@@ -1,0 +1,71 @@
+import Histories from "./histories"
+
+test("tab histories create / remove", () => {
+  const histories = new Histories()
+  histories.create("tab-id-1")
+  expect(histories.count()).toBe(1)
+
+  histories.delete("tab-id-1")
+  expect(histories.count()).toBe(0)
+})
+
+test("tab histories get ", () => {
+  const histories = new Histories()
+  histories.create("tab-id-1")
+
+  const tabHistory = histories.get("tab-id-1")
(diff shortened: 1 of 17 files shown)


mason-fish (commented):
Just curious, why put these in the global object? I understand that's how `location` works in the browser, but seems more controlled to just use our existing redux global store, unless the subscriber to this data won't have access to the store?

jameskerr:
I will close this PR in favor of #1473 which is based off this commit, but has changed somewhat.

PR #1489, opened 2021-03-02
update Windows Networking Activity query

### LINKED ISSUE #1485: Update Z for Query Library entry "Windows Networking Activity"
As of GA Brim tagged `v0.24.0`, the Query Library entry for **Windows Networking Activity** looks like:

```
_path=~smb* OR _path=dce_rpc
```

However, the the recent Z search/expr changes in https://github.com/brimsec/zq/pull/2072, have dropped the special `=~` operator. So left untouched, this query would no longer match is its original intent, as is visible in the canonical Z representation where we see the `~` would become part of the search term.

```
$ ast -C '_path=~smb* OR _path=dce_rpc'
filter _path=/^~smb.*$/ or _path="dce_rpc"
```

We discussed this as a team and the UX Devs acknowledged that it should be possible for us to create a migration that drops the `~` from this query. Since we don't yet have a clean way to isolate what's a "stock" query vs. one the user added vs. a stock one that's been modified by the user, probably the cleanest approach will be to only modify it automatically if it's byte-for-byte the same as its original form. I'm guessing most users probably haven't tweaked this one anyway. 🤞

mason-fish:
fixes #1485 

### DIFF
--- a/src/js/state/migrations/202103011928_adjustWindowsNetworkingActivityQuery.ts
+++ b/src/js/state/migrations/202103011928_adjustWindowsNetworkingActivityQuery.ts
@@ -1,4 +1,5 @@
 import {getAllStates} from "../../test/helpers/getTestState"
+import isEqual from "lodash/isEqual"
 
 export default function adjustWindowsNetworkingActivityQuery(state: any) {
   const shippedQuery = {
@@ -15,10 +16,7 @@ export default function adjustWindowsNetworkingActivityQuery(state: any) {
     const targetQuery = Object.values(s.queries.items).find(({id}) => {
       return id === "3"
     })
-    if (
-      targetQuery &&
-      JSON.stringify(targetQuery) === JSON.stringify(shippedQuery)
-    ) {
+    if (targetQuery && isEqual(targetQuery, shippedQuery)) {
       // @ts-ignore
       targetQuery.value = "_path=smb* OR _path=dce_rpc"
     }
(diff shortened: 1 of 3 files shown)


src/js/state/migrations/202103011928_adjustWindowsNetworkingActivityQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  for (const s of getAllStates(state)) {
+    if (!s.queries) continue
+    const targetQuery = Object.values(s.queries.items).find(({id}) => {
+      return id === "3"
+    })
+    if (
+      targetQuery &&
+      JSON.stringify(targetQuery) === JSON.stringify(shippedQuery)
```
jameskerr, line 20: I was curious, so I looked it up. It appears `JSON.stringify` does not order object keys the same way every time. If the two objects are equal but have keys in different orders, then it might not match. It'd be more reliable to use `Lodash.isEqual()` in this case.

The quote from MDN:

> Note: Properties of non-array objects are not guaranteed to be stringified in any particular order. Do not rely on ordering of properties within the same object within the stringification.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

jameskerr (changes requested):

jameskerr (approved):

PR #1494, opened 2021-03-04
Robust Detail Pane

### LINKED ISSUE #1475: Log Detail pane crashes when expected fields are missing
If a record in the viewer is clicked on that has a `uid` and a `community_id`, but missing `ts` or `duration`, the app will crash if the detail pane is opened.

I was able to reproduce this bug by running this query, then opening the detail pane:

```
_path=conn | cut _path, uid, community_id, ts
```
image

Then if I click on the first result and open the detail pane: 

image

jameskerr:
Fixes #1475

This PR adds a check for the "ts" and "duration" fields before trying to access them. This is done when fetching data for the waterfall. We can't assume that data has a certain shape anymore, now that we can accept multiple sources.

I also added an error boundary around the log detail pane. If an error like this occurs again in the detail pane, the rest of the app will still be usable. Here is what it looks like:

Screen Shot 2021-03-03 at 3 50 45 PM

### DIFF
--- a/src/js/components/RightPane.tsx
+++ b/src/js/components/RightPane.tsx
@@ -1,27 +1,27 @@
-import {connect} from "react-redux"
+import DetailPane from "app/detail/Pane"
 import React from "react"
-
-import {DispatchProps} from "../state/types"
-import {Space} from "../state/Spaces/types"
-import {XRightPaneExpander} from "./RightPaneExpander"
+import {connect} from "react-redux"
+import {zng} from "zealot"
 import {openLogDetailsWindow} from "../flows/openLogDetailsWindow"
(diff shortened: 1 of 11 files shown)


src/js/components/RightPane.tsx
```diff
@@ -85,7 +85,9 @@ export default class RightPane extends React.Component<Props, S> {
           </PaneHeader>
         )}
         <PaneBody>
-          <DetailPane />
+          <AppErrorBoundary>
```
mason-fish, line 88: Awesome, good call 👍

mason-fish (approved):
Nice work!

PR #1496, opened 2021-03-05
Update join cookbook (for merge right before next Brim release)

philrz:
Two changes have happened since the last tagged GA Brim release (`v0.24.0`) that affect the "join" cookbook:

1.  The `join` processor now supports "inner", "left", and "right" joins, with the default switching from "left" to "inner"
2.  The `zq -I` option is now used to include Z scripts from files rather than `zq -z`

Here I've updated the article to account for these changes, and made a few additional clarifications along the way. Since I had to re-run all the examples to test it and get updated outputs, the `uid` values, screenshots, etc. all change.

In an attempt to keep the article in sync with what's GA, once this is approved, I intend to hold off merging it until right before I assemble the next release.

(`join` has changed since this was written and might change more before we next release, so I'm going to hold this one in draft state for now...)


philrz:
This PR no longer represents the way I'd want to introduce `join` to users. While the article that works for the Zed language that shipped with Brim `v0.24.0` still exists at [https://github.com/brimdata/brim/wiki/Joining-Data](https://github.com/brimdata/brim/wiki/Joining-Data),

1. The language has evolved still further in the time since this draft PR sought to update it
2. I've since come to see that that prior `join` article was not very good (e.g., too much leading text on data prep!)
3. We've since agreed that we want to trend toward more generic examples rather than so much emphasis on security/networking/Zeek

Therefore, now that `join` is further along, my preference is to document it alongside the other Zed operators, which is up for review in brimdata/zed#2836. Once that's merged, I'll put up a very small cookbook article for the Brim wiki that shows how for Brim `v0.25.0`, multi-Pool joins like the ones in the Zed doc could be performed outside the app using `zapi` and then the joined results pushed back into a Pool, if desired.

As for this one, I'm closing without merging.

Issue #1376, opened 2021-01-22
Windows support for Brim auth flow


alfred-landrum:
@mason-fish discovered that the easy 'open-url' Electron handler only works on macOS, which means we need to research how to handle the "brim://" protocol on Windows platforms.

PR #1508, opened 2021-03-16
Electron builder transplant

### LINKED ISSUE #1487: Replace Brim release process with "electron-builder"
We currently use a modular solution for the release process which includes `electron-packager` (packaging), `electron-winstaller` (windows installer), `electron-installer-dmg` (mac), `electron-installer-debian`, `electron-installer-redhat`, `electron-installer-zip`, `electron-notorize` (for signing). The packager we currently use does not facilitate registering custom app protocols (`brim://`) for all OSs, just MacOS, which is now a requirement for our authentication mechanism.

Instead of seeking additional modular solutions, let's use [electron-builder](https://www.electron.build/) for this whole process instead. It describes itself as "A complete solution to package and build a ready for distribution Electron app for macOS, Windows and Linux with “auto update” support out of the box". After spending some time seeing how feasible it would be to make this transplant, I've found that the transition should be relatively straightforward.

The primary cost comes with the fact that iteration speed is slow since it depends on building a release with each configuration attempt, as well as that we would need to verify several of the components that this new mechanism would be managing. The following is a list of those components:

**macOS**
- [x] Build release
- [x] Sign release (includes entitlements necessary for suricata updates)
- [x] Notorize release 
- [x] Install custom brim protocol 
- [x] Verify autoupdate

**Windows**
- [x] Build release
- [x] Sign release
- [x] Install custom brim protocol 
- [ ] Verify autoupdate

**Linux**
- [x] Build release
   - [x] Create installer for `debian`
   - [x] Create installer for `redhat`
- [x] Install custom brim protocol 
- [x] Verify autoupdate

mason-fish:
fixes #1487 #1377 #1376 

Uses `electron-builder` as the primary build tool. Aims to achieve parity with the existing process, but there are a couple key differences:

* The Windows installer is now NSIS instead of squirrel.Windows. This is a limitation of `electron-builder`, where squirrel.Windows support is deprecated, but also the autoupdater that `electron-builder` provides only supports NSIS-based installers. The newer installer also gives us a nice installation progress bar. Unfortunately, this also will cause existing releases to fail to update to any newer NSIS-based releases, though NSIS-based releases will autoupdate to future NSIS-based releases successfully.
* The app is now bundled with electron's recommended archive format, .asar, which appears to be the standard practice for electron based apps and gives some small performance boosts to file loading in Windows.
* Custom protocol handling now spans across all OSs we support (this was the original motivation for this whole builder swap)
* The autoupdater is also swapped as part of this change. This [new system](https://www.electron.build/auto-update) will query github directly, so no more autoupdate server is necessary, offers more control over the entire process with more events to listen/respond to, and supports all platforms (the existing [native electron autoupdater](https://www.electronjs.org/docs/api/auto-updater#autoupdater) only supports MacOS and Windows).


electron-builder.json
```diff
@@ -0,0 +1,50 @@
...
+      "dmg",
+      "zip"
+    ],
+    "binaries": [
+      "zdeps/zeek/lib/zeek/plugins/Corelight_CommunityID/lib/Corelight-CommunityID.darwin-x86_64.so"
+    ]
+  },
+  "nsis": {
+    "artifactName": "${productName}-Setup-${version}.${ext}",
+    "deleteAppDataOnUninstall": true
```
philrz, line 36: I think we can revert this most recent commit. Just to write down the wider context for the PR history, I'd made this observation on Slack to @mason-fish:

> I've not found this to cause any problems, but just sharing an observation: With the new builds, after doing an uninstall on Windows, a directory `AppData\Local\brim-updater` is left behind with the single file `installer.exe` in it.

He added this `"deleteAppDataOnUninstall": true` in response to that. But I just confirmed in my testing fork that this setting does _not_ trigger the removal of that file/directory. Rather, I saw it triggered the removal of the `AppData\Local\Brim` directory, which is where the user's `appData.json` lives along with the `data` directory that holds all the ZNG for the Spaces and so forth. Our older builds left that `Roaming` directory intact and I think we might as well continue to behave this way since it allows a user to uninstall/reinstall and still see all their data (I just confirmed that works with the last GA release, tag `v0.24.0`). If we ever wanted to get fancy perhaps one day we could find a way to prompt the user and delete its contents on-demand during uninstall if that's what they want. But for now to keep things simple I'd say we should revert that commit so the `Roaming` dir is left behind after uninstall.

And back to the original point, I'm not too concerned about that `brim-updater` directory. The old system also left behind droppings after uninstall in the form of an `AppData\Local\Brim` directory that contained a file called `.dead` and another file `Update.exe`. So, there's precedent here without ever having heard of it causing a problem. I just like to document a sense of what's typical so that way I know to watch out for variations from this as a sign that something's wrong.

ppl/zqd/zqd.ts
```diff
@@ -10,7 +10,10 @@ import * as cmd from "src/js/stdlib/cmd"
 import electronIsDev from "src/js/electron/isDev"
 
 // Paths for the zqd and zeek programs.
-const zdepsDirectory = join(app.getAppPath(), "zdeps")
+const zdepsDirectory = join(
+  app.getAppPath().replace("app.asar", "app.asar.unpacked"),
+  "zdeps"
+)
```
jameskerr, line 16: Curious about this. Maybe a short comment above describing why this is needed?

src/js/electron/autoUpdater.ts
```diff
@@ -82,13 +24,7 @@ export async function setupAutoUpdater() {
...
 
   setUpdateRepeater(() => {
-    getLatestVersion()
-      .then((latestVersion) => {
-        if (semver.gte(app.getVersion(), latestVersion)) return
-
-        autoUpdater.checkForUpdates()
-      })
-      .catch((err) => log.error(err))
+    autoUpdater.checkForUpdates()
```
jameskerr, line 27: Wow, so much less code.

jameskerr (approved):
This makes the release code much simpler to reason about. I'm very glad we'll be moving to this. 

From a code perspective, this is approved, but it looks like Phil is doing further verification. Let's wait to merge until he approves.

philrz:
> From a code perspective, this is approved, but it looks like Phil is doing further verification. Let's wait to merge until he approves.

Thanks @jameskerr. Yeah, I've got more testing I want to do. With @mason-fish's approval, I just merged some changes into this branch that got the nightly builds working in Actions. But now that I've got those basics working reliably, I want to do some tests with tagged releases, auto-update, and also confirm what happens when it's installed on top of an older install, so I'll proceed with that in yet another fork. I'll circle back here with updates.

.github/workflows/macos-release-candidate.yml
```diff
@@ -89,7 +110,7 @@ jobs:
     - name: Upload release artifacts to Google Cloud Storage bucket
       run: |
         gsutil rm -rf ${{ env.BUCKET_PATH }} || true
-        gsutil cp -r dist/installers ${{ env.BUCKET_PATH }}
+        gsutil cp dist/installers/*.dmg* dist/installers/latest-mac.yml ${{ env.BUCKET_PATH }}
```
philrz, line 113:
```suggestion
        gsutil cp dist/installers/Brim-*.dmg* dist/installers/Brim-*-mac.zip dist/installers/latest-mac.yml ${{ env.BUCKET_PATH }}
```

This this is a change I needed to make in my fork to get auto-updates working on macOS. I'd previously not realized the significance of the `.zip` artifact and it only became clear when auto-update failed and I saw the message in the log indicating it was pining for that file.

While I'm at it, I made the glob for the DMG-centric artifacts a little more strict, just for kicks.

philrz (approved):
I did some thorough testing of the permutations of updates between the old/new systems and opened an Epic #1563 that suggests we invest some cycles into dealing with the findings. However, they all seem like areas that we can chip away at _after_ the new build system goes in. I'm hereby in favor of getting this PR merged so we can continue with follow-on work that affects the builds, such as continuing the unbundling of the Zeek/Suricata components.

PR #1510, opened 2021-03-17
sync with zq uniform-ast

mccanne:
I just merged the zq `uniform-ast` branch and this is passing tests and ready to go...


mccanne:
By the way, I don't think the elements marked with `???` are used anymore...

jameskerr (approved):

PR #1520, opened 2021-03-18
Fix link to immerjs.github.io

philrz:
The markdown link checker caught this as broken last night. It's still broken now and I clicked around and found what I think is a new home for the same page we used to link to.


mason-fish (approved):
Ah nice I had caught this as well and included the fix in the builder PR, not sure when we will want to merge that though so this is better 👍

Issue #1529, opened 2021-03-23
Brim CI failure: "TypeError: Cannot read property 'name' of undefined"


philrz:
[https://github.com/brimsec/brim/actions/runs/677403237](https://github.com/brimsec/brim/actions/runs/677403237) was a run of the Brim CI that was triggered by the advance of the `zq` pointer. The run was at Brim commit 31bbcd0 attempting to run with `zq` commit brimsec/zq@50bd002 (associated with brimsec/zq#2386). The failure was in the "histogram deep inspection" test in `itest/tests/ingest.test.ts` and the failure was attached in this screenshot:

![image](https://user-images.githubusercontent.com/5934157/112200542-92f43800-8bcc-11eb-829b-2613ed3c4349.png)

I've since performed a local run of the Brim CI using that combination of commits and was not able to reproduce the failure. A subsequent `zq` commit triggered a re-run of the CI and it passed, so the `zq` pointer has since been successfully advanced. This seems to indicate that the problem might be intermittent.

jameskerr:
I was able to reproduce this by navigating to a workspace that did not exists. In the PR that fixes this, I'll add a global function called `navTo(url)`. I'll add an integration test that does this too.

PR #1533, opened 2021-03-24
Missing workspace bug

jameskerr:
Added a global function called `navTo(path)`. You can use this in the dev tools to go to a specific url. 

If you open the dev tools and type `navTo("/workspaces/does-not-exist")`, then you will reproduce the bug reported at #1529.

A simple null check fixed it. An integration test has been added to check for this now.

### DIFF
--- a/app/workspaces/show.tsx
+++ b/app/workspaces/show.tsx
@@ -21,6 +21,12 @@ const SpinnerWrap = styled.div`
   justify-content: center;
 `
 
+function GetWorkspace({children}) {
+  const workspace = useSelector(Current.getWorkspace)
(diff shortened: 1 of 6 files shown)


app/workspaces/show.tsx
```diff
@@ -21,6 +21,12 @@ const SpinnerWrap = styled.div`
   justify-content: center;
 `
 
+function GetWorkspace({children}) {
+  const workspace = useSelector(Current.getWorkspace)
+  if (!workspace) return <Redirect to="/workspaces" />
```
mason-fish, line 26: 👍

mason-fish (approved):
Looks good!

PR #1535, opened 2021-03-25
strip commas for count type

### LINKED ISSUE #1113: Strip commas when putting numbers into paste buffer
As of Brim commit `6f7c3be`, larger numbers are printed with commas in them.

image

While this is convenient for readability in the UI, if the user single-clicks then hits a hotkey to copy its contents into a paste buffer, when pasted it comes out with the comma intact, i.e. `10,411`. Since the user might need to paste this into some tool that expects a true number (that is, a tool not smart enough to detect/ignore the comma) it would be a nice convenience if we could strip the comma as it's going into the paste buffer.

mason-fish:
fixes #1113 

### DIFF
--- a/ppl/menus/detailFieldContextMenu.ts
+++ b/ppl/menus/detailFieldContextMenu.ts
@@ -62,7 +62,7 @@ export default function detailFieldContextMenu(
       }),
       detailMenuActions.countBy.menuItem([fieldData], {enabled: !isGroupBy}),
       menu.separator(),
-      {role: "copy"},
+      detailMenuActions.copy.menuItem([fieldData]),
       menu.separator(),
       detailMenuActions.sortAsc.menuItem([fieldData], {enabled: hasCol}),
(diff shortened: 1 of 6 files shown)


jameskerr (commented):
It looks like this takes commas out of the numbers entirely. I think the ticket wants to retain the comma's in the app, but when you copy the value, only copy the number.

philrz:
@jameskerr: Yeah, for display purposes, the commas still seem helpful (though I imagine European users would appreciate a user preference at some point to use dots instead :wink: ).

jameskerr (approved):
An improvement in my opinion. The right click-> copy option will now copy the underlying value, not the formatted value. 

As we discussed, there are still some holes. The app menu => Edit => Copy and the normal CMD+C shortcuts will still copy the formatted value.

Since the data viewer is due for an upgrade, we will consider filling in the holes when that upgrade takes place.

PR #1548, opened 2021-03-29
Make import page generic

### LINKED ISSUE #1537: Generic Import Page
Remove mentions of security specific formats on the import page. Make the list of supported formats the following:  zeek, json, zson, zng, zst, parquet, csv, tsv

jameskerr:
Now looks like this:

image

Closes #1537


mason-fish (approved):
Looks good, we can do more on this page when we have a better idea on the types of data we will focus on moving forward

PR #1553, opened 2021-03-31
Fix Responsive Toolbar Buttons

### LINKED ISSUE #1538: Remove Duration in Space Title
The title of the space on the search page contains a section for the time duration. Remove this if the lake key is not a time type.

jameskerr:
Closes #1538 

The duration under the space title will now not appear if there is no duration (it actually already did this). The focus of this PR is actually some improvements to the page header.

Moved the toolbar buttons up on the same row as the space title. It collapses sensibly as you make the window smaller. It first shrinks the title until the min-width of 150px, then it starts moving toolbar buttons into the context menu. I also removed little blue icon from the title.

image


itest/lib/log.ts
```diff
@@ -13,8 +13,6 @@ export const LOG = createLogger({
     })
   ),
   transports: [
-    new transports.Console({level: "info"}), // No idea how to make this runtime-configurable via command line. This has
```
mason-fish, line 16: Are you removing this because "info" is already the default log level?

package.json
```diff
@@ -14,7 +14,7 @@
     "start": "node scripts/start",
     "test": "jest",
     "test:api": "node scripts/test/api",
-    "itest": "jest --config=itest/config.json --runInBand --verbose",
```
mason-fish, line 17: Will this still show the same detail/verbosity when tests fail?

mason-fish (approved):

itest/lib/log.ts
jameskerr: This was just making the output on CI annoyingly long. The detailed info is in the file still, but doesn't clutter the console.

package.json
jameskerr: Yes, it shows the failures like a normal jest runner.

PR #1556, opened 2021-04-01
change zar to zed lake

nwt:
The "zed lake" command replaces zar in brimdata/zed#2463.


philrz (approved):
I normally don't weigh in on topics that touch JavaScript code since I'm not a Dev, but this all seems quite intuitive and it makes the CI pass, so, 👍 .

PR #1571, opened 2021-04-14
Only publish wiki when docs/ directory changes

philrz:
When setting up the Brimcap wiki in brimdata/brimcap#30, @mattnibs made the helpful suggestion that the Action only be triggered to run when the contents of the `docs/` directory changes. Having made the improvement there, I'm circling back to do the same here.


jameskerr (approved):
Thanks Phil!

PR #1573, opened 2021-04-15
Basic plugin plumbing + brimcap plugin v0

### LINKED ISSUE #1539: Conditional Packets Button
For the next release, the "security module" will be considered "installed" if there is the presence of a `brimcap` executable on the user's system.

If the security module is installed, show the packets button like it is now. If not, remove it from the toolbar. 

- [ ] The packets button in the search window
- [ ] The packets button in the detail window

### LINKED ISSUE #1558: Clicking shark fin should exec "brimcap launch"

### LINKED ISSUE #1559: Dragging a pcap into the app should exec "brimcap load"

mason-fish:
fixes #1539
fixes #1558 
fixes #1559 

Includes the basic shape that I thought we could start out with as we step more in the direction of a plugin architecture. As the first implementation I have included the `brimcap` plugin to manage pcaps for the app. This includes importing pcaps into the app as well as launching wireshark when appropriate.

Plugins live in the `plugin` directory. For now, those plugins must be a directory with an index.ts entry point, and inside that entry point an `activate()` and `deactivate()` function must be exported so the app knows how to run that code. The app will pass in the `BrimApi` to each activate function so that the plugins gain access to modify app.

This first version of the `BrimApi` has a commandRegistry/eventEmitter, a couple ui-specific apis for modifying visual components, and also has a registry for "loaders". Loaders are used when files are dragged into the app and when they match with the file type(s) provided. NOTE: I actually pulled out the zqd /log endpoint logic into it's own loader to handle zeek logs here but if we wanted to keep as a "default loader" that lives app-side instead of plugin-side then I can move that back in.

There are 2 bugs I already know about that I would like to focus on in a followup PR:
1) since we are spawning brimcap now instead of making an http request, refreshing the app produces orphan processes since we lose the child process reference.
2) right now our logic for the wireshark button is basically if (brimcapExists && isConnRecord(data)) but this means that the button will be still active for zeek logs. We previously handled this with a pcap_support: true property on a space provided by zqd, but zqd is dropping that so we will have to manage that ourselves


plugins/brimcap/brimcap-plugin.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  private async downloadPcap(log: zng.Record) {
+    const searchOpts = this.logToSearchOpts(log)
+
+    const searchAndOpen = async () => {
+      await this.cli.search(searchOpts)
+      return open(searchOpts.write, {newWindow: true})
+    }
+
+    toast.promise(
```
mason-fish, line 203: whoops, this should come from the api. I'll update this in the next commit with the ingest handling

jameskerr (approved):
Nice! This is very easy to follow and seems like the start of something really powerful.

I have 2 comments that we can discuss, but they don't need to block it.

jameskerr:
I also loved the comments you added, but I wanted them to wrap at 80 chars so I didn't need to horizontal scroll. Maybe there is a automated way we could do that.

Maybe this works for Webstorm? 

Turn on option:
Preferences | Code style | JavaScript | Wrappings | Wrap at right margin

A VSCODE extension: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

plugins/brimcap/brimcap-cli.ts
```diff
@@ -47,34 +54,41 @@ const OPTION_NAME_MAP = {
 }
 
 export default class BrimcapCLI {
-  private rootCmd = "brimcap"
+  private binPath: string
 
-  constructor() {}
+  constructor(binPath: string) {
+    this.binPath = binPath
+  }
```
jameskerr: No biggie, but this could be:
```ts
constructor(private binPath: string) {}
```

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -214,11 +217,78 @@ export default class BrimcapPlugin {
     )
   }
 
+  private setupLoader() {
```
jameskerr, line 221: Do you like the word loader over importer? Loader in the front end context makes me thing of spinning wheels. Importer might be less over"loaded".  I don't feel too strongly about it though.

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -214,11 +217,78 @@ export default class BrimcapPlugin {
...
+    const load = async (
+      params: IngestParams & {spaceId: string},
+      onProgressUpdate: (value: number | null) => void,
+      onWarning: (warning: string) => void,
+      onDetailUpdate: () => void
+    ): Promise<void> => {
+      const {fileListData, name} = params
+      if (fileListData.length > 1)
+        throw new Error("Only one pcap can be opened at a time.")
```
jameskerr, line 230: Is this still true with brimcap?

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -214,11 +217,78 @@ export default class BrimcapPlugin {
     )
   }
 
+  private setupLoader() {
+    const load = async (
+      params: IngestParams & {spaceId: string},
+      onProgressUpdate: (value: number | null) => void,
+      onWarning: (warning: string) => void,
+      onDetailUpdate: () => void
```
jameskerr, line 226: I think onDetailUpdate should be implied when you call onProgressUpdate.

src/js/api/registries.ts
```diff
@@ -0,0 +1,56 @@
...
+  execute(command: string, ...args: any[]): boolean {
+    return this.commandRegistry.emit(command, ...args)
+  }
+
+  list(): string[] {
+    return this.commandRegistry.eventNames() as string[]
+  }
+}
+
+interface Loader {
```
jameskerr, line 29: Nice! I think this would be better in it's own file.

src/js/flows/executeCommand.ts
```diff
@@ -0,0 +1,9 @@
+import {Thunk} from "../state/types"
+
+export const executeCommand = (command: string, ...args: any[]): Thunk => (
+  dispatch,
+  _gs,
+  {api}
+) => {
+  api.commands.execute(command, args)
```
jameskerr, line 8: 👍

src/js/flows/ingestFiles.ts
```diff
@@ -110,24 +112,46 @@ const unregisterHandler = (dispatch, id) => ({
...
+    if (!loaders || loaders.length === 0) {
+      throw new Error(
+        `No registered loaders match the provided file type: ${filesType}`
+      )
+    }
+
+    // only supporting one loader at this time
+    const l = loaders[0]
+    try {
+      await l.load(params, onProgressUpdate, onWarning, onDetailUpdate)
```
jameskerr, line 149: 🔥

src/js/initializers/initialize.ts
```diff
@@ -1,16 +1,17 @@
...
 import initPlugins from "./initPlugins"
 import initStore from "./initStore"
 import initWorkspaceParams from "./initWorkspaceParams"
+import BrimApi from "../api"
 
 export default async function initialize() {
-  const store = await initStore()
+  const api = new BrimApi()
+  const store = await initStore(api)
+  api.setStoreArgs(store.dispatch, store.getState)
```
jameskerr, line 13: 👍

jameskerr (commented):
Nice! A few nits, questions, and suggestions. Nothing blocking.

plugins/brimcap/brimcap-plugin.ts
mason-fish: looks to be the case: https://github.com/brimdata/brimcap/blob/f95d829bfd0560f4aa53590ee30380ac123520c1/cmd/brimcap/load/command.go#L97

mccanne:
Nice work!

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -0,0 +1,312 @@
...
+import fsExtra, {pathExistsSync} from "fs-extra"
+import errors from "src/js/errors"
+import {reactElementProps} from "../../src/js/test/integration"
+
+export default class BrimcapPlugin {
+  private cli: BrimcapCLI
+  // currentData represents the data detail currently seen in the Brim detail pane/window
+  private currentConn = null
+  // selectedData represents the data detail currently selected and highlighted in the search viewer
+  private selectedConn = null
```
nwt, line 20:
```suggestion
  // currentConn represents the data detail currently seen in the Brim detail pane/window
  private currentConn = null
  // selectedConn represents the data detail currently selected and highlighted in the search viewer
  private selectedConn = null
```

plugins/brimcap/brimcap-plugin.ts
mason-fish: I had originally paired them together but it produced some unwanted re-renders since I couldn't replicate the exact logic we use in https://github.com/brimdata/brim/blob/main/src/js/flows/ingestFiles.ts#L161-L195 without the ability to update details *without* updating progress.

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -0,0 +1,312 @@
...
+
+    this.brimcapDataRoot = path.join(dataRoot, "brimcap-root")
+  }
+
+  public init() {
+    fsExtra.ensureDirSync(this.brimcapDataRoot)
+
+    if (!pathExistsSync(this.brimcapBinPath)) {
+      console.error(
+        "The brimcap plugin requires the 'brimcap' CLI tool to also be installed. Please download it here: https://github.com/brimdata/brimcap/releases"
```
nwt, line 54: Nit: This message is no longer quite right now that we're bundling brimcap. Since the error now indicates a packaging problem, maybe suggest that the user create a brimdata/brim issue or otherwise reach out for help?

plugins/brimcap/brimcap-plugin.ts
mason-fish: I could go either way, I do see your point about the semantic collision there. Part of why I like `load()` is just that it has an obvious rollback verb, "unload", whereas "unimport", well, has a red squiggly under it as I type 😄 

Mind if I keep it as is for now but flag this for more discussion when we chat api design next? I anticipate there will be a lot more changes like this that we may want to make

plugins/brimcap/brimcap-cli.ts
```diff
@@ -0,0 +1,87 @@
...
+  suricataStderr?: string
+  suricataStdout?: string
+  zeek?: boolean
+  zeekStdout?: string
+  zeekStderr?: string
+}
+
+export interface launchOptions extends packetOptions {
+  root: string
+}
```
nwt, line 29: NIt: `brimcap` no longer has the `launch` command.

mattnibs:
Are the integration tests relying on brimcap and not the old pcap post endpoint?

nwt:
@mattnibs: They are!

PR #1575, opened 2021-04-15
Kabob Case Typescript Files

jameskerr:
Renamed all camelCase source files to kabob-case and update their references.

There are still some directories that are title case or camel case, but they are few. Another PR will finish those off.

- [ ] It looks like some files were duplicated because they were the same name just lowercased. Need to fix that


jameskerr:
I'm going to close this, but keep the branch. I'll give it another go after the 25 release.

PR #1580, opened 2021-04-19
remove JSON type config support

nwt:
Zed shaper scripts replace JSON type configs, and support for the latter
will be removed from zqd.


nwt:
This should go in before brimdata/zed#2587 to keep CI green.

jameskerr (approved):
Dang! Look at all that deleted code. What a joy.

nwt:
You love to see it.

PR #1584, opened 2021-04-21
add brimcap to zdeps

### LINKED ISSUE #1574: Add brimcap as a dependency
To make progress on the integration of Brimcap with the Brim app, we reached consensus on a plan is similar to what's currently done for Zed: Add a brimcap dependency to the `package.json` and make it such that the Brimcap binaries at that commit are compiled and added when `npm install` is run.

nwt:
Brimcap includes the Brim builds of Suricata and Zeek.

Depends on brimdata/brimcap#47.

Closes #1574.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7873,6 +7873,11 @@
         }
       }
     },
+    "brimcap": {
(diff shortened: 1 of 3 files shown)


mason-fish (approved):
This looks good, disregard my comment in sync about seeing an error building the release: false alarm

Issue #1567, opened 2021-04-13
Provide clickable migration of file store ZNG to Zed lakes


philrz:
We expect storage in the new Brim experience to be focused exclusively on Zed lakes. However, users may still have existing "file store" Spaces from the old Brim experience that they may want to continue to work with. At a minimum, we could point users to an article that describes the change, tells them where on the filesystem they'll find the `all.zng` files for their old Spaces, and describes `zapi` command lines that could be used to manually re-import those into Zed lakes. However, when discussing as a group, there was some consensus (@jameskerr in particular seemed on-board) toward some clickable helpers, such that if the user clicks a "file store" Space, we pop up notification regarding the new file format (perhaps link to a Support article that describes it in more detail) and provides a push-button option that would invoke the necessary `zapi post` command line to do it for them, perhaps using the same Space name as the old one with `-lake` appended to it. The user should be able to still right-click and Delete the old Space once they're satisfied with how the data looks in the new Zed lake representation.

In the same pop-up we can also explain that if the logs in the Space were originally created from a pcap, they'll lose the ability to pivot from the logs in the Zed lake back to the pcap slices, so if that's something they want to preserve they'll want to re-import the pcap via Brimcap.

philrz:
We had more group discussion today on other ways to approach this. Another possibility we discussed was having the migration commands constructed by Brimcap, and we could even go as far as having them be `brimcap load` commands that re-import an original pcap to preserve the coupling between the logs & the pcap so the user can keep pivoting between them.

@mattnibs and @philrz will talk more about the trade-offs before we pick & implement an approach.

philrz:
[**@mattnibs**](https://github.com/mattnibs), [**@jameskerr**](https://github.com/jameskerr) and I had another discussion regarding the approach to the migration. The current plan goes something like this:

1.  The user launches the new app for the first time
2.  If the `data/spaces/` directory is non-empty, a modal appears
3.  The modal explains in brief that their existing Spaces would need to be migrated to Pools in order to keep querying them, linking to an article in the wiki that has more detail  
    (By linking to an article, we can populate/change content there over time as necessary, depending on struggles/bugs we may encounter as users start working with it. It also gives us freedom to go into as much detail as we want about what's special about Lakes, and so forth.)
4.  The modal offers the user **Ok** and **Cancel** buttons. If they choose:
[...]

PR #1587, opened 2021-04-26
Add initial doc regarding Space migration

philrz:
Part of the plan for #1567 includes an article we can link to from our Space migration UX. While we're hopeful the migration tools should be seamless enough that users shouldn't need to read any supplemental materials to use it, some things I hope to accomplish by having an article:

1.  Educate the curious users on Zed Lakes & Data Pools to provide more detail (and links to even more detail) on why we've switched to this approach, how it benefits them, etc.
2.  We can continue change and/or add to the article if we notice user confusion or bugs as users start to upgrade & migrate

There's a bit of a chicken/egg snag where the article needs to exist before we can link to it from the tools, so here I've started a placeholder article with a stab at some minimal content. As the migration tooling comes together, I'll plan to expand it with screenshots and such.

While I took a crack at summarizing some of the goodness of Lakes/Pools, please don't mistaken this for official marketing copy. I fully expect we're going to to have many more words to say on these topics in the web site, blog posts, etc., and as that wording comes together I fully expect the summary can be revised accordingly. That said, if you think I badly misrepresented something or left out something essential, I'm all ears.

### DIFF
--- a/docs/Migration-of-Spaces.md
+++ b/docs/Migration-of-Spaces.md
@@ -5,17 +5,17 @@ now be stored in "Data Pools" in a "Zed Lake" rather than "Spaces" as they were
 previously.
 
 A [design document](https://github.com/brimdata/zed/blob/main/docs/lake/design.md)
-provides a thorough overview of pools and how they work. In brief, the use of
-pools ultimately enables new functionality that was not previously possible in
+provides a thorough overview of Pools and how they work. In brief, the use of
+Pools ultimately enables new functionality that was not previously possible in
 Brim, including:
 
-* Data may be incrementally added to, or deleted from, a pool.
-* Zed queries/analytics can make use of data stored across multiple pools.
+* Data may be incrementally added to, or deleted from, a Pool.
+* Zed queries/analytics can make use of data stored across multiple Pools.
 * Search indexes can be created to accelerate query performance.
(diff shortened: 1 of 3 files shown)


docs/Migration-of-Spaces.md
```diff
@@ -0,0 +1,21 @@
...
+Brim, including:
+
+* Data may be incrementally added to, or deleted from, a pool.
+* Zed queries/analytics can make use of data stored across multiple pools.
+* Search indexes can be created to accelerate query performance.
+* Pre-computed derived analytics can be leveraged to accelerate computations on stored data.
+
+When upgrading to GA Brim release `v0.25.0` or newer, any data you had stored
+in Spaces that you wish to continue accessing will need to be migrated to
+pools. The tooling to perform this migration is still under development. This
```
jameskerr, line 19: I see "Spaces" was capitalized but "pools" was lower case. Not sure which is correct but they seem on the same level.

jameskerr (approved):
I'll link to this in the UI when I start building it.

docs/Migration-of-Spaces.md
philrz: Good point @jameskerr. I just pushed a commit to capitalize the "Pool" references also. Merging!

PR #1589, opened 2021-04-27
Use zed lake

mattnibs:


src/js/state/rootReducer.ts
```diff
@@ -27,7 +27,7 @@ export default combineReducers<any, any>({
   tabs: Tabs.reducer,
   investigation: Investigation.reducer,
   view: View.reducer,
-  spaces: Spaces.reducer,
+  pools: Pools.reducer,
```
mason-fish, line 30: we will also need to write a migration here to change existing states with `spaces` -> `pools`

src/js/state/Pools/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  | POOLS_INGEST_WARNING_CLEAR
+  | POOLS_REMOVE
+  | POOLS_WORKSPACE_REMOVE
+export type Pool = {
+  name: string
+  id: string
+  min_time: Ts
+  max_time: Ts
+  ingest: PoolIngest
+  size: number
```
mason-fish, line 26: As part of the migration we can also delete existing `parent_id`s and `storage_kind`

mason-fish (changes requested):
Thanks for helping out with this, looks great! The only thing left that I can spot is that we should need a state migration with the reducer name change. It should look very similar to https://github.com/brimdata/brim/blob/af5004190372e2097b7ea0a2599faeed785f0a5b/src/js/state/migrations/202101210823_renameToWorkspace.ts#L19-L26

src/js/state/rootReducer.ts
mattnibs: You actually don't want to do this, the old spaces should be migrated via the brimcap migrate command (in progress).

philrz: Good catch. Folks can read up on the current plan at #1567 and brimdata/brimcap#58.

mason-fish (approved):
I've dropped the spaces state in a migration, I think this is ready to go now

jameskerr:
Didn't see this until just now, I was in a zjson hole for so long. Thanks @mason-fish and @mattnibs !

PR #1591, opened 2021-04-28
remove pcap code

mason-fish:
* Removes both pcap endpoints from zealot, along with corresponding tests
* Removes Packets reducer from store and includes a migration for past appStates
* Removes DownloadProgress component, this is handled as a toast through the app's plugin api and the brimcap plugin now
* Drops the `pcap_support` field, the backend will not manage that for spaces/pools anymore


nwt (approved):
:shipit:

PR #1614, opened 2021-05-06
Fix brimcap zombies

### LINKED ISSUE #1612: Reload during "brimcap load" produces a zombie process
Repro is with the nightly build [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-806860c4.0/macos/Brim-0.25.0-prerelease-806860c4.0.dmg](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-806860c4.0/macos/Brim-0.25.0-prerelease-806860c4.0.dmg) (so, Brim commit 806860c, which talks to Zed commit `dc82704`).

As shown in the attached video, if we start importing a pcap, the `brimcap load` that's launched to complete the task becomes a zombie process if the user selects **View > Reload** while it's still in progress.

https://user-images.githubusercontent.com/5934157/117220512-3a63ad80-adbc-11eb-9208-e89941d1f7b6.mp4

mason-fish:
fixes #1612 

Refreshing the app would close the stdio streams to/from any launched `brimcap` processes which caused the process to terminate, however the app/parent process would loose sight of the spawned process and not always be around to wait for the child exit, causing zombie processes. Now we explicitly kill and await the exit of all spawned processes on reload and quit.

### DIFF
--- a/plugins/brimcap/brimcap-cli.ts
+++ b/plugins/brimcap/brimcap-cli.ts
@@ -53,7 +53,7 @@ export default class BrimcapCLI {
   public load(pcapPath: string, opts: loadOptions): ChildProcess {
     const subCommandWithArgs = ["load", ...toCliOpts(opts), pcapPath]
 
-    return spawn(this.binPath, subCommandWithArgs)
+    return spawn(this.binPath, subCommandWithArgs, {detached: true})
   }
 
   public search(opts: searchOptions) {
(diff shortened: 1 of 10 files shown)


jameskerr (approved):
I like how this fix pretty easily fit into the plugin architecture! I'd suggest we move the concepts of "onReload" and "onClose" in the brim api object. Then we could avoid returning the plugin manager from the init call.

jameskerr:
Why don't you return the brim api itself from the initialize function. That object can have both the store and the plugin manager as properties.

mason-fish:
@jameskerr I don't think the brim api, ultimately, shouldn't make its `store` property public as it is today. That was only meant to be temporary until we flesh out the rest of the api surface enough to satisfy current and future plugins' or api consumers' needs. I also don't think the `pluginManager` would belong inside the api since it currently serves as a liaison between the app and the plugins, passing the api to each plugin so it can gain access to the app. If we put the manager inside the api then plugins would be able to control each other without any regulation.

PR #1617, opened 2021-05-06
notify on failures

### LINKED ISSUE #1616: Brimcap-plugin: "pcap_support" replacement
Issue: We no longer have the `space.pcap_support` data that was previously stored by zqd to indicate whether a pool contains data imported from a pcap file (and therefore has an underlying pcap for it's ingested data). We had been using this to disable the wireshark button for conn records imported directly from zeek logs. However, since the brimcap changes, it now remains enabled for those zeek logs.

Solution: For now, we will continue with this new, not-ideal behavior and notify users if there are no underlying packets *after* they click on the button.

Later on there might be more we can do, such as spawning `brimcap search` every time a log is selected (highlighted in the search results/viewer) to check if pcaps are supported, and then enable/disable the button accordingly. As it stands now, running that search automatically on each log selection change could become expensive, especially if the user quickly highlights several records in succession which each may result in a hefty brimcap operation that gets spawned solely for the purpose of checking if we should enable the wireshark button or not. If we can tweak the search command to offer an inexpensive "peak" mechanism used to check for pcap support then this could be a better solution.

mason-fish:
fixes #1616 

Provides more information to users for two error cases:

**Attempting to launch wireshark on an any `conn` record that has no 5-tuple information**
image

**Attempting to launch wireshark when no pcaps are found in the search**
image

### DIFF
--- a/plugins/brimcap/brimcap-plugin.ts
+++ b/plugins/brimcap/brimcap-plugin.ts
@@ -202,7 +202,9 @@ export default class BrimcapPlugin {
       searchOpts = this.logToSearchOpts(log)
     } catch (e) {
       console.error(e)
-      this.api.toast.error("Missing 5-tuple from log")
+      this.api.toast.error(
+        "Flow's 5-tuple and/or time span was not found in the connection record"
+      )
       return
     }
 
@@ -221,11 +223,11 @@ export default class BrimcapPlugin {
     this.api.toast.promise(
       searchAndOpen(),
       {
-        loading: "Preparing PCAP...",
-        success: "Preparation Complete",
+        loading: "Preparing pcap...",
+        success: "Preparation complete",
         error: (err) => {
           console.error(err)
-          return "Error Preparing PCAP: " + err.message
+          return "Error preparing pcap: " + err.message
         }
       },
       this.toastConfig
--- a/plugins/brimcap/brimcap-plugin.ts
+++ b/plugins/brimcap/brimcap-plugin.ts
@@ -186,7 +186,7 @@ export default class BrimcapPlugin {
     return {
       dstIp: log.get("id.resp_h").toString(),
       dstPort: log.get("id.resp_p").toString(),
-      duration: dur.isSet() ? dur.toString() : "0s",
+      duration: dur?.isSet() ? dur.toString() : "0s",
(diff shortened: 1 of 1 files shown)


plugins/brimcap/brimcap-plugin.ts
```diff
@@ -177,10 +177,10 @@ export default class BrimcapPlugin {
     // )
 
     const tsString = ts.toString()
-    const dur = log.get("duration") as zed.Duration
+    const dur = log.try("duration") as zed.Duration
```
jameskerr, line 180: Try can return null. In the returned object, check for null before calling `isSet()`.

jameskerr (changes requested):
Looks great. I have one change request about checking for null.

plugins/brimcap/brimcap-plugin.ts
mason-fish: I ended up doing just that down a bit further at https://github.com/brimdata/brim/pull/1617/files#diff-00e37cfd83eb8f0d73899b431f6d59bbad7c73d7bc1e8ea6acbf1eba935ee120R189

mason-fish: Oh wait I misunderstood, disregard

jameskerr (approved):

philrz:
Nit: @mccanne has voiced a preference for writing "pcap" rather than "PCAP" because it's not strictly an acronym. And he's the guy who came up with libpcap, so I guess that carries some weight. :wink:

philrz:
For the "`conn` record that has no 5-tuple information" one, I might suggest this wording instead:

> Flow's 5-tuple and/or time span was not found in the connection record

That is, in the current error message as written, just saying "log" doesn't give the user a whole lot of info to go on to understand if they're seeing a bug or they tried something that shouldn't have worked. I suppose if this came up a lot (which we don't expect it to) a Support article to link to would be justified (a more detailed explanation of how we index the pcap & then find stuff in it) but this'll do for now.

I also figured we might as well say "connection record" (as opposed to `conn`) in the abstract since it's theoretically possible we may enhance things in the future to use other summary logs to find the 5-tuple & time data (see https://github.com/brimdata/brimcap/pull/72#discussion_r626202994).

PR #1621, opened 2021-05-08
Space migration

jameskerr:
Will clean up and add a description...


src/js/components/BrimToaster.tsx
```diff
@@ -1,34 +1,9 @@
...
-
-  useEffect(() => {
-    if (isEmpty(toasts)) return
-
-    setLastToast(last(toasts))
-  }, [toasts])
-
   return (
-    <>
-      <HiddenToast {...toastLocator.props}>{lastToast?.message}</HiddenToast>
```
mason-fish, line 29: I think some of the integration tests are expecting this. We added this because CI needed something more consistent to spot as it ran those tests, is that no longer necessary?

app/legacy/space-migration/space-migrator.ts
```diff
@@ -0,0 +1,78 @@
...
+// zdeps/brimcap migrate -zqd="run/data/spaces" -root="run/data/brimcap-root"
+export default class SpaceMigrator {
+  process: ChildProcess | null
+
+  constructor(readonly srcDir: string, readonly destDir: string) {}
+
+  needMigration() {
+    if (fs.existsSync(this.srcDir)) {
+      const files = fs.readdirSync(this.srcDir)
+      return files.some((f) => f.startsWith("sp_"))
```
mattnibs, line 14: You should just have to check if this.srcDir exsits. When the brimcap migrate completes with no further spaces it deletes the `spaces` dir. IIRC the new lake is stored in a folder called "lake".

app/routes/app-wrapper.tsx
```diff
@@ -71,13 +70,13 @@ export default function AppWrapper({children}) {
...
       <Preferences />
       <Modals />
-      <AboutModal />
       <PoolModal />
       <ColumnsModal />
       <HTMLContextMenu />
       <BrimToaster />
-      <PopNotice />
       <BrimTooltip />
+      <SpaceMigration />
```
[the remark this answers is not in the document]
jameskerr: I think we'll have to include it forever, right? Or at least until there is a high chance that none of our users on are version 0.24.0.

I'm hoping that the plugin api would be expanded to handle all of this functionality. Then we can hide it away in a plugin forever.

PR #1622, opened 2021-05-10
Add fuse if format is CSV

### LINKED ISSUE #1530: Tack on "| fuse" when doing CSV Export
The changes in brimsec/zq#2402 will make it such that the `zq` layer will no longer automatically fuse data that's intended for output in CSV format. In a group discussion on March 23, 2021, there was consensus that once `zq` is behaving this way, we can then modify the Export CSV functionality in the app such that it tacks on on `| fuse` to the user's Z if not it's not already present, preserving the current UX for CSV export.

For users that export lots of CSV data and know it well enough to be confident that it would never need to be fused (or are otherwise content to have their exports halt & have to be re-run when necessary), we also discussed adding a user preference to control whether the `| fuse` is auto-appended. But we could open that as a follow-on issue for later.

jameskerr:
Closes #1530 

When exporting results, if the format is csv, we will always add on a ` | fuse` to the end of the their query.

I just tested this with a query that already has `| fuse` at the end and saw no difference. It doesn't matter how many `| fuse`s get added.

### DIFF
--- a/src/js/flows/exportResults.test.ts
+++ b/src/js/flows/exportResults.test.ts
@@ -1,0 +1,9 @@
+import TableColumns from "../models/TableColumns"
+import {prepareProgram} from "./exportResults"
+
+test("prepare program", async () => {
+  const columns = new TableColumns("yo", [])
+  const p = prepareProgram("csv", "*", columns)
+
+  expect(p).toBe("* | fuse")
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
💣

PR #1625, opened 2021-05-10
Import wording

jameskerr:
This is what we've got now:

image

PR #1626, opened 2021-05-10
Wiki article about importing non-auto-detect formats

philrz:
For linking to from the word "docs" in the improved screen proposed in #1625.


jameskerr (approved):
I like it.

PR #1627, opened 2021-05-10
Show warning when old brim exists on the system

### LINKED ISSUE #1566: Notify a Windows user if they've launched new Brim while old Brim is still installed
If a Windows user is running a Brim version based on the old build/packaging such as the last GA release `v0.24.0`, then they install a new release that's based on electron-builder, both the old and new versions will be installed side-by-side. This creates some potential confusion in the **Programs** lists, though at least the newer release _does_ have the version number shown.

The **Start** menu is a bit more troublesome, as the icons to launch the old & new look much the same.

Another hazard is that both of these are pointing at the same data storage in `%APPDATA%\Brim`, so if they were to accidentally launch the old version after the new version, there's the potential for problems due the versions stepping on each other: Incompatible app state, data storage changes for Spaces, etc.

We discussed this one as a group and agreed that since the the Windows app launches immediately after install, we could just have the new app look for the presence of the old app in its common filesystem location and pop up a notification that the old app should be uninstalled. I'll also plan to write a Support article in the wiki that describes this in more detail, and we can link to that from the notification as well. We can put the same guidance in the Release Notes in the GitHub Releases page and on the Downloads page on the web site as well.

jameskerr:
closes #1566

Here is the warning on Mac.

To determine if the old brim is on the system, it asks the electron app for the "appData" path, which according to the docs:

```
appData 
Per-user application data directory, which by default points to:

%APPDATA% on Windows
$XDG_CONFIG_HOME or ~/.config on Linux
~/Library/Application Support on macOS
```

From there, it looks in `\Local\Brim`. This is where I saw brim on my windows 10 VM after installing it fresh from our website at version 0.24.

The next brim release was put at `\Local\Programs\Brim`. 

After the user clicks ok, the new app will immediately quit.

### DIFF
--- a/src/js/electron/windows-pre-25.ts
+++ b/src/js/electron/windows-pre-25.ts
@@ -1,0 +1,20 @@
+import {app, dialog} from "electron"
+import {join} from "path"
+import fs from "fs-extra"
+
+function oldVersionPath() {
+  return join(app.getPath("appData"), "Local", "Brim")
+}
+
+export function windowsPre25Exists() {
+  const dir = oldVersionPath()
+
+  if (fs.existsSync(dir)) {
+    dialog.showErrorBox(
+      "Previous Brim Version Detected",
+      `Please uninstall it before before launching the new version.\n\n${dir}`
+    )
+    return true
+  }
+  return false
+}
--- a/src/js/electron/windows-pre-25.ts
+++ b/src/js/electron/windows-pre-25.ts
@@ -18,6 +18,16 @@ function getExe() {
   return join(getDir(), "Brim.exe")
 }
 
+/**
+ * Is there a duplicate, older version of Brim on the windows system?
+ * In version 25, we switched our installer. Previous versions placed
+ * the Brim.exe file in /%LocalAppData%/Brim/Brim.exe. It now places
+ * it in /%LocalAppData%/Programs/Brim/Brim.exe.
+ *
+ * If the old version exists, we inform the user to uninstall it first,
+ * then launch the new one.
+ * @returns boolean
+ */
 export function windowsPre25Exists() {
   if (os.platform() !== "win32") return false
 
--- a/src/js/electron/windows-pre-25.ts
+++ b/src/js/electron/windows-pre-25.ts
@@ -1,11 +1,11 @@
 import {app, dialog} from "electron"
-import {join} from "path"
 import fs from "fs-extra"
 import os from "os"
+import {join} from "path"
 
 const VAR = "LocalAppData"
 
-function oldVersionPath() {
+function getDir() {
   if (VAR in process.env) {
     const dir = process.env[VAR]
     return join(dir, "Brim")
@@ -14,11 +14,16 @@ function oldVersionPath() {
   }
 }
 
+function getExe() {
+  return join(getDir(), "Brim.exe")
+}
+
 export function windowsPre25Exists() {
   if (os.platform() !== "win32") return false
 
-  const dir = oldVersionPath()
-  if (!fs.existsSync(dir)) return false
+  const dir = getDir()
+  const exe = getExe()
+  if (!fs.existsSync(exe)) return false
 
   dialog.showErrorBox(
     "Previous Brim Version Detected",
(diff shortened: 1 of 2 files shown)


philrz (dismissed):
@jameskerr: Something didn't quite look right here, so I made a test build based on this branch (https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-0beafb57.0/windows/Brim-Setup-0.25.0-prerelease-0beafb57.0.exe) and confirmed that something is indeed wrong.

The problem I see is that what they're calling "appData" isn't resolving to where the _app code_ is installed but instead where the app puts the _saved user data_. Based on my observations, _that_ directory is the same across both app versions, which is the problem in a nutshell.

On both my personal Windows 10 laptop and a scratch Windows 2019 VM on Google Cloud, `%APPDATA%` resolves to `C:\Users\%USERNAME%\AppData\Roaming`.

image

As you can see, there's no `Local` directory below that.

image

And that `Brim` directory you see there includes the saved stuff like `appState.json`, the `data/` directory, etc.

image

Your observations about `Local\Brim` for the old app and `Local\Programs\Brim` are only accurate "one level up", i.e. where the `Local` in each of those paths is the one at the tail end of `%LOCALAPPDATA%`. (Notice the absence of `Roaming` in both these paths.)

image

In conclusion, the directory the code in this branch is currently looking for winds up being `C:\Users\%USERNAME%\AppData\Roaming\Local\Brim`, which doesn't exist in a typical `v0.24.0` install, so this code was never going to fire. I did create the path manually and launched the new version of the app and saw the message, so the code is at least doing what we expect, but it's just not looking in the right place.

philrz:
Using the latest from this branch, I created a scratch release artifact [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-106f20ff.0/windows/Brim-Setup-0.25.0-prerelease-106f20ff.0.exe](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-106f20ff.0/windows/Brim-Setup-0.25.0-prerelease-106f20ff.0.exe) and tested this out on Windows. I started from having GA `v0.24.0` installed, then installed that scratch artifact. When the app launched at the tail end of the installation procedure, I was greeted with the expected error window:



At that point, after closing the error window, I could still launch the `v0.24.0` version, import new data, etc. But each time I'd close it and try to run the new version again, the error message would still be shown. So far so good.

I then went to the Control Panel and did an uninstall of the `v0.24.0`. Unfortunately, when I relaunched the new version, the error message persisted. I was then reminded that the uninstall of the old app doesn't _completely_ remove the `Local\Brim` directory. Rather, it leaves these droppings behind:



If I deleted that directory, indeed, the new app now launches without complaint. However, I assume we want users to be able to rely on the proper Windows "uninstall" workflow (since that also cleans up the registry and the installed Programs list) rather than deleting the whole directory by hand.

Should we do something like look for the `.dead` file and treat that as evidence that it's not installed and stop popping up the error at that point as well?

philrz:
With the benefit of the changes in the most recent commit, I made yet another scratch release build based on the branch for this PR (https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-eaeac447.0/windows/Brim-Setup-0.25.0-prerelease-eaeac447.0.exe) and it tested out well: Once I uninstalled `v0.24.0` I was able to successfully launch the new version of the app. I'll hold off for @mason-fish to give final approval based on the code, but functionally I'm :+1: on what I've seen.

src/js/electron/windows-pre-25.ts
```diff
@@ -0,0 +1,33 @@
...
+  } else {
+    return join(app.getPath("home"), "AppData", "Local", "Brim")
+  }
+}
+
+function getExe() {
+  return join(getDir(), "Brim.exe")
+}
+
+export function windowsPre25Exists() {
```
mason-fish, line 21: Would you mind copying a bit of your PR description into the code as a comment for posterity?

mason-fish (approved):

src/js/electron/windows-pre-25.ts
jameskerr: Yes, thank you.

PR #1633, opened 2021-05-11
remove async from onbeforeunload

### LINKED ISSUE #1631: Reload no longer shows new pools (regression at be864df)
Background: Per the first attached video, as of Brim commit 33dc417, when new pools were added from outside of Brim (such as via `zapi post` as shown here), a subsequent selection of **View > Reload** from the pull-down menu would make the new pool visible.

https://user-images.githubusercontent.com/5934157/117874329-aed39c00-b255-11eb-95db-c9f70ab50fd7.mp4

This stopped working recently. Via binary search, I confirmed that the failure started in Brim commit be864df, which is associated with #1614. Repeating the steps in the next video when running at that commit, we see my repeated attempts to Reload appear to be no-ops, and to ultimately see the new pool data I had to do a brute force Reset State.

https://user-images.githubusercontent.com/5934157/117874601-04a84400-b256-11eb-9f92-23daa4668cda.mp4

mason-fish:
`window.onbeforeunload` expects a synchronous return, so we will still fire off these async functions but immediately return instead of awaiting their completion.

Fixes #1631

### DIFF
--- a/src/js/detail.tsx
+++ b/src/js/detail.tsx
@@ -15,8 +15,8 @@ import theme from "./style-theme"
 
 initDetail()
   .then(({store, pluginManager}) => {
-    window.onbeforeunload = async () => {
-      await pluginManager.deactivate()
+    window.onbeforeunload = () => {
+      pluginManager.deactivate()
     }
     ReactDOM.render(
       <Router history={global.windowHistory}>
--- a/src/js/search.tsx
+++ b/src/js/search.tsx
@@ -16,11 +16,11 @@ import TabHistories from "./state/TabHistories"
 
 initialize()
   .then(({store, pluginManager}) => {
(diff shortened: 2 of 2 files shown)


jameskerr (approved):
I did some research about async tasks in the onbeforeunload handler. It turns out that they are not guaranteed to finish.

User's don't normally use the reload functionality, it's more a dev thing, so I think this is fine for now. But going forward we should find a reliable way to handle this. Maybe dispatch something to the main process? Or ensure the backend knows what to do with closed connections?

We've handled this reliably for closing the window and quitting the app. This reload piece is the last to solve.

jameskerr:
We could also take over the "Reload (cmd + r)" action, perform our async cleanup, wait for it to finish, then call the browser reload.

PR #1636, opened 2021-05-13
Migration Bugs

### LINKED ISSUE #1567: Provide clickable migration of file store ZNG to Zed lakes

jameskerr:
Closes #1567 

The path to the brimcap exe is now correct for the migration. Previously it was correct in dev, but not in a release.

I've been excited to introduce "ipc service objects" to our electron app and decided this was a good time. There is a new file `app/ipc/paths.ts` that includes a class with a single function `brimcap()`. It returns the path to the brimcap executable. This class then gets transformed into an ipcClient and can be used in the renderer like so.

```
import {pathsClient} from "app/ipc/paths"

const cmd = await pathsClient.brimcap()
```

And it can be used in the main process like so:
```
import {paths} from "app/ipc/paths"
const cmd = paths.brimcap()
```

The main process just needs to "serve" this object to create ipc handlers for all its methods.

```
// main.ts
import {paths} from "app/ipc/paths"

function main() {
  serve(paths)
}
```

I think this is very nice pattern and would love feedback. Maybe we can tweek the api to make it a one-liner to create the client and server. I can think of lots of ipc calls we make that could be converted to little service objects like this.


app/legacy/space-migration/space-migrator.ts
```diff
@@ -24,19 +29,22 @@ export default class SpaceMigrator {
...
       ])
+
       const linesErr = readline.createInterface({input: this.process.stderr})
       let total = 0
       let count = 0
       let space = ""
       let totalRegexp = /^migrating (\d+) spaces$/
 
       linesErr.on("line", (line) => {
+        console.log(line)
```
mason-fish, line 47: Was this just for debug or did you want to keep this in?

mason-fish (approved):
Overall this lgtm! I'd like to chat more about this new service pattern, which looks v interesting, but I don't want to hold back the migration fix included here. :shipit:

philrz:
@mason-fish: RE: chatting more, I just made a note to tee it up as an "after party" topic at our team sync tomorrow. Particularly with @jameskerr about to be away for a bit, I know other Devs are taking an interest in understanding more of the JavaScript.

app/legacy/space-migration/space-migrator.ts
jameskerr: I want that in, in case the user needs more info about a problem.

Issue #1593, opened 2021-04-29
Replace runner/updater prefs with a Brimcap config YAML pref


philrz:
In discussing how to proceed with #1545, @mattnibs, @mason-fish, @nwt and I agreed on the plan to replace these three preferences:

![](https://user-images.githubusercontent.com/5934157/116603737-1dbe0600-a8e2-11eb-8c36-2d5f2ad2a550.png)

...with a single Preference that the user could point at the pathname of a Brimcap config YAML file.

The Preference will start out blank. As long as it's blank, `brimcap load` will continue to be invoked the way it is today, which has the effect of `brimcap load` invoking the paths to the "stock" embedded Zeek+Suricata Runners and applying the default Suricata shaper. If the preference is populated with a valid path, the `brimcap load` command line instead should reference it via its `-config` option, and therefore whatever Zeek/Suricata/other analyzers and alternate shapers in the config YAML will be leveraged.

At one time we also discussed having an editor in the app so the user could directly edit the YAML config rather than just pointing at a path. However, the wider topic of having an editor in the app (#1521) has been waiting behind some other priorities, so for now the plan is to just point to the path.

philrz:
Verified in Brim commit bd5d993.

In the first video I test on macOS using scratch build https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-bd5d9930.0/macos/Brim-0.25.0-prerelease-bd5d9930.0.dmg. To check whether I'm running the bundled Zeek/Suricata or my custom one, I keep this `while` loop running during pcap import:

```
while true; do ps auxww | egrep -i zeek\|suricata; done
```

During the baseline run at defaults, you can see the Zeek/Suricata paths start with `/Applications/Brim.app/Contents/Resources/app.asar.unpacked/zdeps`, indicating the bundled one is being run. Then I switch the setting in **Preferences** to point at this custom Brimcap YAML instead:

```
analyzers:
  - cmd: /Users/phil/bin/myzeek
  - cmd: /Users/phil/bin/mysuricata
    globs: ["eve.json"]
    shaper: |
      type alert = {
      	timestamp: time,
      	event_type: bstring,
      	src_ip: ip,
      	src_port: port=(uint16),
      	dest_ip: ip,
      	dest_port: port=(uint16),
      	vlan: [uint16],
      	proto: bstring,
      	app_proto: bstring,
      	alert: {
      		severity: uint16,
      		signature: bstring,
      		category: bstring,
      		action: bstring,
[...]

PR #1637, opened 2021-05-13
plugin configs + brimcap yaml config/preference

### LINKED ISSUE #1545: Conditional Preferences
Only show Zeek and suricata options in the preferences if the security module is installed (brimcap is present). @mattnibs can show us if/how we can use these options in the brimcap invocation. He can also tell us if some the options are no longer needed in which case remove them from the preferences.

mason-fish:
fixes #1545, #1593 

This PR extends the brimApi to expose a generic persisting storage for plugins, and adds a new UI api which corresponds to what we currently see as preference options.

Following a bit of the naming choices made in VSCode, I have named this new api and its new backing storage/reducer `configs`, and have kept the remaining `prefs` that are not related to brimcap in their current place. We should eventually move those into a core plugin and store them under the new `configs` as well, so that we may fully transition away from the `prefs` state.

As it stands, the new redux slices we've added to facilitate plugins contributing ui elements (`toolbars` first, and now `configs`, and soon we will want something for `contextMenus` too) are NOT persisted. Every time the app starts up, plugins will populate and maintain these slices via the brimApi, the app will "react" to those state changes appropriately, and then when the app quits it will not save that ui state. What I like about this pattern is that by not persisting that plugin-generated state, and by regulating it's manipulation through the brimApi, we have more flexibility to make changes to the underlying app state shape without breaking things or requiring migrations.

This does create a need for plugins to persist their own data though, hence including this new generic `storage` object on the brimApi. For now it functions as a simple key value store, where the key is a plugin namespace and the value is arbitrary. I'd like to eventually protect this better so that plugins cannot access each other's storage which is currently possible, and when the need arises we could also enhance the storage to optionally use the keychain for a more secure persisting storage. This would then align the api pretty well for transitioning the auth related work into it's own plugin too.


src/js/components/Preferences/Preferences.tsx
```diff
@@ -7,24 +7,68 @@ import ModalBox from "../ModalBox/ModalBox"
...
+              type="text"
+              placeholder="default"
+              defaultValue={defaultValue}
+            />
+          </div>
+        )
+    }
+  })
+
+  return <>{formInputs}</>
```
jameskerr, line 54:
```suggestion
  return formInputs
```

src/js/initializers/initialize.ts
```diff
@@ -10,7 +10,7 @@ import BrimApi from "../api"
 export default async function initialize() {
   const api = new BrimApi()
   const store = await initStore(api)
-  api.setStoreArgs(store.dispatch, store.getState)
+  api.init(store.dispatch, store.getState)
```
jameskerr, line 13: 👍

src/js/api/ui-apis.ts
```diff
@@ -27,10 +14,20 @@ export class ToolbarApi implements BrimUIContainerApi<ToolbarItem> {
...
-//   update() {}
-// }
+export class ConfigsApi {
+  constructor(private dispatch: AppDispatch, private getState: () => State) {}
+
+  public add(config: Config) {
+    this.dispatch(Configs.create(config))
+  }
+
+  public updatePropertyDefault(
```
jameskerr, line 24:
```suggestion
 updatePropertyDefault(
```

src/js/api/ui-apis.ts
```diff
@@ -27,10 +14,20 @@ export class ToolbarApi implements BrimUIContainerApi<ToolbarItem> {
...
-//   constructor(private store: Store) {}
-//
-//   add() {}
-//
-//   update() {}
-// }
+export class ConfigsApi {
+  constructor(private dispatch: AppDispatch, private getState: () => State) {}
+
+  public add(config: Config) {
```
jameskerr, line 20:
```suggestion
 add(config: Config) {
```

jameskerr (changes requested):
Looking good!

It looks like the "value" of a config is stored in the plugin storage slice and in the config slice (as the default value). I also see a new command created for each config property which is like an "onChange" listener.

I propose we store the config settings in a state slice (like you have), but save the config values in another separate slice of state. (config values). Config values will be persisted. The form could handle updating the config values automatically whenever it gets submitted. Then we wouldn't need to manually save the values in storage or update the default value. We also wouldn't need the custom command.

If we did that, the brimcap plugin would become...
```ts
// brimcap-plugin.ts

setupConfig() {
  this.api.configs.add({
      name: this.pluginNamespace,
      title: "Brimcap Settings",
      properties: {
        [configPropertyName]: {
          name: configPropertyName,
          type: "file",
          label: "Brimcap YAML Config File",
          defaultValue: "" // only used in the form if the config value is unset
        }
      }
  })
}
```

The configs api could become...
```ts
class ConfigsApi {
  add(config: Config) {
    this.dispatch(Configs.create(config))
  }

  update(key, value) {
    this.dispatch(ConfigValues.set(key, value))
  }

  get(key) {
     ConfigValues.get(key)(this.getState())
  }
}
```

And the useConfigsForm hook would change the submit function to...
```ts
// usePreferencesForm.ts line 28

const submit = (value) => dispatch(ConfigValues.set(key, value))
```

I like the idea of having the values all stored together and managed by the preferences form. As a plugin, I just add the configuration object and I know that whenever I ask for the value, it will be up to date.

I like the storage state slice and api, so let's leave that. It can be used for any data that the plugin wants to keep around and it will be responsible for updates.

What do you think?

src/js/state/ConfigPropValues/test.ts
```diff
@@ -0,0 +1,110 @@
...
+})
+
+test("Create or Update (Upsert)", () => {
+  // add one prop to one config
+  dispatch(
+    ConfigPropValues.set({
+      configName: configName1,
+      propName: configProperty1,
+      value: data1
+    })
```
jameskerr, line 29: What do you think about making the config values a flat object, and we trust that the keys will be prefixed properly? I think it would make for a more clear, less verbose api if it looked like: `ConfigValues.set(key, value)`

src/js/state/Configs/index.ts
```diff
@@ -0,0 +1,42 @@
+import {createEntityAdapter, createSlice} from "@reduxjs/toolkit"
+import {State} from "../types"
+
+export type ConfigItemType = "file" | "string" // | "number" | "boolean"
+
+export type ConfigItem = {
+  name: string
```
jameskerr, line 7: Maybe this name doesn't need to be stored? Since the key and the name are the same and should be in sync?

jameskerr (commented):
This is coming together! I left a few suggestions for ya. The most important to me is saving the config values as a flat object, instead of nested by name. What do you think of that?

jameskerr (approved):

PR #1640, opened 2021-05-14
Pool name bug

### LINKED ISSUE #1638: Pools not shown post-migration on macOS
Repro was with Brim commit 89a053b on macOS using scratch artifact [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-89a053b9.0/macos/Brim-0.25.0-prerelease-89a053b9.0.dmg.](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-89a053b9.0/macos/Brim-0.25.0-prerelease-89a053b9.0.dmg.)

As shown in the attached video, I start out running GA Brim tagged `v0.24.0` and have an existing Space based on attached pcap [ifconfig.pcapng.gz](https://github.com/brimdata/brim/files/6474580/ifconfig.pcapng.gz) (after uncompressing). I then close the app and update to the newer Brim and accept the prompt to migrate Spaces. The migration appears to have completed successfully, but at the end of the migration the Pools list is still not visible. I try selecting **View > Reload** and they're still not visible. As you can see, the DevTools doesn't seem to show a glaring error.

https://user-images.githubusercontent.com/5934157/118175176-9cd43380-b3e4-11eb-98b5-f6c47fd1789a.mp4

Interestingly, I'd tried testing migration on Windows first (admittedly using different Spaces) and _that_ seemed to work ok with the pools being visible at the end. I'm not sure why they'd be different.

https://user-images.githubusercontent.com/5934157/118175242-b4abb780-b3e4-11eb-8626-c938e362ba7b.mp4

jameskerr:
Fixes #1638

The id of the sidebar section changed from "spaces" to "pools". But the persisted data from version 24 still has the id set to "spaces". This caused the sidebar to not render that section in version 25.

I've added to the migration that Matt wrote to include changing the section ids.

I wonder if there is a better way to save layout data like this. Seems odd to write migrations each time something small changes. I don't have an answer yet, but I will be thinking about it.

The screenshot below shows that the pools section is now visible.

Screen Shot 2021-05-14 at 8 26 51 AM

### Side fix

I fixed a bug in the RightPaneExpander component that would always display an error in the console. Those who look in the devtools will be familiar with it. It was the result of not clearing a timeout when the component unmounted.

Screen Shot 2021-05-13 at 3 29 15 PM

### DIFF
--- a/src/js/state/migrations/202104291255_dropSpaces.ts
+++ b/src/js/state/migrations/202104291255_dropSpaces.ts
@@ -1,9 +1,14 @@
-import {getAllStates} from "../../test/helpers/getTestState"
+import {getAllStates, getAllTabs} from "../../test/helpers/getTestState"
 
 export default function dropSpaces(state: any) {
   // Default search records to "events"
   for (const s of getAllStates(state)) {
     delete s.spaces
+  }
+  for (const tab of getAllTabs(state)) {
+    tab.layout.sidebarSections.forEach((section) => {
+      if (section.id === "spaces") section.id = "pools"
+    })
   }
   return state
 }
--- a/src/js/lib/MouseoverWatch.ts
+++ b/src/js/lib/MouseoverWatch.ts
@@ -28,6 +28,7 @@ export default class MouseoverWatch {
     document.removeEventListener("mousemove", this.listener)
     const bod = document.body
     bod && bod.removeEventListener("mouseleave", this.startExit)
+    clearTimeout(this.tid)
     return this
   }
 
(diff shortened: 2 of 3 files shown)


src/js/state/migrations/202104291255_dropSpaces.ts
```diff
@@ -1,9 +1,14 @@
...
+import {getAllStates, getAllTabs} from "../../test/helpers/getTestState"
 
 export default function dropSpaces(state: any) {
   // Default search records to "events"
   for (const s of getAllStates(state)) {
     delete s.spaces
   }
+  for (const tab of getAllTabs(state)) {
+    tab.layout.sidebarSections.forEach((section) => {
+      if (section.id === "spaces") section.id = "pools"
```
mason-fish, line 10: ah, sneaky

mason-fish (approved):

PR #1643, opened 2021-05-14
Detail formats

### LINKED ISSUE #1642: Timestamp in Log Detail lacks precision & doesn't reflect Preferences
Repro is with Brim commit a52d915.

Importing the attached [sample.pcap.gz](https://github.com/brimdata/brim/files/6480393/sample.pcap.gz) (after uncompressing), we see that the timestamp in the Log Detail view differs significantly from the one in the main events panel. Specifically:

1. The precision (`.328`) has been dropped.
2. It shows the time relative to UTC (`-07:00`). The default **Preferences** setting for **Timezone** is **UTC**, so based on that I'd expect the format ending in `Z` like we see in the main events panel.

image

The context in which this became significant is that I'm updating some Support articles that advise users on how to debug pcap problems in the Brimcap era. To that end, I was trying to manually reconstruct a `brimcap search` command line equivalent to the one the app would launch by cut & pasting the values from the Log Detail panel. I see now that if I'd done a right-click **Copy** I actually would have gotten the format I was expecting, and in a pinch I could stress in the article that the user take this approach and advise against using keyboard shortcuts like Cmd-C, which is what I was doing when I bumped into the problem. But even outside the cut & paste context, having the presentation in both spots line up seems like a good idea.

Incidentally, the difference in behavior between Cmd-C and right-click **Copy** reminded me of https://github.com/brimdata/brim/pull/1535 where it was noted that there was similar unfinished business regarding the stripping of commas when data was headed to the paste buffer. I bumped into this again when copying the port values like `id.orig_p`, though it was pretty easy to spot and remove the comma in that case. A comment was made in that PR:

> Since the data viewer is due for an upgrade, we will consider filling in the holes when that upgrade takes place.

Maybe this implies that now is still not the time to make fixes to Log Detail?

jameskerr:
Fixes #1642 

* Removes commas from ports.
* Formats time the same way in the detail window and the search window.
* Adds type colors to the detail pane in both windows.

Screen Shot 2021-05-14 at 12 10 22 PM

image


jameskerr:
Side fix: I moved the notifications (toasts) to be anchored on the bottom right of the page. I think this is a better place because the "menu" buttons and the "close buttons" are positioned on the right and will not be moving around if the width of the notification needs to change as it updates.

mason-fish (approved):

src/js/components/BrimToaster.tsx
```diff
@@ -28,7 +28,7 @@ const BrimToaster = () => {
     <>
       <HiddenToast {...toastLocator.props}>{lastToast?.message}</HiddenToast>
       <Toaster
-        position="bottom-center"
+        position="bottom-right"
```
mason-fish, line 31: 👍

PR #1645, opened 2021-05-14
Zed Lang Change

### LINKED ISSUE #1644: Zed Lang Update
Update the app to use the updated language described in the zed pr: https://github.com/brimdata/zed/pull/2692

jameskerr:
Fixes #1644 

Update the app with the new equality, assignment, and matches syntax. Based off the zed branch : 
https://github.com/brimdata/zed/pull/2692

|before | after|
|---| ---|
|`_path=conn` | `_path=="conn"`|
|`query=google*` | `query matches google*`|
|`put sum = a + b` | `put sum := a + b`|

* Updates all app queries
* Updates queries in the itest suite
* Migrates the query library
* Adds parse tests for the default queries in the query library

**Note:** It is likely that user's history entries will contain queries with wrong syntax now. This will be addressed prominently in the release notes, but that's it. No attempt to migrate old history queries will be made. The tooling to do so is not present and not easily built.

### DIFF
--- a/zealot/zealot.ts
+++ b/zealot/zealot.ts
@@ -37,6 +37,7 @@ export function createZealot(
       return promise({method: "GET", path: "/auth/identity"})
     },
     search: (zql: string, args?: Partial<SearchArgs>) => {
+      console.log("zql", zql)
       return stream(search(zql, {...searchArgs, ...args}))
     },
     archive: {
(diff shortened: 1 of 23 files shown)


zealot/zealot.ts
```diff
@@ -37,6 +37,7 @@ export function createZealot(
       return promise({method: "GET", path: "/auth/identity"})
     },
     search: (zql: string, args?: Partial<SearchArgs>) => {
+      console.log("zql", zql)
```
nwt, line 40: Do you want this?

nwt (approved):

zealot/zealot.ts
jameskerr: Nope!

jameskerr, line 40:
```suggestion
```

PR #1651, opened 2021-05-17
invoke suricata-updater on app load

### LINKED ISSUE #1595: Invoke the embedded Suricata updater when Brim launches
brimdata/brimcap#36 describes this topic from the Brimcap perspective in some detail.

In a more recent discussion, @mattnibs, @nwt, @mason-fish and I agreed that for now we should just have Brim execute the embedded `zdeps/suricata/suricataupdater` once whenever the app is launched. This takes the place of the previous functionality when `zqd` used to do the equivalent.

While users might end up pointing the Brimcap config YAML at some custom Suricata that has its own updater (#1593) and hence not depend on the rules being updated for the embedded artifact, we agreed that for now this can live as a harmless no-op for that minority of users. If anyone ends up having a problem with it or complaining we could add a Preference later to turn off the update-at-launch.

mason-fish:
fixes #1595 

Runs on app load, and also on refresh. If we only wanted this to run on app load then I could add some api plumbing so that plugins can listen for a brim load event (i.e. `this.api.onAppLoaded()`) but this seemed simpler and better.

### DIFF
--- a/plugins/brimcap/brimcap-plugin.ts
+++ b/plugins/brimcap/brimcap-plugin.ts
@@ -10,7 +10,7 @@ import {AppDispatch} from "../../src/js/state/types"
 import {Config} from "../../src/js/state/Configs"
 import {reactElementProps} from "../../src/js/test/integration"
 import BrimcapCLI, {loadOptions, searchOptions} from "./brimcap-cli"
-import {ChildProcess} from "child_process"
+import {ChildProcess, spawn} from "child_process"
 
 export default class BrimcapPlugin {
   private pluginNamespace = "brimcap"
@@ -24,7 +24,7 @@ export default class BrimcapPlugin {
   private selectedConn = null
   private yamlConfigPath = ""
   private cleanupFns: Function[] = []
-  private loadingProcesses: {
+  private processes: {
     [pid: number]: ChildProcess
   } = {}
   private brimcapDataRoot = ""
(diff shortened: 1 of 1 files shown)


plugins/brimcap/brimcap-plugin.ts
```diff
@@ -24,7 +24,7 @@ export default class BrimcapPlugin {
   private selectedConn = null
   private yamlConfigPath = ""
   private cleanupFns: Function[] = []
-  private loadingProcesses: {
+  private processes: {
```
mason-fish, line 27: reusing this as just a dump of any async processes, so that I can also reuse the cleanup logic invoked when users quit or refresh the app

plugins/brimcap/brimcap-plugin.ts
```diff
@@ -329,21 +331,43 @@ export default class BrimcapPlugin {
     this.api.configs.add(brimcapConfig)
   }
 
+  private async updateSuricata() {
+    const {zdepsDirectory} = this.api.getAppConfig()
+    const cmdName =
+      process.platform === "win32" ? "suricataupdater.exe" : "suricataupdater"
+    const cmdPath = path.join(zdepsDirectory, "suricata", cmdName)
```
jameskerr, line 338: I hope to consolidate code like this in `app/ipc/paths`, and expose it through the plugin api.

Maybe like `this.api.getZdepPath("suricata/suricataupdater")`. It could be responsible for adding the exe to the end of commands.

jameskerr (approved):
Cool!

jameskerr:
We can merge this now, but since we as devs rely on reload so much, I wonder if it would make things slower to always be updating suricata. And if user's open a new browser window, this will run. Maybe in the future, instead of on load, we run it once every 24 hours. And keep a timestamp of the last time it was run.

plugins/brimcap/brimcap-plugin.ts
mason-fish: Ah yeah that makes sense

PR #1654, opened 2021-05-18
handle adapter bug for global dispatch

### LINKED ISSUE #1653: "TypeError: Cannot convert undefined or null to object" when opening new window
Repro is on Brim commit 207cb69 via scratch build [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-207cb691.0/macos/Brim-0.25.0-prerelease-207cb691.0.dmg](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-207cb691.0/macos/Brim-0.25.0-prerelease-207cb691.0.dmg).

If I bring up the app "fresh" and then immediately select **File > New Window** from the pull-down, I get a crash dump:

```
TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at /Applications/Brim.app/Contents/Resources/app.asar/dist/src/js/components/Preferences/usePreferencesForm.js:39:12
    at Array.forEach (<anonymous>)
    at useConfigsForm (/Applications/Brim.app/Contents/Resources/app.asar/dist/src/js/components/Preferences/usePreferencesForm.js:38:11)
    at Preferences (/Applications/Brim.app/Contents/Resources/app.asar/dist/src/js/components/Preferences/Preferences.js:117:60)
    at renderWithHooks (/Applications/Brim.app/Contents/Resources/app.asar/node_modules/react-dom/cjs/react-dom.development.js:14803:18)
    at updateFunctionComponent (/Applications/Brim.app/Contents/Resources/app.asar/node_modules/react-dom/cjs/react-dom.development.js:17034:20)
    at beginWork (/Applications/Brim.app/Contents/Resources/app.asar/node_modules/react-dom/cjs/react-dom.development.js:18610:16)
    at HTMLUnknownElement.callCallback (/Applications/Brim.app/Contents/Resources/app.asar/node_modules/react-dom/cjs/react-dom.development.js:188:14)
    at Object.invokeGuardedCallbackDev (/Applications/Brim.app/Contents/Resources/app.asar/node_modules/react-dom/cjs/react-dom.development.js:237:16)
```

https://user-images.githubusercontent.com/5934157/118677003-bd1f3c00-b7b0-11eb-9e33-119a722efa63.mp4

mason-fish:
fixes #1653 

This is an issue we hadn't seen before because until recently we hadn't combined the redux toolkit entity adapters we've begun to use with our existing globally dispatched actions. It has a simple fix though.

Actions may have a `remote` key set if dispatched from another window. In such a case the `adapter.addOne()` and similar CRUD methods will not recognize the passed in action param to be of type `PayloadAction` and will instead interpret the value as an Entity (i.e. a Config type in this case). So instead, we will always pass in the entity/action.payload when using adapters for global reducers.

I also changed the method from `create` to `set` and used upsertOne, since it was already overwriting the existing value (semantically speaking, "set" indicates that overwriting behavior to me more than "create" does) and now matches the same `set`/`delete` api as the `ConfigPropValues`, and `PluginStorage` reducers so that we are more consistent

### DIFF
--- a/src/js/api/ui-apis.ts
+++ b/src/js/api/ui-apis.ts
@@ -23,6 +23,6 @@ export class ConfigsApi {
   }
 
   add(config: Config) {
-    this.dispatch(Configs.create(config))
+    this.dispatch(Configs.set(config))
   }
 }
--- a/src/js/state/Configs/index.ts
+++ b/src/js/state/Configs/index.ts
@@ -27,8 +27,12 @@ const slice = createSlice({
   name: "$configs",
   initialState: adapter.getInitialState(),
   reducers: {
-    create: adapter.addOne,
-    delete: adapter.removeOne
+    set(state, action) {
+      adapter.upsertOne(state, action.payload)
+    },
+    delete(state, action) {
+      adapter.removeOne(state, action.payload)
+    }
   }
 })
 
--- a/src/js/state/Configs/test.ts
+++ b/src/js/state/Configs/test.ts
@@ -41,18 +41,18 @@ beforeEach(() => {
 
 test("Create", () => {
   // add one
-  dispatch(Configs.create(testConfig1))
+  dispatch(Configs.set(testConfig1))
   expect(select(Configs.all)).toHaveLength(1)
   expect(select(Configs.get(testConfig1.name))).toEqual(testConfig1)
 
   // add second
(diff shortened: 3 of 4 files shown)


jameskerr (approved):
Tricky!

PR #1655, opened 2021-05-18
Release notes

### LINKED ISSUE #1594: Release Notes Tab Upon Upgrade
When discussing the plan for #1593, we recognized it would be a helpful courtesy to the user if we pop up a notification upon upgrade that describes in brief what has changed and offers a link to the article described in brimdata/brimcap#64 so they know they need to recreate a new config if they want to continue using the customizations they had before.

While this should help us with the tactical problem of guiding users through this specific Preferences transition, during the discussion we recognized that this is helpful functionality we might want to use in future upgrades generally, i.e. (similar to other apps) a "What's New" pop-up that displays or links to the Release Notes or other content relevant to the upgrade. This seems like good practice since, especially given the app's tendency to auto-update, we can't always assume users are going to read the Release Notes and pay attention to important changes. Anything we can do to put this content front & center should help avoid user frustration or excess support load.

When first proposing this I imagined this initial notification only being specific to if the user had populated any of those three Preferences settings. However, on further reflection I realize it probably would not be a bad thing to notify _all_ users just in case, e.g. maybe a user _had_ changed those settings at one time, then reverted them to their default blank state, but would still benefit from knowing about the change for the next time they need it. Considering all the other new stuff we're planning to introduce in the next GA release, I'm guessing we'll think of other content/links to add to a general "everybody sees it" notification, so I hereby walk back that initial part of my proposal, and hopefully that simplifies the implementation somewhat. :smile:

jameskerr:
Closes #1594 

Follows the spec described in the issue. Here are some screenshots.

Screen Shot 2021-05-18 at 2 18 23 PM

Screen Shot 2021-05-18 at 3 32 27 PM

Screen Shot 2021-05-18 at 2 19 04 PM

Screen Shot 2021-05-18 at 2 18 53 PM


app/release-notes/release-notes.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const version = getVersion()
+
+export default function ReleaseNotes() {
+  const notes = useReleaseNotes(version)
+
+  return (
+    <Scrollable>
+      <BG className="release-notes">
```
mason-fish, line 39: I can't find where you're using this class (and am hoping that it is simply unused so that all of this component's styling is in one place 😬 ). Do we need it?

src/js/state/rootReducer.ts
```diff
@@ -20,8 +20,10 @@ import Toolbars from "./Toolbars"
 import PluginStorage from "./PluginStorage"
 import Configs from "./Configs"
 import ConfigPropValues from "./ConfigPropValues"
+import Launches from "./Launches"
 
 export default combineReducers<any, any>({
+  launches: Launches.reducer,
```
mason-fish, line 26: I think you'll also want to add this to the global reducer, no?

package.json
```diff
@@ -125,7 +124,6 @@
     "jwt-decode": "^3.1.2",
     "keytar": "^7.3.0",
     "lodash": "^4.17.19",
-    "markdown-it": "^10.0.0",
```
mason-fish, line 128: Oh nice, I forgot we even had this

mason-fish (approved):
lgtm! I pulled it down to click around and it seems very solid 👍 Just left a couple comments (I think my global state comment is the only one of significance)

app/release-notes/release-notes.tsx
jameskerr: Nope, that was used when I was developing, but it's not needed anymore. Thanks!

src/js/state/rootReducer.ts
jameskerr: Yes I do!

PR #1657, opened 2021-05-19
space migrate: refresh pool on fail

mattnibs:
If migrate fails because it was cancelled, the current pool it is
operating on will be deleted. As a result, refresh pools on failure so
things remain up to date.

Also:
- Use tee-1 package to tee migration logs into the brim process stderr.
  Better to have some visibility than none.


jameskerr (approved):
Checks out. Thanks!

PR #1664, opened 2021-05-20
Download pcap context menu

### LINKED ISSUE #1583: Brimcap Plugin: Download PCAPs right-click option
We currently have a right-click option to download pcaps when appropriate. Let's keep that behavior but move it into the brimcap plugin where it will use the `brimcap` cli instead of zqd. The main chunk of work to do here is to integrate our context menus with the new plugin architecture, much like the toolbar buttons in https://github.com/brimdata/brim/pull/1573

mason-fish:
fixes #1583 

returns the context menu item for launching wireshark for "conn" logs, but now through the brim api which has been extended to support this feature.


src/js/api/registries.ts
```diff
@@ -1,6 +1,8 @@
 import {EventEmitter} from "events"
 import {remove} from "lodash"
 import {IngestParams} from "../brim/ingest/getParams"
+import {zed} from "../../../zealot"
```
jameskerr, line 4:
```suggestion
import {zed} from "zealot"
```

jameskerr (approved):
Cool! I like how simple it ended up.

PR #1667, opened 2021-05-20
Increase expected pool size due to Zed dcce5c22 change

philrz:
The `npm run test:api` test in the Brim CI recently started failing because the pool size after importing `itest/testdata/custom-sample.ndjson` had increased from 4546 to 4467 as a result of changes in Zed commit `dcce5c22`  (brimdata/zed#2729). At a team discussion today it was agreed that this change is expected, so I've modified the expected value so the CI will start passing again.


mason-fish (approved):
Seeing as how the tests pass now, this lgtm

nwt (approved):
:shipit:

PR #1671, opened 2021-05-21
legacy space migrate: delete pool

mattnibs:
Due to issues with not being able to send a SIGINT signal in windows, if
a migrate is aborted have the brim app delete the current pool.

### DIFF
--- a/app/legacy/space-migration/space-migrator.ts
+++ b/app/legacy/space-migration/space-migrator.ts
@@ -53,7 +53,7 @@ export default class SpaceMigrator {
         if ("space" in status && status.space !== space) {
           space = status.space
           count++
-          onUpdate({total, space: space, count})
+          onUpdate({total, space, count})
         }
         if ("pool_id" in status) {
           this.currentPoolID = status.pool_id
--- a/app/legacy/space-migration/space-migrator.ts
+++ b/app/legacy/space-migration/space-migrator.ts
@@ -11,6 +11,7 @@ import tee from "tee-1"
  */
 export default class SpaceMigrator {
   process: ChildProcess | null
+  currentPoolID: String | null
 
   constructor(readonly srcDir: string, readonly destDir: string) {}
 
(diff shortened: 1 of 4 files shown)


mattnibs:
Depends on https://github.com/brimdata/brimcap/pull/89

philrz:
I tested this out on all platforms using scratch artifacts based on this branch:

*   [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/macos/Brim-0.25.0-prerelease-a1201962.0.dmg](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/macos/Brim-0.25.0-prerelease-a1201962.0.dmg)
*   [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/windows/Brim-Setup-0.25.0-prerelease-a1201962.0.exe](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/windows/Brim-Setup-0.25.0-prerelease-a1201962.0.exe)
*   [https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/linux/Brim-0.25.0-prerelease-a1201962.0.deb](https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-a1201962.0/linux/Brim-0.25.0-prerelease-a1201962.0.deb)

In all cases I observed the expected improvement that after clicking **Cancel**, the entry for the Pool-currently-being-migrated ultimately disappeared from the list. At that point I could reload/restart the app, the migration pop-up would appear again, and it would restart and complete the remaining migrations, including the one that had been canceled.

FWIW, I did find one more way to break it, but I'm not advocating we sink more time into it. During one test where the migration-in-progress was fairly large, I noticed there was a delay of several seconds between when I hit **Cancel** and when the "Migration was canceled" notification was shown and the Pool-currently-being-migrated disappeared from the list. I then confirmed that if I happened to quit the app during those few seconds of exposure, the empty Pool would be left behind. This meant a re-tried migrations pass would continue failing on that one Pool, such that I could only shake it loose by deleting the Pool manually, much as before. However, I see this as a corner case related to code that each user should only have to run once, so I'm content to just mention this in updates I'll make to the [wiki article](https://github.com/brimdata/brim/wiki/Migration-of-Spaces).

app/legacy/space-migration/space-migrator.ts
```diff
@@ -39,24 +40,23 @@ export default class SpaceMigrator {
...
-            total = parseInt(match[1])
-          }
+        if (status.msg === "migrating spaces") {
+          total = status.count
         }
         if ("space" in status && status.space !== space) {
           space = status.space
           count++
-          onUpdate({total, space, count})
+          onUpdate({total, space: space, count})
```
jameskerr, line 56:
```suggestion
          onUpdate({total, space, count})
```

app/legacy/space-migration/space-migrator.ts
```diff
@@ -39,24 +40,23 @@ export default class SpaceMigrator {
...
 
       linesErr.on("line", (line) => {
         console.log(line)
         const status = tryJson(line.toString())
-        if ("msg" in status) {
-          const match = status.msg.match(totalRegexp)
-          if (match) {
-            total = parseInt(match[1])
-          }
+        if (status.msg === "migrating spaces") {
```
jameskerr, line 50: Looks like the status will always be json. Cool.

jameskerr (approved):
One little nit. 👍

PR #1674, opened 2021-05-21
Detail Pane Integration Tests

jameskerr:
These are new integration tests for the log detail pane, along with a new helper method called `itest` to make integration tests easy to write. I've left the old ways of writing integration tests in place so that I can get feedback on this design. If it looks good, then I'll follow up with a PR that converts our existing tests to this pattern. There are not that many so it wouldn't take too long. Maybe a day.

I've updated the readme in the itest directory with this:

 ---

## Adding a New Integration Test

Writing an integration test usually involves performing an action in the UI,  finding an element in the DOM, and asserting it contains the expected value.

For example:

```js
import itest from "../lib/itest"

describe("ingest and check viewer count", () => {
   const [brim, $] = itest("ingest-viewer-count")

   beforeAll(() => brim.ingest("sample.tsv"))
   
   test("row count", () => {
      const rows = brim.findAll($.viewerRows)
      expect(rows.length).toBe(30)
   })
})
```

The helper function `itest()` must be called within a describe block. It takes a string name argument that is used for the log file of this test run. It returns an instance of  `BrimDriver` (`brim`) and `BrimSelectors` (`$`).

You can think of these classes like this:

1. **BrimDriver**: Performs actions (click, input, ingest...)
2. **BrimSelectors**: Creates string selectors for known elements (viewer headers, detail pane...)

## Element Locators

When you need to find a known element, tag it in the code with a locator. A locator provides
a maintainable way to keep track of which elements are needed for integration tests.

```js
type Locator = {
   css: string // a css selector
   xpath: string // an xpath selector
   props: {data-test-locator: string} // props to pass to the react component
}
```

First create one in `src/js/test/locators.ts`.

```js
// The argument will be the value of the data attribute.
export const detailPane = createLocator("detail-pane")
```

Add it to your component.

```js
import {detailPane} from "../test/locators"

<DetailPane {...detailPane.props}>
```

Add it to the Brim Selectors class.

```js
class BrimSelectors {
   detailPane = detailPane.css
}
```

Now you can find the detail pane in the test cases.

```js
   const [brim, $] = itest()

   test("find the detail pane", () => {
      brim.find($.detailPane)
   })
```

> Selectors in the BrimSelectors class don't need to be locator instances, but having a locator will signal to developers that an integration test needs that component.


itest/lib/brim-driver.ts
```diff
@@ -0,0 +1,151 @@
+import {Application} from "spectron"
```
jameskerr, line 1: This file is almost a copy of "createTestBrim". I'll remove the old version if this design gets approval.

itest/lib/brim-selectors.ts
```diff
@@ -0,0 +1,14 @@
...
+  toolbarView = `button[aria-label="View"]`
+  detailPane = locators.detailPane.css
+  detailPaneSections = this.detailPane + " h4"
+
+  viewerCellContaining(text) {
+    return (
+      locators.viewerResults.xpath + `//span[contains(text(), ${quote(text)})]`
+    )
+  }
+}
```
jameskerr, line 14: I will add all of the selectors in `src/js/test/integration.ts` and probably all the locators in `src/js/test/locators` into this file if this design gets approval.

mason-fish (approved):
This is great stuff! So nice to finally get some testing around the detail pane, and simplify the process of writing itests too 🔥

PR #1676, opened 2021-05-24
Update zed lake endpoints

mattnibs:
Update client to reflect changes to the zed lake api reflected in
brimdata/zed#2741

### DIFF
--- a/src/js/flows/refreshPoolNames.ts
+++ b/src/js/flows/refreshPoolNames.ts
@@ -6,10 +6,21 @@ import {getZealot} from "./getZealot"
 export default function refreshPoolNames(): Thunk<Promise<void>> {
   return (dispatch, getState) => {
     const zealot = dispatch(getZealot())
-    return zealot.pools.list().then((pools) => {
-      pools = pools || []
-      const id = Current.getWorkspaceId(getState())
-      if (id) dispatch(Pools.setPools(id, pools))
-    })
+    let pools = []
+    return zealot.pools
+      .list()
+      .then((data) => {
+        pools = data || []
+        return Promise.all(
+          pools.map(async (pool, i) => {
+            let stats = await zealot.pools.stats(pool.id)
+            pools[i] = {...pool, ...stats}
+          })
+        )
+      })
+      .then(() => {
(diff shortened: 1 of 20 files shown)


zealot/fetcher/fetcher.ts
```diff
@@ -11,7 +11,7 @@ export type FetchArgs = {
   path: string
   method: string
   body?: string | FormData
-  headers?: Headers
+  headers?: any
```
jameskerr, line 14: Can this be left as `Headers`? Or `object`

jameskerr (approved):

zealot/fetcher/fetcher.ts
mattnibs: Could be object, headers didn't work.

src/js/flows/refreshPoolNames.ts
```diff
@@ -6,10 +6,21 @@ import {getZealot} from "./getZealot"
...
+      .list()
+      .then((data) => {
+        pools = data || []
+        return Promise.all(
+          pools.map(async (pool, i) => {
+            let stats = await zealot.pools.stats(pool.id)
+            pools[i] = {...pool, ...stats}
+          })
+        )
+      })
```
jameskerr, line 20: You don't need to make the calls for all the stats here. The pool config object array will work fine when being passed into `Pools.setPools()`. It's a bad name, but it will merge the data with whats already there. And the app knows that calling `refreshPoolNames()` doesn't update all the pools stats.

jameskerr (commented):
I like that the bigint dep is gone and it looks like we've got zjson everywhere! Very cool.

src/js/flows/refreshPoolNames.ts
mattnibs: You do actually James. If this isn't called the pools in state don't have span info and when you click on the pool it shows the pool doesn't have any data. Probably something to revisit in another pr.

zealot/fetcher/fetcher.ts
mattnibs: Actually ended up needing to use Header, will change this back...

PR #1679, opened 2021-05-24
Release notes bug

### LINKED ISSUE #1677: Redundant Release Notes tabs keep opening on Brim relaunch
To attempt to verify #1594 I've taken the approach of creating a personal fork of the Brim repo at https://github.com/philrz/brim, updated the `repository` key in `package.json` to point to [https://github.com/philrz/brim,](https://github.com/philrz/brim,) then recreated all the GitHub Secrets necessary for the build process to work. In the personal repo I then made a release `v0.25.0` and populated its Release Notes on the GitHub side with a couple pages of text, most of which happened to be the Markdown in the `LICENSE.txt` file at the top level of the repo. A scratch build of this release for macOS was used to repro this issue and can be downloaded from https://storage.googleapis.com/phil-tmp/brim/rc-v0.25.0/macos/Brim-0.25.0.dmg .

Upon first launch, the tab with the rendered Release Notes is open front & center, seems like expected behavior. However, if I click to close the tab, quit Brim, then re-launch, the tab is opened yet again. If I close and relaunch the app repeatedly, it also keeps opening redundant Release Notes tabs. This seems out of sync with the design goal described in #1594:

> On startup, check to see if the release notes have already been presented to the user. If so, do nothing...

https://user-images.githubusercontent.com/5934157/119369731-59da5180-bc69-11eb-8729-d9588a4638e8.mp4

jameskerr:
Fixes #1677

An ipc call mistakenly made its way into the Launches reducer. The
reducer is run in both the main and the renderer process, because
it is global. When the ipc was called on the main process, there
was an error.

The error was being thrown in the console:

```
Error occurred in handler for 'globalStore:dispatch': TypeError: Cannot read property 'app' of undefined
```

This caused the "last-launched" state to never be saved.


src/js/state/Launches/index.ts
```diff
@@ -10,9 +10,8 @@ const slice = createSlice({
   name: "$Launches",
   initialState: {},
   reducers: {
-    touchVersion: (state, action: PayloadAction<string | undefined>) => {
-      let key = action.payload || getVersion()
```
jameskerr, line 14: This getVersion() call was what was causing the error. It makes a sync ipc call to the 'main' process. But this code was being run in the "main" process.

mattnibs (approved):

PR #1683, opened 2021-05-25
Article about RPM upgrades

### LINKED ISSUE #1565: Provide guidance to RPM users to uninstall old Brim before installing new Brim
If a Linux user is running an RPM-installed Brim version based on the old build/packaging such as the last GA release `v0.24.0`, then they install a new release RPM based on electron-builder, they'll see this kind of problem:

image

If they uninstall the old version first, there's no problem.

Right now Linux users are not auto-updated and instead are given a notice about availability of the new version along with a **Get Update** button that links to https://www.brimsecurity.com/download/. We discussed this as a group and figured that for now we can put some prominent guidance (big red letters, etc.) on the Download page that informs the user of the need to uninstall. I'll also plan to write a Support article in the wiki that describes this in more detail, and we can link to that from the page as well. We can put the same guidance in the Release Notes in the GitHub Releases page as well.

philrz:
As described in #1565, testing has revealed a minor incompatibility in the Linux RPM packaging between our old and new installers, such that these Linux users will need to manually uninstall the old release before the new one can install successfully. This is a bummer, but thankfully RPM Linux happens to be the least popular of our OS platform permutations, and it's only a one-time procedure to get over this transition, so it's hopefully not an excessive burden.

In addition to adding the article to the wiki, when we announce the `v0.25.0` release, my intent is to also put a link to the article in a prominent red-lettered notice at [https://www.brimsecurity.com/download/](https://www.brimsecurity.com/download/) near the links to the Linux artifacts in hopes of calling the appropriate users' attention to the article before they download the RPM and see the installation fail. I'll also plan to emphasize it in the Release Notes. I've made notes-to-self regarding both of these details in my checklist for when we're doing final assembly of the next release.

Closes #1565


docs/Linux-RPM-Upgrade-(v0.25.0+).md
```diff
@@ -0,0 +1,33 @@
...
+GA Brim release `v0.25.0` uses a new toolkit for packaging and installing the
+app. The [RPM](https://en.wikipedia.org/wiki/RPM_Package_Manager) packages
+for this newer installer are unfortunately incompatible with the RPM packages
+from the older system, such that an attempt to install RPM packages of Brim
+`v0.25.0` (or newer) while an installation of Brim `v0.24.0` (or older) is
+still present results in a failure similar to the one shown below.
+
+![RPM unable to install](media/RPM-unable-to-install.png)
+
+To resolve the issue, simply uninstall the prior Brim release (e.g. by
```
nwt, line 12:
```suggestion
To resolve the issue, simply uninstall the prior Brim release (e.g., by
```

docs/Linux-RPM-Upgrade-(v0.25.0+).md
```diff
@@ -0,0 +1,33 @@
...
+executing `yum remove brim` or `dnf remove brim` in a Terminal window) before
```
nwt, line 13:
```suggestion
executing `yum remove brim` or `dnf remove brim` in your shell) before
```

docs/Linux-RPM-Upgrade-(v0.25.0+).md
```diff
@@ -0,0 +1,33 @@
...
+app settings (which are stored in `$HOME/.config/Brim`, separate from the
+Brim binaries) will still be intact when you now proceed with the install of
+the newer release.
+
+This is a one-time procedure. Once you successfully install the newer Brim
+release, upgrading via RPM packages to future releases newer than `v0.25.0`
+will not require this manual uninstall step.
+
+> **Note:** The procedure described above is _only_ required for Linux
+> distributions that use the RPM-based package format. On distributions that
```
nwt, line 24:
```suggestion
> distributions that use the RPM package format. On distributions that
```

nwt (approved):

PR #1684, opened 2021-05-26
Use Repo In Autoupdate Url

### LINKED ISSUE #1681: New release notification not shown on Linux build based on forked repo
To thoroughly verify #1594 required something that looked like "real" releases a user could upgrade through. To test, I made a personal fork of the Brim repo at https://github.com/philrz/brim, updated the `repository` key in `package.json` to point to https://github.com/philrz/brim, then recreated all the GitHub Secrets necessary for the build process to work. During the testing I noticed that if I installed the release tagged `v0.25.0` and let it hang out in an open state (I left it overnight!), I never saw the notification pop-up regarding the availability of the newer release `v0.26.0` that was already also published on the GitHub Releases page. My verification notes from when the new build system first went in indicate this used to work, so either I was testing wrong back then or something else broke.

The artifacts are linked from https://github.com/philrz/brim/releases, but here's a direct link to the `.deb` artifacts:

* https://storage.googleapis.com/phil-tmp/brim/rc-v0.25.0/linux/Brim-0.25.0.deb
* https://storage.googleapis.com/phil-tmp/brim/rc-v0.26.0/linux/Brim-0.26.0.deb

jameskerr:
Fixes #1681

The main feature of this PR is that the autoupdate url will use the repo specified in the package.json repository property. 

However, I've also discovered a very promising new way to test our components. It sits in between the spectron tests and the unit tests. Traditionally, testing whole react components has been difficult, and too tied to the implementation details. In order to test them properly with jest, we need to mock the DOM, mock the electron module, mock the network, and mock the ipc calls.

I've discovered a few new libraries that provide excellent mocks for all those areas. `msw` creates a fake server that we mock responses too. `electron-mock-ipc` allows us to use ipc in the tests.

This has lead me to think that it'd be best if we make it so that our zealot client library can run in both node and the browser. Libraries like `cross-fetch` would help make this possible. If we did that, we could remove the `mock_zealot` class, remove  `rollup`, and remove deno as a dependency. We'd only have to rewrite a little code to not rely on some of the new experimental DOM apis that don't have a nodejs equivalent. This seems like such a good trade off. We can be so much more confident that our code works.

https://www.youtube.com/watch?v=eg_TFYF_cKM
https://kentcdodds.com/blog/stop-mocking-fetch
https://mswjs.io/
https://www.npmjs.com/package/electron-mock-ipc


app/ipc/paths.ts
```diff
@@ -16,6 +15,10 @@ class Paths {
   root() {
     return app.getAppPath()
   }
+
```
mattnibs, line 18: nit: consistent new lines in between class members.

mattnibs: I suppose but a new line above root since that seems to be what you're doing in the rest of this pr.

app/release-notes/test.tsx
```diff
@@ -0,0 +1,19 @@
...
+    (req, res, ctx) => {
+      return res(ctx.status(200), ctx.json({body: "Testing Release Notes"}))
+    }
+  )
+)
+
+test("fetches the release notes", async () => {
+  render(<ReleaseNotes />)
+  await screen.findByText("Testing Release Notes")
+})
```
mattnibs, line 19: Cool!

app/release-notes/use-release-notes.ts
```diff
@@ -1,29 +1,46 @@
...
+  const url = `https://api.github.com/repos/${repo}/releases/tags/v${version}`
+
+  try {
+    const resp = await fetch(url)
+    if (resp.ok) {
+      const json = await resp.json()
+      return json.body
+    } else {
+      throw new Error("Not found " + url)
+    }
```
mattnibs, line 16: From MDN's documentation on [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API):

> The Promise returned from fetch() won’t reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing

You should throw a more generic error if the stats code isn't 404.

app/release-notes/use-release-notes.ts
```diff
@@ -1,29 +1,46 @@
-import {getRepo} from "app/core/utils/get-repo"
-import {useEffect, useState} from "react"
+import {metaClient} from "app/ipc/meta"
+import {useEffect, useState, version} from "react"
 
-export function useReleaseNotes(version) {
+async function fetchNotes() {
```
mattnibs, line 4: Shouldn't you still be passing version in here? Or is version imported from react the set Version?

jameskerr: Nope, you're correct. That was a mistake.

jameskerr: I updated the test to catch this and fixed it.

test/unit/__mocks__/electron.ts
```diff
@@ -0,0 +1,87 @@
+import {join} from "path"
+import createIPCMock from "electron-mock-ipc"
+
+const mockIpc = createIPCMock()
```
mason-fish, line 4: 💯

mason-fish (approved):
Lgtm

PR #1691, opened 2021-06-01
Take Screenshot if Detail Pane Fails

jameskerr:
Trying to debug what's going on in the windows env.


jameskerr:
This functionality got merged in: #1745

PR #1692, opened 2021-06-02
Replace enzyme with @testing-library

jameskerr:
Enzyme was only being used in 3 unit tests. It was very easy to replace them with their @testing-library equivalents.

I also found that when I merged master into the node-zealot branch, some duplicate test setup files were added back in. I've removed those here.


.importjs.js
```diff
@@ -1,36 +0,0 @@
-module.exports = {
```
jameskerr, line 1: This was config for a development tool that only I was using, that I now don't use.

mason-fish (approved):

PR #1700, opened 2021-06-07
Restore wrccdc link

philrz:
I've been in contact with the wrccdc people during the long time their packet archive has been offline. It's confirmed to be back, hopefully for good, so I'm restoring this link that I'd taken out in #1586.


mason-fish (approved):
Indeed, the link works!

Issue #1716, opened 2021-06-21
Use brimcap analyze instead of load


mason-fish:
The brimcap plugin currently invokes `brimcap load` when it encounters pcaps. Within the `brimcap` tool, that load operation ultimately uses the deprecated `POST /pool/{id}/logs` endpoint which does not publish its completion of the commit as an event sent back to the app to inform it of potential stale data.

The fastest fix could be to add the publishing logic to the deprecated endpoint's handler. However, since we want to move the app to the commit model anyway, let's instead let the brimcap plugin handle this logic by invoking `brimcap analyze` then make the add/commit api requests itself, piping the data from `brimcap` into the request.

This way, the commit event will be properly published and the app will be notified of the change in data.

philrz:
@mason-fish: Since `brimcap load` also populates the packet index, I assume a separate invocation of `brimcap index` will now also need to be a part of this effort, correct?

mason-fish:
@philrz Yep that's right

philrz:
Verified on Brim commit dab9821 on macOS using nightly build https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-dab9821b.0/macos/Brim-0.25.0-prerelease-dab9821b.0.dmg.

As shown in the attached video, when dragging a pcap into the app, we can see from `ps` output that `brimcap analyze` is now being used instead of `brimcap load`. Once the pcap is loaded, I can hit the **Packets** button and extract a flow based on a log entry as has always been true in the past.

https://user-images.githubusercontent.com/5934157/125180529-e4422a80-e1af-11eb-8662-bd36f0d27299.mp4

Here's the same verification on Windows 10.

https://user-images.githubusercontent.com/5934157/125180605-b8737480-e1b0-11eb-89ea-0612c24fd64a.mp4

And on Linux:

https://user-images.githubusercontent.com/5934157/125202323-f2cc2880-e227-11eb-917e-57d2280e97a3.mp4

Thanks @mason-fish!

PR #1722, opened 2021-06-24
Notify stale data on new commit

### LINKED ISSUE #1708: Automatically show new Pool commits & time span updates
Repro is with Brim commit fd23d53 via nightly build https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-fd23d537.0/macos/Brim-0.25.0-prerelease-fd23d537.0.dmg.

As confirmed in a discussion with @mason-fish, the Pool updates the user now sees as a result of the changes linked to #950 are:

1. Newly-created pools
2. Pool rename
3. Deletion of pools

This still leaves significant changes to pool _contents_ which the user would not see reflected without manual steps. In the attached video, I'm testing with the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default) and start with having already imported the `dns` logs via typical drag into the app. Note the span in the time picker that reflects the range of these records (17:15-17:36). Then via `zapi` outside the app, I create a new Pool and we see it reflected immediately in the Pools list (✅ ), but when I then load the sample `ftp` data into it, it's not visible when I click on the space nor if I hit Enter to refresh the default query. I ultimately still have to do a manual **View > Reload** and re-click the Pool, at which time I do see the data. Note the time picker reflects a narrower time range for these records (17:15-17:20) than we saw for the `dns` records. Now I load the `dns` records incrementally to that same Pool via `zapi`. Hitting Enter to refresh the default query, we _do_ see the `ftp` records filled in, but note that the time picker still shows the narrower 17:15-17:20 range, hence not reflecting the wider 17:15-17:36 that we know covers this data we just added. Even clicking **Whole Pool** does not expand the time picker. Only after doing another **View > Reload** do I now see the full range in the time picker.

https://user-images.githubusercontent.com/5934157/121708098-10bd3680-ca8c-11eb-97ff-600ea0617911.mp4

@mason-fish and I discussed this at some length. At a high-level, the way we'd intended to handle these updates would be by popping up some kind of notification to the user indicating that new data is available in the Pool and giving them the option to "refresh", or dismiss the notification now with the knowledge that they'd need to manually refresh if they want to see the new data (perhaps we'd leave behind some kind of visual cue, like showing a red dot as a reminder near the pool name, until they _do_ refresh). If they chose to click the button to refresh now, any currently-entered query would be re-executed, allowing the newly-added to come into play. The reason for the pop-up notification approach is because the user might have been in the middle of typing out a query or staring at a page of results and would find it jarring/undesirable to have their work refreshed without warning.

However, thinking this through in the light of what's shown in the video, we recognize there's unresolved design problems related to the Pool's time span and current values shown in the time picker. At the moment, we maintain no record as to whether the values in the time span happen to have been set that way by having clicked on a fresh Pool and being shown the "Whole Pool" time span or if the user intentionally narrowed the time span as part of some active investigation. Therefore, it's not clear whether a single-click "refresh data" should

1. Maintain the current time picker settings, or,
2. Expand the picker to include what might be a wider overall time span for the Pool

Prior instincts were to opt for the former, but like was shown with our incremental add of the wider `dns` data atop the narrower `ftp` data, unless the user already had intimate knowledge of the time spans of the data being added, this would have the potential to cause confusion. @mason-fish explained that @jameskerr has some draft designs that would introduce visual cues in the UX that would more clearly show how the time range of the data being shown compares to the overall time range of the Pool. @mason-fish and I agreed that implementing such designs would be the ideal long-term solution, since it would allow us to leave the specific time picker settings undisturbed during such refreshes, but make it obvious to the user that they may want to expand the range to include the new data. In the absence of such visual cues, @mason-fish and I agreed that for now we should probably opt to widen the time picker to reflect any wider time span of the Pool.

@mason-fish noted there's some complexity covering all the bases here, since there's the possibility a user might have the data for a Pool already open in one or more tabs, so it opens up questions about the tab(s) in which such notifications might appear, how they're dismissed, etc.

mason-fish:
fixes #1708 

The app now listens for pool commit events from the server. Whenever a pool gets a new commit, the app will refresh it's state of the pool details/stats. If the current tab is accessing a pool during a commit, we will notify the user that new data is available and give them the option to refresh. This refresh will use the currently set span unless that span matches the pre-commit pool's full span, in which case the refresh will automatically switch to the post-commit's new full span.

There is also a bit of foundational code here relating to #1716 which includes extending the zealot client to use the pool add and pool commit apis in the backend, as well as extending the brimcap cli class to use the `analyze` subcommand. I had originally thought to include the switch from `load` to `analyze` in this work but decided instead to focus on just the app's response to new commits.

NOTE: `brimcap load` currently does not use the add and commit api (it uses the deprecated log post api, which internally calls add/commit) so we do not yet receive commit events from that operation. In testing, I modified the backend to publish commit events in the deprecated endpoint, but a better solution is to modify brimcap to switch to the newer api instead.


zealot/util/utils.ts
```diff
@@ -54,3 +55,23 @@ export function uniq(array: any[]) {
   }
   return u
 }
+
+export const nodeStreamToWebStream = (
```
mason-fish, line 59: Not using this yet, but will need it in the brimcap plugin when we pipe the `brimcap analyze` output to the backend's /add endpoint

nwt (approved):

PR #1732, opened 2021-07-06
Visualize a type value as a string

### LINKED ISSUE #1725: Result of typeof(this) is presented as "[object Object]"
Repro is with Brim commit 4cb8e61 which uses Zed commit `e17ba2f`.

Back in the last Brim GA release `v0.24.0` and up until right before the merge of #1589, type definitions returned by `typeof(.)` (now `typeof(this)` is preferred) were shown in the app. Granted, they were mere strings, but it still provided a starting point for type exploration. An example with the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default) in `v0.24.0`:

Then starting with Brim commit 6078840 (associated with the merge of #1589) this stopped working. At first it started showing the `typeof:type`, then a short while later at Brim commit f7dabb8 (associated with the merge of #1603) and since then on up to current Brim commit 4cb8e61, it's showing `[object Object]`.

My recollection is that one of the expected follow-ons from #1603 would be to make use of the granular typing information now being returned from the backend to print the type definitions with more structure than they were in the past. Perhaps this is still on the to-do list, and I respect that a comprehensive design & UX might be a significant undertaking. However, in looking toward assembling the next GA Brim release, we've already started encouraging Zed users to look at `typeof(this)` as part of exploring their data, so users are likely to bump into this after working with Zed at the CLI and then trying the same in the app, and hence would likely see this as a bug/regression.

Short of doing something fancy, is there a low cost way to bring back something minimal for MVP0, e.g., even as minimal as the former string-like presentation?

jameskerr:
This is a case where we are using a "type" as a "value". That means we need to visually represent a "type" in a cell in the viewer. I looked at what zq does for this and tried to copy.

```
{
    _path: "ssl",
    typeof: ({_path:string,ts:time,uid:bstring,id:{orig_h:ip,orig_p:port=(uint16),resp_h:ip,resp_p:port},version:bstring,cipher:bstring,curve:bstring,server_name:bstring,resumed:bool,last_alert:bstring,next_protocol:bstring,established:bool,cert_chain_fuids:[bstring],client_cert_chain_fuids:[bstring],subject:bstring,issuer:bstring,client_subject:bstring,client_issuer:bstring,validation_status:bstring,ja3:bstring,ja3s:bstring})
}
```

To implement this, I've added a `toString()` method to every zed type. I also created an abstract `BasePrimitive<T>` class to represent the expected interface of a primitive type, and to implement 2 methods that are the same for all primitives (`toString()` and `serialize()`).

image

fixes #1725

### DIFF
--- a/zealot/zed/types/type-record.ts
+++ b/zealot/zed/types/type-record.ts
@@ -1,5 +1,5 @@
 import {isNull} from "../utils"
-import {RecordType, Value} from "../../zjson"
+import {RecordFieldType, RecordType, Value} from "../../zjson"
 import {ZedContext} from "../context"
 import {typeId} from "../utils"
 import {Field} from "../values/field"
@@ -18,13 +18,25 @@ export class TypeRecord implements ContainerTypeInterface {
     this.fields = fields
   }
 
-  static stringify(fields) {
+  static stringify(fields: RecordFieldType[] | null) {
     if (isNull(fields)) return "null"
     let s = "{"
     let sep = ""
     fields.forEach((f) => {
       // XXX need to check if name has funny chars
       s += sep + f.name + ":" + typeId(f.type)
+      sep = ","
+    })
+    s += "}"
+    return s
+  }
+
+  toString() {
+    if (isNull(this.fields)) return "null"
+    let s = "{"
+    let sep = ""
+    this.fields.forEach((f) => {
+      s += sep + f.name + ":" + f.type.toString()
       sep = ","
     })
     s += "}"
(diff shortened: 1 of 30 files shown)


zealot/zed/types/type-record.ts
```diff
@@ -31,6 +31,18 @@ export class TypeRecord implements ContainerTypeInterface {
     return s
   }
 
+  toString() {
```
mason-fish, line 34: 👍

mason-fish (approved):
lgtm, nice that extending instead of implementing the primitive ended up cleaning a lot of repeated code

PR #1733, opened 2021-07-07
Refresh with new span

### LINKED ISSUE #1728: Clicking "Refresh" is not showing the latest data
Repro is with Brim commit 478f85c via prerelease macOS build https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-478f85cc.0/macos/Brim-0.25.0-prerelease-478f85cc.0.dmg and the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default).

Starting from a fresh app install, the repro steps as shown in the attached video:

1. Create a new pool via `zapi create -p sampledata`
2. Click on the Pool and observe the expected lack of data
3. Hit "Enter" in the query input box, which feels like a no-op since there's still no data
4. Post the `ftp` log that has the narrower time span 17:15-17:20 via `zapi load -p sampledata ~/work/zed-sample-data/zeek-default/ftp.log.gz`
5. Decline the "Refresh" button that's offered
6. Post the `dns` log the has the wider time span 17:15-17:36 via `zapi load -p sampledata ~/work/zed-sample-data/zeek-default/dns.log.gz`
7. Click the "Refresh" button that's offered

https://user-images.githubusercontent.com/5934157/123691811-e8169a00-d80a-11eb-8ea5-8ce6a82ba019.mp4

At the end, I was expecting the final "Refresh" would show me the full time span & data, but instead it seems "stuck" on showing me no data until I click on the Pool name in the Pools list or select "Whole Pool" from the time drop-down.

Note that the "hitting Enter" in step 3 is significant. If I skip that step, I _do_ see all the data after the final Refresh.

mason-fish:
fixes #1728 

@jameskerr and I discussed this a bit and decided that an even better fix would be to add a new "empty pool" page which would not present any of the toolbar/searchbar/span-picker tools we currently show since those are not of any use on an empty pool anyway, and instead would assist users with how to ingest data into their empty pool. Much like github does for empty repositories.

In the meantime, however, here is a quick fix that simply checks if the last search's span is the empty span default that we set for empty pools (which is the first milisecond of unix time/1970), and if it is then we will use the fresh full span of the now-populated pool instead.   

### DIFF
--- a/src/js/components/CommitNotification.tsx
+++ b/src/js/components/CommitNotification.tsx
@@ -8,6 +8,7 @@ import Current from "../state/Current"
 import {useHistory} from "react-router"
 import {SpanArgs} from "../state/Search/types"
 import {decodeSearchParams} from "../../../app/search/utils/search-params"
+import {Span} from "../brim"
 
 export default function CommitNotification() {
   const dispatch = useDispatch()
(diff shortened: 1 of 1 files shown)


philrz:
FWIW, I just tested this out in the branch and the fix seems functionally :+1:  to me.

jameskerr (approved):
Yup. This makes sense to me.

PR #1734, opened 2021-07-08
Provide explicit values to the pool search url

### LINKED ISSUE #1727: Unexpected refresh of histogram and time pickers
Repro is with Brim commit 478f85c via prerelease macOS build https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-478f85cc.0/macos/Brim-0.25.0-prerelease-478f85cc.0.dmg and the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default).

Starting from a fresh app install, the repro steps as shown in the attached video:

1. Create a new pool via `zapi create -p sampledata`
2. Post the `ftp` log that has the narrower time span 17:15-17:20 via `zapi load -p sampledata ~/work/zed-sample-data/zeek-default/ftp.log.gz`
3. Click on the Pool
4. Post the `dns` log the has the wider time span 17:15-17:36 via `zapi load -p sampledata ~/work/zed-sample-data/zeek-default/dns.log.gz`
5. Click the "Refresh" button that's offered

https://user-images.githubusercontent.com/5934157/123684062-d8468800-d801-11eb-9977-ff3596478289.mp4

While the final Refresh got things to the expected end state, after executing the `zapi` command in step 4, two things occurred that I did not expect:
* The histogram was redrawn before I'd taken any action with the "Refresh" button. The redraw shows the wider time span on the histogram X axis that reflects the nature of the data that was just added, but the bars don't indicate the presence of that new data. I was expecting no redraws at this point.
* The time picker values are decorated with blue dots. As a user, I wasn't sure how to interpret them. Previously the way I've interpreted the blue dots is that they show up if I've made changes to the current values in the time picker but I've not yet executed a query that brings them into effect and locks them in as the active new settings. But if was to click the blue dot instead of executing a query, I'm accustomed to seeing the value in the picker revert to what it was before I made the change. However, in this case, if I click the blue dot, it ends up getting set to the updated time span value from the most recently added data (see the next video below). I don't think that's a bad thing necessarily, but since it seems like a mismatch with how the dot behaves in other contexts, plus the tooltip doesn't give a lot of insight, it might be worthy of closer scrutiny. The two options that come to mind are: 1) The blue dots shouldn't even show up in this case, since the Refresh notification alone implicitly communicates that a new time span may be in play, or, 2) Show the newer time values along with the blue dot, and if the user clicks the dots, align to the existing behavior of reverting to the previous values, and if the user declines the Refresh, then it should probably revert the picker settings and remove the blue dots as well.

https://user-images.githubusercontent.com/5934157/123687413-c1099980-d805-11eb-8985-a3d7619bfe63.mp4

jameskerr:
When no span params are provided, all the components will
implicitly use the pool defaults (saved in redux).
When new commits arrive, the
pool span will change. This results in
some of the UI re-rendering even though the url has not changed.

The only case we want this re-rendering behavior is when we are
dragging a pcap into the app. Then we want to have an empty span
and rely on the updating defaults. (although this feels confusing and
worthy of simplifying)

By being explicit with the spanArgs in the url, we can ensure that
the UI will not re-render when the span of the pool changes in redux.

fixes #1727


mason-fish (approved):
Looks straight forward to me! Approved after green build 👍

PR #1735, opened 2021-07-08
use brimcap analyze instead of load

### LINKED ISSUE #1716: Use brimcap analyze instead of load

mason-fish:
fixes #1716 

As discussed, we have to force zealot to use `node-fetch` in order to attach a stream to a request body for uploading the output of `brimcap analyze`. With that concession, everything else is pretty straightforward. Now, we handle everything that `brimcap load` was doing in the app:

* analyze pcap
* add data to pool
* commit all added data to pool
* generate a pcap index

Also bumps the brimcap dep to commit 6c52930


jameskerr (approved):
Right on!

PR #1736, opened 2021-07-09
Remove Security Summary Components

jameskerr:
There is alot of code that does not seem like it will be used for a while. When trying to add new features, this codebase has been getting in the way. I deleted it all here. If we need to reference it, we can look back in time to this PR, but I'd be when we do make a dashboard feature it will be different enough from what we had here.


mason-fish (approved):
Nice! I'm sad to see the summary work go away but it's nice to simplify the code base too. 👍

Issue #1724, opened 2021-06-24
Configurable decimal/thousands separator


philrz:
Right now Brim presents integer types U.S.-style with the `,` as the "thousands separator".

![](https://user-images.githubusercontent.com/5934157/123338134-26efdb80-d4fd-11eb-8c16-82fee856f7d6.png)

  
Similarly, float types Brim use a `.` as the "decimal separator".

![](https://user-images.githubusercontent.com/5934157/123338212-48e95e00-d4fd-11eb-9b83-377bfbf3e371.png)

A wikipedia article goes into extensive detail on this subject and shows a nice table of [examples](https://en.wikipedia.org/wiki/Decimal_separator#Examples_of_use). My main take-away from this is that if we wanted to satisfy all potential international users, we'd probably want to have a setting in **Preferences** so the user could make their own custom picks for what should be used for each of these.

Part of what motivated filing this issue right now is that there was also a community beta tester in a [public Slack thread](https://brimsec.slack.com/archives/C024JHN96SC/p1624570036146100) that voiced a preference to be able to remove the thousands separator entirely, since in their case they were looking closely at the `alert.signature_id` field in Suricata data and are accustomed to seeing these presented without commas. Therefore, in addition to picking between different characters like `,` or `.`, it therefore sounds like we'd want to have the option to show no separator at all. I confirmed in the conversation that having a single global setting would meet their needs (i.e., no requirement at this time to vary the presentation on a per-named-field basis).

(If someone's interested in cleaning up all related issues at once, see also #1552).

jameskerr:
Cool, that makes sense.

philrz:
Verified in Brim commit 2e3c791.

As shown in the attached video, the user can now go into the **Preferences** settings and change the thousands separator and decimal separator to the character of their choosing. They also can choose to empty these out entirely, such as to suit the desires of the user that originally inspired the issue.

https://user-images.githubusercontent.com/5934157/130139834-0f4001d2-ccf1-4bd8-8a06-eff626792c82.mp4

While testing, I did happen to notice that an unintentional side effect of the change is that some of the large numeric values that typically require the use of the thousands separator (such as the Suricata `flow_id` field as shown in the opening screenshots of this issue) are now displayed using scientific notation whereas they were not before.

![image](https://user-images.githubusercontent.com/5934157/130140522-b45acd23-c751-4f58-a2d3-20a086bfb6e2.png)

[...]

PR #1740, opened 2021-07-13
ConfigValueProp defaults

jameskerr:
I needed to adjust some of the way we get and set config values to implement #1724 .

1. When "getting" a config value, if the key is not set, return undefined instead of null. We were previously sometimes returning null, and other times undefined.

2. When adding config properties, loop through them at startup and set the value to the defaultValue, if there is no value already present.


mason-fish (approved):
lgtm after updating the test to expect undefined too 👍

PR #1741, opened 2021-07-13
Display Preferences

jameskerr:
Adds the thousands separator option to the preferences menu. @philrz what do you think the default should be? I'm leaning towards no separator at all because of all the numeric id fields out there. It looks weird to add commas to those. It almost seems like these options should be on a per-column, opt in, basis in the future. 

Part of #1724 

uI0nWAgjom

### DIFF
--- a/app/core/format.ts
+++ b/app/core/format.ts
@@ -2,21 +2,27 @@ import * as d3 from "d3"
 import brim from "src/js/brim"
 import {zed} from "zealot"
 
-export function formatPrimitive(data: zed.Primitive, config: object) {
+export type FormatConfig = {
+  thousands: string
+}
+
+export function formatPrimitive(
+  data: zed.Primitive,
+  config: Partial<FormatConfig> = {}
+) {
   if (data.isUnset()) return "⦻"
   if (zed.isNamed(data.type, "port")) return data.toString()
-  if (zed.isInt(data)) return formatInt(data.toString(), config)
+  if (zed.isInt(data)) return formatInt(data.toInt(), config)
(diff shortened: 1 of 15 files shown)


philrz:
@jameskerr: My instincts are to have the commas present by default. Your statement about numeric ID fields is certainly correct, but that may be largely a reflection of the network security data sources we've handled to date. As we look to a future where we hope users will bring data of all kinds, it's possible that we may have more "classic" numeric data like sums and counts. My understanding of why the comma separators became typical in pre-compute contexts was that it provided a quick way for a reader to interpret the magnitude of large values, and I expect users would appreciate still having this. The one user who spawned this specific issue seemed perfectly happy to pop into **Preferences** and adjust this setting to suit their more network-centric use case.

I agree that granular per-field controls might be nice as a fancy feature we could consider in the future. Of course, thinking of tying it to named fields makes me think of how different field names could have different meanings in different contexts, which makes me think of Pools in Zed Lakes since that's where per-field treatments truly get locked in (e.g. indexing policy). Maybe there's the potential to have "data presentation metadata" in the Pool that act as hints that a client could act on or override as they see fit. That way you could mark a field in a Pool as _not_ being intended to be presented with a separator, then both a Brim app and a `zq` that have no other preference overrides could present them that way. The same metadata approach could be used to describe things like units. Just a thought.

jameskerr:
Ok, that makes sense @philrz , I'll take your suggestion.

jameskerr:
I'm closing this in favor of a new PR that will be up shortly.

PR #1745, opened 2021-07-14
Unit Test Setup Brim

jameskerr:
This PR provides a new way to write integrationy-unit tests. For example: 

```js
const workspace = fixtures("workspace1")
const pool = fixtures("pool1")
const record = createRecord({ts: new Date(0), name: "Alice", age: 99})
const brim = setupBrim({page: "detail", workspace, pool})

beforeEach(async () => {
  brim.dispatch(viewLogDetail(record))
  brim.navTo(lakePath(pool.id, workspace.id))
})

test("right click => new search focuses search window", async () => {
  const {store, main} = brim
  render(<LogDetailsWindow />, {store})

  fireEvent.contextMenu(screen.getByText("Alice"))
  fireEvent.click(screen.getByText(/new search with this value/i))
  await flushPromises()

  const searchWindow = main.windows
    .getWindows()
    .find((w) => w.name === "search")
  expect(searchWindow.ref.focus).toHaveBeenCalled()
})
```

Let's find a name for these type of tests. Component tests? App tests? 

The most interesting files in this PR are `test/unit/helpers/setup-brim` and `src/js/components/LogDetailsWindow/test.tsx`. Everything else supports those files.


package.json
```diff
@@ -72,7 +72,7 @@
     "electron": "^11.2.1",
     "electron-builder": "^22.9.1",
     "electron-builder-notarize": "^1.2.0",
-    "electron-mock-ipc": "^0.3.9",
+    "electron-mock-ipc": "brimdata/electron-mock-ipc#master",
```
mason-fish, line 75: Are we forking this bc we need unreleased features from upstream?

test/unit/helpers/utils.ts
```diff
@@ -0,0 +1,7 @@
+export function onPage(name: string) {
+  window.history.replaceState(null, `Testing Page: ${name}`, `${name}.html`)
+}
+
+export function flushPromises() {
```
mason-fish, line 5: I see this used in the test you provided, would you mind explaining why we need it?

src/js/components/LogDetailsWindow/test.tsx
```diff
@@ -0,0 +1,32 @@
...
+const pool = fixtures("pool1")
+const record = createRecord({ts: new Date(0), name: "Alice", age: 99})
+const brim = setupBrim({page: "detail", workspace, pool})
+
+beforeEach(async () => {
+  brim.dispatch(viewLogDetail(record))
+  brim.navTo(lakePath(pool.id, workspace.id))
+})
+
+test("right click => new search focuses search window", async () => {
```
mason-fish, line 20: Love how much simpler it is to test this! 👍

mason-fish (approved):
Looks great! Left a couple questions. Also that CI failure looks just like the one I hit last week where the release build was not running at all, and I had to run the macOS release in order to get a hint at what was failing (which, in that particular case, was a dev-dep that should have been a regular dep)

package.json
jameskerr: Correct, I opened this issue: https://github.com/h3poteto/electron-mock-ipc/issues/361

Then found a way to fix it here: https://github.com/h3poteto/electron-mock-ipc/pull/363

Once that gets merged and a new release is cut, we can remove the fork..

test/unit/helpers/utils.ts
jameskerr: Yeah, it's for when there are pending promises out there in the app that can't be awaited here in the test. This says, wait for all the pending promises to be fulfilled in the next event loop. 

This happens with our mock-ipc calls. A promise is returned when we invoke something, then the next tick, the promise is fulfilled. This test needs to wait for those ipc channels to send their replies.

src/js/initializers/pluginManager.ts
```diff
@@ -56,4 +37,56 @@ export default class PluginManager {
...
+  deactivate() {
+    if ("deactivate" in this.functions) {
+      this.functions.deactivate()
+    }
+  }
+
+  private indexFile() {
+    if (process.env.BRIM_ITEST === "true") return "index.js"
+    if (process.env.NODE_ENV === "test") return "index.ts"
+    return "index.js"
```
jameskerr, line 90: This is the bug. 
* In integration tests, we want index.js
* In unit tests we want index.ts 
* In production/dev we want index.js.

PR #1747, opened 2021-07-16
Loader abort

### LINKED ISSUE #1739: "brimcap analyze" and its child processes still running after deleting Pool during pcap load
Repro is with Brim commit 8d26e6e on macOS beta artifact https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-8d26e6e4.0/macos/Brim-0.25.0-prerelease-8d26e6e4.0.dmg.

As shown in the attached video, I open the pcap https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz (after uncompressing) and right-click the Pool during import and select **Delete**. In the window where I'm running `ps`, we can see that the `brimcap analyze` process and the `zeek` and `suricata` processes it spawned are still running at this point. Eventually they do seem to exit cleanly on their own at which point a "Pool not found" error is popped up.

https://user-images.githubusercontent.com/5934157/125378575-dc1af400-e343-11eb-8ec2-b889abab4503.mp4

I assume we must be skipping some step where the `brimcap analyze` process should have been cleanly killed.

mason-fish:
fixes #1739 

Uses abort controllers and event emitters to coordinate between the `deletePool` thunk, the `ingestFiles` thunk, and the brimcap loader's `load` method. We send "SIGINT" to spawned processes so that they can gracefully terminate. This can sometimes take a little bit of time (not a lot but enough to make the UI seem like there is a short delay upon deleting a pool), so we also wrap this user action in a toast.promise which gives continuous "deleting..." feedback to the user in those cases.

### DIFF
--- a/src/js/api/registries.ts
+++ b/src/js/api/registries.ts
@@ -29,31 +29,45 @@ export class CommandRegistry {
 }
 
 interface Loader {
+  match: string
   load: (
     params: IngestParams & {poolId: string},
     onProgress: (value: number | null) => void,
     onWarning: (warning: string) => void,
-    onDetailUpdate: () => Promise<void>
+    onDetailUpdate: () => Promise<void>,
+    signal?: AbortSignal
   ) => Promise<void>
-  match: string
-  unLoad?: (params: IngestParams) => Promise<void>
+  unload?: (params: IngestParams) => Promise<void>
 }
 
 export class LoaderRegistry {
   private loaders: Loader[] = []
+  private emitter = new EventEmitter()
 
   constructor() {}
 
   add(l: Loader): void {
     this.loaders.push(l)
   }
-
   remove(l: Loader): void {
     if (this.loaders.includes(l)) remove(this.loaders, (l) => l === l)
   }
-
   getMatches(loadType: string): Loader[] {
     return this.loaders.filter((l) => l.match === loadType)
+  }
+  onWillAbort(listener: (...args: any[]) => void): Cleanup {
+    this.emitter.on("will-abort", listener)
+    return () => this.emitter.removeListener("will-abort", listener)
+  }
+  onDidAbort(listener: (...args: any[]) => void): Cleanup {
+    this.emitter.on("did-abort", listener)
+    return () => this.emitter.removeListener("did-abort", listener)
+  }
+  willAbort(handlerId: string): boolean {
+    return this.emitter.emit("will-abort", handlerId)
+  }
+  didAbort(handlerId: string): boolean {
+    return this.emitter.emit("did-abort", handlerId)
   }
 }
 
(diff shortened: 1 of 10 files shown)


src/js/api/registries.ts
```diff
@@ -29,32 +29,46 @@ export class CommandRegistry {
...
   }
-
   remove(l: Loader): void {
     if (this.loaders.includes(l)) remove(this.loaders, (l) => l === l)
   }
-
   getMatches(loadType: string): Loader[] {
     return this.loaders.filter((l) => l.match === loadType)
   }
+  onWillAbort(listener: (...args: any[]) => void): Cleanup {
```
mason-fish, line 58: @jameskerr here are the api's abort-related events. I thought having explicit `willAbort` and `didAbort` methods was better than just exposing an `emit` method so that consumers don't need to know anything about internal event names ("will-abort", "did-abort", etc.). Wdyt?

jameskerr (commented):
This is looking great!

My only hesitation is that the `api.loaders.willAbort` method in deletePool is dependent on ingestFiles listening for that event and performing the action.

I'm envisioning a method called `api.loaders.abort(handlerId)` which looks like:

```js
async abort(id) {
  const fn = this.getAbortFunction(id)
  if (!fn) return

  this.emit("will-abort")
  await fn()
  this.emit("did-abort")
}
```

Maybe this means that we save the abort functions inside the LoadersRegistry instance. We'd give that abort function in ingestFiles to the api. Then we remove them when ingest completes.

```js
// in ingestFiles.ts

const l = loaders[0]
const ctl = new AbortController()

api.loaders.addAbortFunction(handlerId, async () => {
  abortCtl.abort()
  await l.unload()
})
```

This might allow us to avoid rolling back the transaction too... no sure though.


```js
// in delete pool

if (poolHandler) {
  await api.loaders.abort(poolHandler.id)
}
```

src/js/flows/ingestFiles.ts
```diff
@@ -148,11 +149,30 @@ const executeLoader = (
 
     // only supporting one loader at this time
     const l = loaders[0]
+    const abortCtl = new AbortController()
+
+    let didAbort = false
```
jameskerr, line 154: If you wanted to avoid this variable, you could check `abortCtl.signal.aborted` 

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/aborted

jameskerr (approved):
The names of the functions and events sound better in this version. The only other thing that feels off here is that the handlerID and the abortHandler are stored in two different places. The handler id is in the store, while the abort function for that handler id is stored in a closure in an event listener on the registry. 

I think it would be easier to follow along if the handlerId and functions associated with it were stored in the same place. Either the redux store or the loaders registry.

I wonder if the loaders registry should have it's own "currentLoads" property which stores all the ingests that are taking place at that moment. 

I'll approve since I think we've talked about it enough, but these can be ideas to consider later.

PR #1751, opened 2021-07-19
HTML Dialog Error Box for Windows Pre-25 Check

### LINKED ISSUE #1742: Hyperlink from "Previous Brim Version Detected" error
As described at [https://github.com/brimdata/brim/wiki/Installation#windows-installation-v0250](https://github.com/brimdata/brim/wiki/Installation#windows-installation-v0250), right now if a Windows user installs and starts a new Brim release without explicitly uninstalling their older one (presumed `v0.24.0`), the new one will pop up an error and quit.

image

This guidance is pretty good, but could still be improved.

1. The location on the filesystem shown in the error is probably opaque to the user. in Windows, users are generally discouraged from deleting apps directly from the filesystem since the registry droppings are still left behind.

2. The guidance at https://github.com/brimdata/brim/wiki/Installation#windows-installation-v0250 is arguably more useful since it's able to describe what's going on in more detail and shows how to differentiate the old/new versions in the appropriate Windows control panel screens, which is where they should be triggering the uninstall from.

Therefore it seems like it would be helpful to replace the currently-shown filesystem path with a hyperlink to https://github.com/brimdata/brim/wiki/Installation#windows-installation-v0250. Unfortunately, my own brief research found https://github.com/electron/electron/issues/7111, which seems to imply hyperlinks may not work here. If the text in the dialog could be highlighted with the mouse for cut & paste I'd maybe be fine with just printing the link and not having it be "hyper", but unfortunately the text isn't like that. Maybe the info at that Electron issue that's now inaccurate, or maybe the tip in the comment about "building your own dialog" would not be too painful.

In any case it'd be great if someone could spend a few minutes and see what's possible. Given the number of Windows users we know we have and that they're all going to have to manually upgrade, since we have the guidance already written, it's a shame to not call their attention to it.

jameskerr:
This PR creates a module available in the main process called `htmlDialog`. It has one method which mimics the electron `dialog` module, `showErrorBox(title, content)`. However in our version, the title can be a markdown string.

I've organized the commits into nice logical chunks of changes and given detailed messages in each. Have a look through those when reviewing.

image

And this is what it looks like on windows.

Screen Shot 2021-07-20 at 5 21 40 PM

Closes #1742


philrz:
I tested this out successfully on Windows using a scratch artifact https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-0d796f2d.0/windows/Brim-Setup-0.25.0-prerelease-0d796f2d.0.exe that's based on the branch in this PR. As shown in the attached video, the pop-up offers the desired link, and clicking it opens the correct anchor in my browser.

https://user-images.githubusercontent.com/5934157/126242506-c5092225-64e5-4fce-b91a-049c61c51b3d.mp4

src/js/electron/html-dialog/dialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+import theme from "src/js/style-theme"
+import {ThemeProvider} from "styled-components"
+import {BG} from "./dialog.style"
+import {closeChannel, readyChannel} from "./index"
+
+const ROOT_ID = "root"
+/**
+ * These come from the main process
+ */
+const {title, content, id} = parseSearchParams()
```
mason-fish, line 18: So we are storing the markdown content in a search param of the url/location?

jameskerr: Correct. Since I figured the content of dialog boxes will be short enough to fit into a url. The only other option is to request the content with ipc once the window has rendered.

mason-fish: Okay, yeah that seems a bit limiting but I also agree it's unlikely we will have much to say in the dialogs anyway. Will make a mental note of it too.

PR #1756, opened 2021-07-21
Use our npm package for electron mock ipc

jameskerr:
I had to fork electron-mock-ipc because they had a bug that hasn't been merged upstream yet. But in order to depend on the fork, we have to build the project and publish it somewhere. So I've created a brimdata npm org and put this package under it. Now we depend on the npm hosted package.

This might be able to fix our ci problems. CI doesn't need to download another version of electron just to build the electron-mock library.


jameskerr:
#1758 in favor of

PR #1757, opened 2021-07-21
Remove parens from type values

### LINKED ISSUE #1753: Remove parentheses around typeof(this) output
#1725 shows an example of how the the string rendering of a `typeof(this)` has changed from being just wrapped in curly braces:

before

...to now having an additional wrap of parentheses:

now

We discussed this as a group and reached consensus that we should _not_ wrap with the parens in the app.

jameskerr:
Closes #1753


mason-fish (approved):
👍

PR #1758, opened 2021-07-21
Update electron-mock-ipc

jameskerr:
The changes have been merged upstream! Now we don't depend on the fork.

This will take the place of #1756


mason-fish (approved):
Glad we didn't need the fork for long! :)

PR #1759, opened 2021-07-22
Remove unnecessary test setup

jameskerr:
Now that we call "setupBrim()" when we write component tests, we don't need these files.


mason-fish (approved):
👍

PR #1760, opened 2021-07-23
default loader: handle error status

### LINKED ISSUE #1749: Warnings are no longer surfaced when garbage data is imported (regression at 6078840)
Repro is with Brim commit 404ed5e. It seems there may multiple problems happening, so I expect this issue may need to be moved to the Zed repo and/or one or more separate Zed issues opened.

In the past (such as in the verification notes at https://github.com/brimdata/brim/issues/1070#issuecomment-781431459) when garbage data was imported into the app (such as the attached [passwd.txt](https://github.com/brimdata/brim/files/6833326/passwd.txt), which is a macOS `/etc/passwd` from a GitHub Actions VM) the user would be offered a Warnings pop-up that explained the failure and showed the Zed backend's failed attempts to read it as any of the supported auto-detect formats. However, the attached video taken with Brim commit 404ed5e shows that the warnings no longer appear.

https://user-images.githubusercontent.com/5934157/126007329-ed59eec6-c9c2-4e1e-bc7b-98fd7fea5b80.mp4

Via binary search, I narrowed down that the warnings disappeared starting with Brim commit 6078840, which is associated with #1589 (cc: @mattnibs). Knowing a Zed change triggered what's seen in the app, here's more Zed findings:

1. `zq` provides the output we'd expect.

```
$ zq -version
Version: v0.29.0-422-gb00db736

$ zq -z passwd.txt 
passwd.txt: format detection error
	tzng: line 1: bad format
	zeek: line 11: bad types/fields definition in zeek header
	zjson: line 1: invalid character '#' looking for beginning of value
	zson: zson syntax error
	zng: zng type ID out of range
	csv: auto-detection not supported
	json: auto-detection not supported
	parquet: auto-detection not supported
	zst: auto-detection not supported
```

2. If I have a `zed lake serve` running and `zapi load` the file to it, I get an HTTP 500 error. I'm not a REST guru, but @mattnibs has me conditioned to think that HTTP 500 errors should usually be something more specific, so this may be its own issue.

```
$ zapi create -p foo
pool created: foo

$ zapi load -p foo passwd.txt 
passwd.txt: format detection error
	tzng: line 1: bad format
	zeek: line 11: bad types/fields definition in zeek header
	zjson: line 1: invalid character '#' looking for beginning of value
	zson: zson syntax error
	zng: zng type ID out of range
	csv: auto-detection not supported
	json: auto-detection not supported
	parquet: auto-detection not supported
	zst: auto-detection not supported
status code 500: pool "foo": internal error: serialize transaction: empty transaction
```

3. As you could see in the DevTools of the attached video, the console _did_ show the HTTP 500 error had occurred, though this never made it through to where the user would typically see it. While maybe HTTP 500 errors should be rare, I'm inclined to think they should be surfaced to the user regardless.

4. If I sniff the port `9867` traffic during the attempted import ([port9867.pcapng.gz](https://github.com/brimdata/brim/files/6833371/port9867.pcapng.gz) attached) I do see the HTTP 500 error, but I don't see the auto-detect failure messages, so it seems like we'd not even be equipped to surface the detail even if the app were equipped to present the error.

mason-fish:
fixes #1749 

When this logic got moved into a loader plugin, the error case handling didn't come with it. Putting it back now. When the zed service changes it's error message it should get passed through the app properly now instead of being swallowed.

### DIFF
--- a/plugins/log-loader/index.ts
+++ b/plugins/log-loader/index.ts
@@ -21,6 +21,8 @@ export const activate = (api: BrimApi) => {
     // @ts-ignore
     for await (const {type, ...status} of stream) {
       switch (type) {
+        case "Error":
+          throw new Error(status.error)
         case "UploadProgress":
           onProgressUpdate(status.progress)
           await onDetailUpdate()


jameskerr (approved):
Easy enough!

PR #1762, opened 2021-07-27
Fetch stats when initializing a pool

### LINKED ISSUE #1743: "Migration complete" pop-up is premature (regression at d38a617)
I bumped into this one while testing with an artifact for the most recent beta release, but for ease of repro/verification the repro recipe below has been done with a Dev install of Brim commit 5b47168.

The repro uses [https://storage.googleapis.com/brimsec-public-test-data/brim-1743/brimdir.tar.gz](https://storage.googleapis.com/brimsec-public-test-data/brim-1743/brimdir.tar.gz) which is the saved copy of `~/Library/Application Support/Brim` on my Mac laptop after having run Brim `v0.24.0`, imported 
[ifconfig.pcapng.gz](https://github.com/brimdata/brim/files/6811353/ifconfig.pcapng.gz) and https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz (both after uncompressing), then quit the app.

To use this in a repro, such as the one below that shows the problem in Brim commit 5b47168, from the base of my Brim checkout I do:

```
git checkout 5b471683d84225959c5b4bc6d68879cb5ed4f9be && npm install && rm -rf run && tar xzvf ~/brimdir.tar.gz && mv Brim run && npm start
```

As shown in the attached video, the key to the repro is pouncing on the "wrccdc.pcap" Pool as soon as the "Migration complete" pop-up appears. As a user, I assume the pop up is communicating that all of my data should now be present. Because of this, I'm confused because when I click on it, I see the default "1970" values in the time picker that's synonymous with an empty Pool. It seems that if I continue staring at the empty-looking Pool, the status never changes. However, as soon as I take some action in the app (such as going to the time range selection drop-down with intent to check what happens when I select **Whole Pool**), now I get the "More data is now available" pop-up, and clicking its Refresh option shows me the expected data. So, it could be worse. But between that behavior and the possibility that bigger legacy Spaces could have the app in this transition state for a while, I see a potential for confusion with the "Migration complete" pop-up appearing early.

https://user-images.githubusercontent.com/5934157/125512556-5b6adeed-43e1-4e2b-8ccc-e7ec3051dced.mp4

I narrowed down via binary search that this behavior changed starting at Brim commit d38a617, which is associated with #1676 (cc: @mattnibs). As a basis for comparison, here's how it looked at Brim commit ddf38ea right before that PR merged, when pouncing on that "wrccdc.pcap" Pool did not trigger the issue.

https://user-images.githubusercontent.com/5934157/125512976-adc683c6-bbaf-42d6-a4b8-4b2c52df1ede.mp4

jameskerr:
**What happened?**

Each time the app got an update from the migration process, it would call "refreshSpaceNames". This would ask zqd for all the names of the spaces, then it would request each of their "stats" from the stats endpoint. 

The stats endpoint for the wccdc pool was returning null with 204 No Content (see screenshot) @mattnibs is that a timing race? So when you clicked on it after the migration was complete (it actually didn't matter how quickly you clicked it), you got the default time span (1 second in 1970).

When you click on a pool in the side bar, it "should have" checked that stats are up to date, but it did not.

**The Fix**

Do not request pool stats in "refreshSpaceNames", but do request pool stats in "initPool". 

**History:** The `GET /pool/:id` endpoint used to include the stats, but they have since been moved to `GET /pool/:id/stats`

Fixes #1743

### DIFF
--- a/src/js/flows/refreshPoolNames.ts
+++ b/src/js/flows/refreshPoolNames.ts
@@ -1,33 +1,22 @@
-import {Thunk} from "../state/types"
+import {BrimWorkspace} from "../brim"
 import Current from "../state/Current"
 import Pools from "../state/Pools"
+import {Thunk} from "../state/types"
 import {getZealot} from "./getZealot"
-import {BrimWorkspace} from "../brim"
-import interop from "../brim/interop"
 
+/**
+ * This only gets the list of names from the server, all the
+ * details like the min_time and max_time need to be requested
+ * somewhere else. (like in initPool)
+ */
 export default function refreshPoolNames(
   ws?: BrimWorkspace
 ): Thunk<Promise<void>> {
   return (dispatch, getState) => {
     const zealot = dispatch(getZealot(ws))
-
-    let pools = []
-    return zealot.pools
-      .list()
-      .then((data) => {
-        pools = data || []
-        return Promise.all(
-          pools.map(async (pool, i) => {
-            const stats = interop.poolStatsPayloadToPool(
-              await zealot.pools.stats(pool.id)
-            )
-            pools[i] = {...pool, ...stats}
-          })
-        )
-      })
-      .then(() => {
-        const id = ws?.id || Current.getWorkspaceId(getState())
-        if (id) dispatch(Pools.setPools(id, pools))
-      })
+    return zealot.pools.list().then((pools) => {
+      const id = ws?.id || Current.getWorkspaceId(getState())
+      if (id) dispatch(Pools.setPools(id, pools))
+    })
   }
 }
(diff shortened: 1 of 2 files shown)


src/js/flows/refreshPoolNames.ts
```diff
@@ -1,33 +1,22 @@
...
 import {getZealot} from "./getZealot"
-import {BrimWorkspace} from "../brim"
-import interop from "../brim/interop"
 
+/**
+ * This only gets the list of names from the server, all the
+ * details like the min_time and max_time need to be requested
+ * somewhere else. (like in initPool)
+ */
 export default function refreshPoolNames(
```
mason-fish, line 12: nice, the logic matches the function name better now too imo

mason-fish (approved):
lgtm

mattnibs:
@jameskerr s/zqd/lake service

PR #1763, opened 2021-07-27
Increase Unit Test Wait Timeout

jameskerr:
I made a module in `app/core/env.ts` that holds all the methods needed for inquiring about the environment.

```js
env.isCI
env.isDevelopment
env.isTest
env.isIntegrationTest
env.isMac
env.isLinux
env.isWindows
```

I also increased the @testing-library timeout setting from 1 second to 5 seconds.

This is to fix some CI flakiness.

### DIFF
--- a/test/unit/setup/after-env/index.ts
+++ b/test/unit/setup/after-env/index.ts
@@ -3,3 +3,9 @@ import "@testing-library/jest-dom"
 import "./mock-network"
 import "./polyfill-dom"
 import "./polyfill-fetch"
+import {configure} from "@testing-library/react"
+import env from "app/core/env"
+
+if (env.isCI) {
+  configure({asyncUtilTimeout: 5000})
+}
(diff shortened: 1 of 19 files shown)


test/unit/setup/after-env/index.ts
```diff
@@ -3,3 +3,9 @@ import "@testing-library/jest-dom"
 import "./mock-network"
 import "./polyfill-dom"
 import "./polyfill-fetch"
+import {configure} from "@testing-library/react"
+import env from "app/core/env"
+
+if (env.isCI) {
+  configure({asyncUtilTimeout: 5000})
```
mason-fish, line 10: just curious if this significantly increases the time it takes for tests to run when they are failing?

mason-fish (approved):
lgtm, just left one q

test/unit/setup/after-env/index.ts
jameskerr: Good question. This is only for our component tests. If there are many of them and they all are failing then it would increase the time by 5x. If that happens we can have CI "bail early" if a single test fails (or if 5 or 10 fail).

PR #1765, opened 2021-07-28
Thousands Separator + Migration Preferences State

jameskerr:
The remaining preferences that were not stored in the `configPropValues` state slice have been migrated. The thousands separator preferences has been added.

All preferences should work the same way they did before.


src/js/components/Preferences/Preferences.tsx
```diff
@@ -83,17 +112,14 @@ export default function Preferences() {
   const [f, setForm] = useCallbackRef<HTMLFormElement>()
   const [errors, setErrors] = useState([])
 
-  // TODO: convert remaining prefs into configs via a 'core' plugin
-  const prefsForm = usePreferencesForm()
```
mason-fish, line 87: Nice, it's complete now! 👏

src/js/state/migrations/202005181133_addTimeFormatPref.test.ts
```diff
@@ -1,12 +0,0 @@
-import Prefs from "../Prefs"
```
mason-fish, line 1: Sure we want to remove this? For better or worse, I thought these migrations were effectively immutable and need to stay in the app as is.

mason-fish (approved):
lgtm, just left one Q about the migrations which I'm assuming there is a good reason, so still approving

src/js/state/migrations/202005181133_addTimeFormatPref.test.ts
jameskerr: The migrations themselves shouldn't be touched, but this is a migration test file. I figure if this old migration has been working until now, we can delete the test. The test incorrectly relied on code that might change like the selector functions in Prefs. So rather than rewrite it, I just thought we could remove it .

jameskerr:
Before I merge this, I want to test that a user's old preferences from version 24 for data dir, time zone, and time format, do in fact get migrated when they upgrade to this branch.

jameskerr:
I just tested this and the preferences did get migrated. Along the way I found a bug #1773 , but this PR is good to go.

PR #1766, opened 2021-07-28
Add types to hover

jameskerr:
This seems like an obvious place to highlight the type system. When you hover over a cell, you'll see the name of the cell along with the type.

image

image

image


philrz:
I just tried this out in the branch and it looks great! :+1:

mason-fish (approved):

PR #1772, opened 2021-08-15
remove logs.post() and logs.postPaths() from Zealot

nwt:
The corresponding Zed API routes are going away, so replace calls
with logs.add()/logs.commit().

### DIFF
--- a/plugins/log-loader/index.ts
+++ b/plugins/log-loader/index.ts
@@ -13,17 +13,17 @@ export const activate = (api: BrimApi) => {
   ): Promise<void> => {
     const files = params.fileListData.map((f) => f.file)
     const totalBytes = files.reduce((sum, file) => sum + file.size, 0)
-    let readBytes = 0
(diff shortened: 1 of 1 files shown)


plugins/log-loader/index.ts
```diff
@@ -53,4 +49,17 @@ export const activate = (api: BrimApi) => {
   })
 }
 
+function nodeJSReadableStreamFromReadableStream(
```
nwt, line 52: Style-wise, should this anonymous (`const nodeJSReadableStreamFromReadableStream = ...`) rather than named?

nwt: Also, I'd like to acknowledge that this is ugly. If there's a way to avoid it that I missed, let me know!

jameskerr: Nope, named is still cool.

jameskerr (commented):
Thanks for converting this!

jameskerr (approved):

PR #1776, opened 2021-08-17
Enable Strict Type Checking

jameskerr:
This is a PR that will slowly be worked on over time. Right now there are 2169 typescript errors in this branch that will need to be addressed by hand. Anyone who wants to do some mindless busywork, pull the branch and take a few errors.

This should be rebased periodically with main.


jameskerr:
Closing this in favor of a more piece by piece approach.

PR #1782, opened 2021-08-19
activate fix

### LINKED ISSUE #1770: Activating Brim on Mac with no existing windows does nothing
This issue was reported by a user in a [thread on public Slack](https://brimsec.slack.com/archives/C024JHN96SC/p1628609137008200).

It's expected that a new search window would be opened if there are no open brim windows. This is how most mac apps work.

This code will be handled in the on "activate" event in the main process.

I imagine this is related to the new hidden window.

mason-fish:
fixes #1770 

Part of the activation handler was still conditional to there being zero windows rendered, which should ideally never exist since there should always be at least a hidden window. This adjusts the logic accordingly. I also found a way to render multiple hidden windows using this activation flow, so I've included some protection around that as well.

### DIFF
--- a/src/js/electron/brim.ts
+++ b/src/js/electron/brim.ts
@@ -21,6 +21,7 @@ import tron, {Session} from "./tron"
 import formatSessionState from "./tron/formatSessionState"
 import {sessionStateFile} from "./tron/session"
 import windowManager, {$WindowManager} from "./tron/windowManager"
+import log from "electron-log"
 
 type QuitOpts = {
   saveSession?: boolean
@@ -69,7 +70,9 @@ export class BrimMain {
   }
 
   activate() {
-    if (this.windows.count() === 0) this.windows.init()
+    log.info("activate fired! window count: ", this.windows.count())
+    if (this.windows.count() === 0 || this.windows.count() === 1)
+      this.windows.init()
   }
 
   async resetState() {
--- a/src/js/electron/brim.ts
+++ b/src/js/electron/brim.ts
@@ -21,7 +21,6 @@ import tron, {Session} from "./tron"
 import formatSessionState from "./tron/formatSessionState"
 import {sessionStateFile} from "./tron/session"
 import windowManager, {$WindowManager} from "./tron/windowManager"
-import log from "electron-log"
 
 type QuitOpts = {
   saveSession?: boolean
--- a/src/js/electron/brim.ts
+++ b/src/js/electron/brim.ts
@@ -70,9 +70,7 @@ export class BrimMain {
   }
 
   activate() {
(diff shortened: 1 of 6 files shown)


src/js/electron/brim.ts
```diff
@@ -69,7 +70,9 @@ export class BrimMain {
   }
 
   activate() {
-    if (this.windows.count() === 0) this.windows.init()
+    log.info("activate fired! window count: ", this.windows.count())
+    if (this.windows.count() === 0 || this.windows.count() === 1)
+      this.windows.init()
```
jameskerr, line 75: Let's add a method called `this.windows.getVisible().length` I think that would help this read more clearly.

src/js/electron/tron/windowManager.ts
```diff
@@ -207,6 +207,12 @@ export default function windowManager(
       }
     },
 
+    openHidden() {
+      // only open hidden window if one doesn't already exist
+      if (this.getWindows().find((w) => w.name === "hidden")) return
+      this.openWindow("hidden")
```
jameskerr, line 213: Let's make methods called

`getAll()`

`getVisible()`

`getHidden()`

Then we could use them in all the places that make sense.

jameskerr (changes requested):

src/js/electron/tron/windowManager.ts
```diff
@@ -47,7 +47,7 @@ export default function windowManager(
   return {
     init() {
       // hidden renderer/window is never persisted, so always open it with rest
-      this.openWindow("hidden")
+      this.openHidden()
```
jameskerr, line 50: Maybe this could be called `ensureHiddenRenderer`?

src/js/electron/tron/windowManager.test.ts
```diff
@@ -52,3 +52,17 @@ test("when all closed waits until windows are done", (done) => {
...
+  manager.openHidden()
+  let windows = manager.getWindows()
+  expect(windows).toHaveLength(1)
+  expect(windows[0].name).toEqual("hidden")
+
+  // try again, still expect only 1
+  manager.openHidden()
+  windows = manager.getWindows()
+  expect(windows).toHaveLength(1)
+  expect(windows[0].name).toEqual("hidden")
```
jameskerr, line 67: Nice!

src/js/electron/brim.ts
```diff
@@ -69,7 +70,9 @@ export class BrimMain {
   }
 
   activate() {
-    if (this.windows.count() === 0) this.windows.init()
+    log.info("activate fired! window count: ", this.windows.count())
```
jameskerr, line 73: Need this log line?

jameskerr (approved):
Right on!

PR #1801, opened 2021-08-31
Document how temporary storage is used

philrz:
When working on an issue-at-scale recently with a community user, I became concerned that maybe our software was consuming too much temporary storage on their system. This made me realize that we're overdue to reveal this information so users have a chance of doing this kind of debug on our own, or at least I could easily point them to an article if this comes up again. As it turns out, I think this info grafts nicely onto the existing [Filesystem Paths](https://github.com/brimdata/brim/wiki/Filesystem-Paths) article.

While running tests to write this content, I actually found multiple bugs where we're _not_ currently cleaning up after ourselves the way we should (brimdata/brimcap#136, brimdata/zed#2983). Normally in these situations I'd confess these issues in the article, link to them, and then update the article once they get fixed. However, I have high hopes we'll get these fixed pretty quick, so for now I've kept it tidy.


docs/Filesystem-Paths.md
```diff
@@ -111,3 +112,66 @@ it may prove necessary during [[Troubleshooting]] or if you use
...
+a larger pcap, the extracted packets are placed in a temporary file that's
+opened in Wireshark. For this, Brim uses a file with a name of the format
+`packets-[date+timestamp].pcap` (e.g., `packets-2020-03-09T17_27_53.399273Z.pcap`).
+This file is _not_ deleted automatically, as Brim does not make any assumptions
+about what you may continue to do with the pcap file outside the app.
+
+Standard conventions are used for creating these directories/files in the
+appropriate location for each supported platform. As a result, temporary
+storage should always be freed up automatically by the OS during normal
+operations, such as when performing a system restart. However, bugs or abrupt
```
nwt, line 145: > As a result, temporary storage should always be freed up automatically by the OS during normal operations, such as when performing a system restart.

I don't think Windows does this.

docs/Filesystem-Paths.md
```diff
@@ -111,3 +112,66 @@ it may prove necessary during [[Troubleshooting]] or if you use
...
+shutdowns to Zed/Brimcap processes (e.g., `kill -9` or similar) may prevent
```
nwt, line 146:
```suggestion
shutdowns of Zed/Brimcap processes (e.g., `kill -9` or similar) may prevent
```

docs/Filesystem-Paths.md
```diff
@@ -111,3 +112,66 @@ it may prove necessary during [[Troubleshooting]] or if you use
...
+is used. In a default test environment we've found this variable is already
+set and is typically of a format like
+`C:\Users\%USERNAME%\AppData\Local\Temp` or
+`C:\Users\%USERNAME%\AppData\Local\Temp\2`.
+
+* **macOS** - The base path specified in the `$TMPDIR` environment variable is
+used. In a default test environment we've found this variable is already set
+and is typically of a format like `/var/folders/yn/jbkxxkpd4vg142pc3_bd_krc0000gn/T/`.
+
+* **Linux** - A base path specified in a `$TMP` environment variable would be
```
nwt, line 166:
```suggestion
* **Linux** - A base path specified in a `$TMPDIR` environment variable would be
```

nwt (approved):

docs/Filesystem-Paths.md
philrz: @nwt: Thanks for mentioning this! My bad... I'd checked on macOS and Ubuntu and made a bit of "QED" leap. Indeed, I just ran a quick test and confirmed that Windows does not clean `%TMP%` on a reboot. Then I was reminded of the many free (or paid!) "cleanup" tools for Windows that purport to address this pressing challenge. :roll\_eyes: 

As if that's not bad enough, this made me paranoid enough that I figured I couldn't necessarily assume that what I observed for Ubuntu could hold for other Linux distros, so I tried CentOS. I figured if I covered at least those two I could call it a trend. Sure enough, CentOS didn't clean `/tmp` between reboots either!

I've now pushed a commit that's more vague about how reboots or "common cleanup utilities" might reap the temp space, but no guarantees. I know I'm probably overcompensating by saying any of this, but since in light of recent bugs we know Zed has the capacity to chow a whole lot of temp space, I'd like to be able to say we'd warned the people _somewhere_...

docs/Filesystem-Paths.md
philrz: @nwt: Oops! Thanks for catching!

docs/Filesystem-Paths.md
philrz: Oops! Thanks for catching @nwt!

PR #1802, opened 2021-08-31
Zed update through brimdata/zed#2976

nwt:
brimdata/zed#2976 contains two Zed API changes.
1. For pool creation, the `POST /pool` response now describes both the pool and the default branch (called main).
2. For loading data, `POST /pool/{pool}/{branch}` replaces the combination of `POST /pool/{pool}/add` followed by `POST /pool/{pool}/staging/{commit}`.

### DIFF
--- a/src/js/flows/ingestFiles.ts
+++ b/src/js/flows/ingestFiles.ts
@@ -61,14 +61,14 @@ const validateInput = (files: File[], poolNames) => ({
 
 const createPool = (client: Zealot, gDispatch, workspaceId) => ({
   async do(params: IngestParams) {
-    const create = await client.pools.create({name: params.name})
+    const {pool, branch} = await client.pools.create({name: params.name})
     gDispatch(
       Pools.setDetail(workspaceId, {
-        ...create.pool,
+        ...pool,
         ingest: {progress: 0, warnings: []}
       })
     )
-    return {...params, poolId: create.pool.id, branchId: create.branch.id}
+    return {...params, poolId: pool.id, branchId: branch.id}
   },
   async undo({poolId}: IngestParams & {poolId: string}) {
     await client.pools.delete(poolId)
--- a/src/js/flows/ingestFiles.ts
+++ b/src/js/flows/ingestFiles.ts
@@ -61,15 +61,14 @@ const validateInput = (files: File[], poolNames) => ({
 
 const createPool = (client: Zealot, gDispatch, workspaceId) => ({
   async do(params: IngestParams) {
-    const pool = await client.pools.create({name: params.name})
+    const create = await client.pools.create({name: params.name})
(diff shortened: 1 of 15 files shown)


src/js/flows/ingestFiles.ts
```diff
@@ -61,15 +61,14 @@ const validateInput = (files: File[], poolNames) => ({
 
 const createPool = (client: Zealot, gDispatch, workspaceId) => ({
   async do(params: IngestParams) {
-    const pool = await client.pools.create({name: params.name})
+    const create = await client.pools.create({name: params.name})
```
jameskerr, line 64: Not a blocker, but an FYI, but you could do a destructure on this object like so: `const {pool, branch} = await ...`

mason-fish (approved):
This lgtm overall, thanks @nwt!

There are a couple pieces that I thought were a little unusual but they all ultimately relate to the api:

1) Using custom headers ("Zed-Commit") to send resource-specific data is a little different than what I'm used to, why don't we include the commit as part of the url? either the path or a query param? I'm not sure if there is anything ultimately abusive about using custom headers, though I do know that some http proxies can strip unknown ones. 
2) the response to create a pool is a "pool" object with with another pool key nested in it. Why not flatten that part in the response since the api route effectively names it? This would also help with the naming issue I saw you changed (calling the response `create` so we can access the pool id with `create.pool.id`)

src/js/flows/ingestFiles.ts
nwt: Oh, that's much nicer! So much so that I did it in 7cc0fb5d11a6d551e4aea6a6d43a599b888fafdf.

nwt:
> * Using custom headers ("Zed-Commit") to send resource-specific data is a little different than what I'm used to, why don't we include the commit as part of the url? either the path or a query param? I'm not sure if there is anything ultimately abusive about using custom headers, though I do know that some http proxies can strip unknown ones.

This is admittedly peculiar. @mccanne, @mattnibs, and I also considered putting this commit metadata in a query parameter or in the request body (using multipart), but there was a desire to avoid both of those, so we settled on the header. If it becomes a problem, we can always change our approach.

> * the response to create a pool is a "pool" object with with another pool key nested in it. Why not flatten that part in the response since the api route effectively names it? This would also help with the naming issue I saw you changed (calling the response `create` so we can access the pool id with `create.pool.id`)

Strictly speaking, the response is now a [`BranchMeta`]( https://github.com/brimdata/zed/blob/f2cae0deaa01d8dc1ad168fc80a7c6cf3863c1e9/lake/root.go#L226-L229) describing both the pool and its default branch. I followed @jameskerr's destructuring suggestion to address the naming issue, but flattening is certainly reasonable too, though I'd prefer to do that separately.

Issue #1666, opened 2021-05-20
Redundant requests at end of import


philrz:
While investigating the repro of #1656, @nwt spotted what appeared to be redundant requests coming from Brim.

![](https://user-images.githubusercontent.com/5934157/119019693-49b53000-b952-11eb-8cff-1fc5d792dac2.png)

These were shown to @jameskerr at a team meeting and he agreed to take a closer look.

philrz:
There was an exchange in a PR review (https://github.com/brimdata/brim/pull/1814#issuecomment-914721344) where it was speculated that this issue might have been fixed. I've circled back and checked, and based on what I see in the attached video, it looks like the issue is still with us. Even though the app has moved from the "search" endpoint to the "query" endpoint, there's still redundant requests I can see in DevTools, specifically, two of the `| every 1s count() by _path` and two of the `| head 500` (cc: @jameskerr).

https://user-images.githubusercontent.com/5934157/135920305-c743822c-3775-4902-b7e0-98fe2fe918b0.mp4

philrz:
Verified in Zui commit a9d64be.

While I've not tracked down the precise commit(s) that led to the resolution, the data import workflow has changed significantly in the time since this issue was opened and the problem has gone away during that overhaul. Now the load of data is its own operation, then the **Query Pool** button can be clicked to see the initial splash of the loaded data, and it's at that step that the queries like the ones shown above are issued. Repeating the equivalent repro steps as shown in the attached video, the redundant queries we saw in the past are no longer present.

https://github.com/brimdata/zui/assets/5934157/36c1a619-68c0-4569-abbf-a150b9f161a3

Specifically, the unique queries we see issues here were:

```
"from :pools | id == 0x12b072de308ff796eeb49df7e78c2f5a6646be33 or name == \"0x12b072de308ff796eeb49df7e78c2f5a6646be33\""
"from 'wrccdc.pcap'\n | !has(ts) | count()"
"from 'wrccdc.pcap'\n | ts == null | count()"
"from 'wrccdc.pcap'\n | ts != null | count() by time := bucket(ts, 500ms), group := _path | sort time"
"from 0x12b072de308ff796eeb49df7e78c2f5a6646be33 | min(ts), max(ts)"
"from 'wrccdc.pcap'\n | count()\n  | { i: count(), v: this}\n  | i > 0\n  | head 500\n  | yield v"
[...]

PR #1814, opened 2021-09-08
Kill Search & Aborted Searches

### LINKED ISSUE #1800: "Kill search" seems to result in an infinite spinner
Repro is with Brim commit a1dfa38 which used Zed commit 78ce875.

Before starting the attached video, I've already imported the `dns.log.gz` from the full 70 GB wrccdc data set, and that file consists of 246 MB of GZIP'ed Zeek TSV logs. Then in the video I start a heavyweight operation `sort uid` and then click "Kill search" from the "..." menu. As you can see, the spinner keeps going for 3+ minutes and never stops before I stop capturing video, making it appear to me as a user as if the kill was not successful and the search is still running. As the video also shows, no errors are logged in the DevTools console.

https://user-images.githubusercontent.com/5934157/131406352-a7f12464-b523-4c8f-aa58-af6230e3c1ce.mp4

If allowed to run to completion, this same operation on this laptop takes just a little over a minute. This indicates that after having clicked "Kill search" the app was truly in some hung state rather than just waiting on the operation to complete.

https://user-images.githubusercontent.com/5934157/131406670-795f4dd8-21bf-4999-aca1-f440315480e6.mp4

jameskerr:
So this was a few bugs...

The bug was that the "Kill Search" button would sometimes not work. This was because the main search function registers a handler for the search with an id. If an id is reused, that means cancel the previous search and run this fresh one. It also was because as soon as a request was aborted, that "handler id" was removed. Also, whenever the search ends (aborted succeeded or failed), the code removes the handler with that id as well. So the bug was in this sequence of events.

1. Click search button
2. Abort handler with id = "Table" (also removes the handler)
3. Search and register handler to id = "Table"
4. Before it completes....
5. Click search button again (before first finishes)
6. Abort handler with id = "Table" (removes the handler)
7. Search and register handler to id = "Table"
8. The first search search's "finally" cause is run and removes handler with id = "Table"

At this point there is no handler for the currently running search. If you try to click the Kill search button, nothing will happen.

This PR address this by creating a "tag" for an "abortable" and and unique id. The start of the search aborts other things with the same tag, while the finally block removes that specific abortable by id.

We also we not emitting an "ABORT" status which left the search status in "FETCHING" forever. This was why the spinner never stopped.

If there is a ticket that showed duplicated searches when the space was loaded, this will fix that issue too. 

Here is a video showing everything working now.

https://user-images.githubusercontent.com/3460638/132425440-cd519344-b78c-4fee-86f1-455cd3937f7d.mp4

fixes #1800

### DIFF
--- a/app/core/context.tsx
+++ b/app/core/context.tsx
@@ -1,0 +1,31 @@
+import React, {useContext, ReactElement} from "react"
+import {Provider} from "react-redux"
+import BrimApi from "src/js/api"
+import AppErrorBoundary from "src/js/components/AppErrorBoundary"
+import {Store} from "src/js/state/types"
+import theme from "src/js/style-theme"
+import {ThemeProvider} from "styled-components"
+
+const BrimApiContext = React.createContext<BrimApi | null>(null)
+
+export function useBrimApi() {
+  const value = useContext(BrimApiContext)
+  if (!value) throw new Error("No Brim Api Provided")
+  return value
+}
+
+export function BrimProvider(props: {
+  store: Store
+  api: BrimApi
+  children: ReactElement
+}) {
+  return (
+    <AppErrorBoundary>
+      <BrimApiContext.Provider value={props.api}>
+        <Provider store={props.store}>
+          <ThemeProvider theme={theme}>{props.children}</ThemeProvider>
+        </Provider>
+      </BrimApiContext.Provider>
+    </AppErrorBoundary>
+  )
+}
--- a/app/core/models/abortables.ts
+++ b/app/core/models/abortables.ts
@@ -14,8 +14,17 @@ type NewAbortable = Omit<Abortable, "id">
 export class Abortables {
   registry: Abortable[] = []
 
-  abort(predicate?: Partial<Abortable>) {
-    this.filter(predicate).forEach((a) => a.abort())
+  async abort(predicate: string | Partial<Abortable>) {
+    if (isString(predicate)) {
+      const a = this.get(predicate)
+      if (a) return await a.abort()
+    } else {
+      return Promise.all(this.filter(predicate).map((a) => a.abort()))
+    }
+  }
+
+  async abortAll() {
+    return Promise.all(this.all().map((a) => a.abort()))
   }
 
   add(a: Abortable | NewAbortable) {
@@ -25,11 +34,11 @@ export class Abortables {
   }
 
   all() {
-    return this.registry
+    return [...this.registry]
   }
 
   filter(predicate?: Partial<Abortable>) {
-    if (!predicate) return this.registry
+    if (!predicate) return this.all()
     return this.registry.filter(this.matchFn(predicate))
   }
 
--- a/app/core/models/abortables.ts
+++ b/app/core/models/abortables.ts
@@ -1,0 +1,54 @@
+import {nanoid} from "@reduxjs/toolkit"
+import {isString, remove} from "lodash"
+
+export type Abortable<Meta extends any = any> = {
+  id: string
+  tab?: string
+  tag?: string
+  meta?: Meta
+  abort: () => void | Promise<void>
+}
+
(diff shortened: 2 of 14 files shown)


philrz:
@jameskerr: Regarding your question:

> If there is a ticket that showed duplicated searches when the space was loaded, this will fix that issue too.

Are you thinking of #1666?

jameskerr:
@philrz Yeah I was thinking of that one. I need to verify that though.

app/core/models/abortables.ts
```diff
@@ -0,0 +1,54 @@
...
+  id: string
+  tab?: string
+  tag?: string
+  meta?: Meta
+  abort: () => void | Promise<void>
+}
+
+type NewAbortable = Omit<Abortable, "id">
+
+export class Abortables {
```
mason-fish, line 14: This interface seems great, I like the predicate options. My only hesitancy is that it might read more clearly and be a little safer if we have separate and explicit methods for the most destructive abort/remove all options (e.g. `.abort` would require it's predicate instead of being optional, and then there would be a separate `.abortAll` method).

app/core/models/abortables.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.registry.push({id, ...a})
+    return id
+  }
+
+  all() {
+    return this.registry
+  }
+
+  filter(predicate?: Partial<Abortable>) {
+    if (!predicate) return this.registry
```
mason-fish, line 32: Safer to return a ref to a copy of the registry, instead of a ref to the registry itself when there is no predicate so it can't be tampered with externally. Then in the case where there is a predicate that's what filter is already doing.

app/core/context.tsx
```diff
@@ -0,0 +1,31 @@
+import React, {useContext, ReactElement} from "react"
+import {Provider} from "react-redux"
+import BrimApi from "src/js/api"
+import AppErrorBoundary from "src/js/components/AppErrorBoundary"
+import {Store} from "src/js/state/types"
+import theme from "src/js/style-theme"
+import {ThemeProvider} from "styled-components"
+
+const BrimApiContext = React.createContext<BrimApi | null>(null)
```
mason-fish, line 9: Ah cool, I like this! I've actually been looking for an opportunity to use the context api more in our app and this seems like the perfect use case.

mason-fish (approved):
Left a couple comments but nothing worth blocking, nice work with the abortables!

app/core/models/abortables.ts
```diff
@@ -0,0 +1,54 @@
+import {nanoid} from "@reduxjs/toolkit"
+import {isString, remove} from "lodash"
+
+export type Abortable<Meta extends any = any> = {
+  id: string
+  tab?: string
+  tag?: string
+  meta?: Meta
+  abort: () => void | Promise<void>
```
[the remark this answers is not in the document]
jameskerr: I like this idea. How about the individual abort function that gets added can return void or a promise, but when calling abort through the container class, it always returns a promise. For example:

```js
// This way we don't need to wrap this in a promise everywhere
abortables.add({id: "1", abort: () => ctl.abort})

// But this will always be a promise
const promise = abortables.abort("1")
```

app/core/models/abortables.ts
jameskerr: Agreed. I'll update this.

app/core/models/abortables.ts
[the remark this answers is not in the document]
mason-fish: Yeah, that sounds great 👍

app/core/context.tsx
jameskerr: Sweet!

app/core/models/abortables.ts
jameskerr: Agreed! I'll add abort all. And make abort require a predicate.

PR #1836, opened 2021-09-15
query annotation and test fixes

### LINKED ISSUE #1835: CI failure due to dropped trailing zero on timestamp
Since the merge of #1792, we've had intermittent failures in the Brim CI. I created a branch and triggered the CI several times and found it failed 6 out of 10 times. I think I can see the pattern of what's wrong.

The failures look like this:

```
FAIL app/search/flows/viewer-search.test.ts
  ● a normal response › zealot gets the request

    expect(received).toEqual(expected) // deep equality

    Expected: "from '1' | ts >= 2021-09-15T01:15:19.060Z | ts <= 1970-01-01T00:00:00.001Z | dns query | head 500"
    Received: "from '1' | ts >= 2021-09-15T01:15:19.06Z | ts <= 1970-01-01T00:00:00.001Z | dns query | head 500"
```

Peeking at the code at [https://github.com/brimdata/brim/blob/0dcf10c452efba1de841a512b38dc1c424f01e25/app/search/flows/viewer-search.test.ts#L62](https://github.com/brimdata/brim/blob/0dcf10c452efba1de841a512b38dc1c424f01e25/app/search/flows/viewer-search.test.ts#L62) we see the `from` value starts life as a response from `new Date()`, which returns the _current_ date and time, and hence is going to vary on each run, which explains the non-determinism. While I don't claim to understand all aspects of the test code, assuming the "Received" value is coming from the backend in some way, we can see by hitting Zed directly how it drops trailing zeros like this.

```
$ zq -version
Version: v0.30.0-5-g5c4868de

$ echo '{mytime: 2021-09-15T01:15:19.060Z}' | zq -
{mytime:2021-09-15T01:15:19.06Z}
```

I'm guessing that for the runs that succeed there's no trailing zero in the current timestamp.

mason-fish:
fixes #1835

The problem with the missing `0`s we see in the indeterminate test is not the backend, it is the `dateToNanoTs` function. It's not entirely a problem so much as an inconsistency: 

The current default behavior is to not include trailing `0`s, which is an acceptable format for the timestamp as the backend parses it. In our tests, we use the date `.toISOString()` method to compare expected values, and that method always includes the full millisecond width, which revealed the formatting difference.

The fix I'm proposing here then is to ensure the decimal width returned by `dateToNanoTs` is at least the full ms, trailing zeros or not, while µs and ns widths may vary by trailing zeros.

Also proposed here is a temporary workaround for the query annotator, as discussed with @nwt: If a query looks like it begins with any sort of `from ` then we will *NOT* annotate it at all and just pass it on as is to the request. This way we can expose the feature of querying from multiple pools while still supporting the existing/expected search behavior. Ultimately, we should discuss the "pool search" vs "lake search" (searching across multiple pools in a lake) workflows a bit more to come up with a better solution for this.  

### DIFF
--- a/src/js/flows/search/mod.ts
+++ b/src/js/flows/search/mod.ts
@@ -27,6 +27,8 @@ type annotateArgs = {
 }
 
 export const annotateQuery = (query: string, args: annotateArgs) => {
+  // if query already starts with 'from', we do not annotate it further
+  if (/^from\s*\(?/i.test(query)) return query
   const {
     poolId,
     from = new Date(new Date().getTime() - 30 * 24 * 60 * 60 * 1000), // 30 days
@@ -52,10 +54,10 @@ const isZeroDefaultSpan = (
   )
 }
 
-const dateToNanoTs = (date: Date | Ts | bigint): string => {
+export const dateToNanoTs = (date: Date | Ts | bigint): string => {
   const NanoFormat = new DateTimeFormatterBuilder()
     .appendPattern("yyyy-MM-dd'T'HH:mm:ss")
-    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)
+    .appendFraction(ChronoField.NANO_OF_SECOND, 3, 9, true)
     .appendLiteral("Z")
     .toFormatter()
 
--- a/src/js/flows/search/mod.ts
+++ b/src/js/flows/search/mod.ts
@@ -28,7 +28,7 @@ type annotateArgs = {
 
 export const annotateQuery = (query: string, args: annotateArgs) => {
   // if query already starts with 'from', we do not annotate it further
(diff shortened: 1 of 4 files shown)


src/js/flows/search/mod.ts
```diff
@@ -27,6 +27,8 @@ type annotateArgs = {
 }
 
 export const annotateQuery = (query: string, args: annotateArgs) => {
+  // if query already starts with 'from', we do not annotate it further
+  if (/^from\s*\(?/i.test(query)) return query
```
nwt, line 31: A slightly simpler regular expression should do the job.
```suggestion
  if (/^from[\s(]/i.test(query)) return query
```

nwt (approved):

PR #1842, opened 2021-09-20
Relax the shortcuts for Zoom In and Out

### LINKED ISSUE #1818: "Zoom In" hot key should work even when not holding down Shift
This issue was last reproduced on Brim commit af1f358 and was stumbled upon by one of our own employees. :wink: 

The "Zoom In" hot key of Command-+ on macOS or Ctrl-+ on Windows currently only works if you're holding down the Shift key, such that you're typing a true `+`. Many other apps are forgiving and also recognize the hot key when Shift is not being pressed, effectively accepting Command-= on macOS or Ctrl-= on Windows.

jameskerr:
It is now possible to zoom the window with `CmdOrCtrl++` or `CmdOrCtrl+Shift++`. The same goes for zoom out. `CmdOrCtrl+-` and `CmdOrCtrl+Shift+-`

This was done by adding out own app wide shortcuts that re-implement the zoom in and out behavior for the currently focused window.

I also added a folder called `electron/initializers`. Any file in it will get required during the "start" phase of the brim app. We can put anything there that only needs to be set up in the main process once.

Fixes #1818

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11216,6 +11216,22 @@
         }
       }
     },
+    "electron-is-accelerator": {
+      "version": "0.1.2",
+      "resolved": "https://registry.npmjs.org/electron-is-accelerator/-/electron-is-accelerator-0.1.2.tgz",
+      "integrity": "sha1-UJ5RDCala1Xhf4Y6SwThEYRqsns="
+    },
+    "electron-localshortcut": {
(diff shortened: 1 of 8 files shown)


mason-fish (approved):
The new shortcuts work for me! The new initializers pattern seems cool too, especially as we more of them 👍

PR #1845, opened 2021-09-22
Safer object access in <ErrorNotice />

### LINKED ISSUE #1815: "TypeError: details.flatMap is not a function" when attempting "Delete All" of pools
This issue was reported by a community user in a [public Slack thread](https://brimsec.slack.com/archives/C024JHN96SC/p1631055087018800). They were using the beta version based on Brim commit ec5da6b. In their own words:

> I don't know if I can recreate it or not but I tried to "Delete All" from the Brim App and got this error on the screen.  There was one pool left when I restarted the application and when I tried to delete it, it couldn't find it.  Tried to "Delete All" again and got the same thing but this time all pools where removed.  
>  image
> 
> I think I had around 10 or so pools and the newest pcap I was looking at was around 400mb.

Since we didn't have a precise repro recipe, I tried importing lots of pcaps then trying "Delete All" in various permutations (e.g., varying which pool I had highlighted at the time I invoked it, which pools were open in which tabs, etc.) but I was unable to trigger the problem. Perhaps the UX team could look at the functions mentioned in the stack trace and provide some insight into what might trigger this.

mason-fish:
fixes #1815 

Just adds some type inspection to the error notice renderer. Included a component test which forces the detail string input that caused the original error described in the ticket above for good measure.

### DIFF
--- a/src/js/components/ErrorNotice.test.tsx
+++ b/src/js/components/ErrorNotice.test.tsx
@@ -1,0 +1,55 @@
+import React from "react"
+
+import logInto from "../../../test/unit/helpers/loginTo"
+import provide from "../../../test/unit/helpers/provide"
+import Notice from "../state/Notice"
+import ErrorFactory from "../models/ErrorFactory"
+import ErrorNotice from "./ErrorNotice"
+import {screen} from "@testing-library/react"
+
+test("renders Error notice with no details", async () => {
+  const {store} = await logInto("workspace1", "pool1")
+  store.dispatch(
+    Notice.set(
+      ErrorFactory.create({
+        type: "Bad",
+        message: "Test error message"
+      })
+    )
+  )
+  provide(store, <ErrorNotice />)
+  expect(screen.getByRole("alert")).toHaveTextContent("Test error message")
+})
+
+test("renders Error notice with detail string", async () => {
+  const {store} = await logInto("workspace1", "pool1")
+  store.dispatch(
+    Notice.set(
+      ErrorFactory.create({
+        type: "Bad",
+        message: "Test error message",
+        details: "Test detail"
+      })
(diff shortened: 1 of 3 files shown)


src/js/components/ErrorNotice.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import logInto from "../../../test/unit/helpers/loginTo"
+import provide from "../../../test/unit/helpers/provide"
+import Notice from "../state/Notice"
+import ErrorFactory from "../models/ErrorFactory"
+import ErrorNotice from "./ErrorNotice"
+import {screen} from "@testing-library/react"
+
+test("renders Error notice with no details", async () => {
+  const {store} = await logInto("workspace1", "pool1")
```
jameskerr, line 11: These tests are great! However, I'd like to use the `setupBrim` helpers instead of the logInto and Provide helpers. 

There is an example usage in Preferences.test.tsx and in ReleaseNotes.test.tsx

The reason is because it exercises more of the same codepaths that the app will go through in production.

src/js/components/ErrorNotice.tsx
```diff
@@ -13,7 +13,7 @@ export default function ErrorNotice() {
   const visible = useSelector(Notice.getVisible)
 
   return (
-    <NoticeBanner show={visible}>
+    <NoticeBanner role="alert" show={visible}>
```
jameskerr, line 16: Nice!

src/js/components/ErrorNotice.test.tsx
mason-fish: Done, good call 👍

jameskerr (approved):
Ship it

PR #1849, opened 2021-09-27
Filter == value in menus

### LINKED ISSUE #1839: "Filter = value" should probably be "Filter == value"
Now that the Zed shipped with GA Brim `v0.25.0` uses the newer Zed `==` syntax for equality, the right-click menu option should probably say "Filter == value" to help reinforce this to users.

jameskerr:
A simple change to update the context menu labels to match the zed syntax.

Fixes #1839


mason-fish (approved):
👍

PR #1850, opened 2021-09-27
Improve brimcap analyze error handling

### LINKED ISSUE #1820: Errors from custom Brimcap config not surfaced in Brim app
Repro is with Brim commit 7a72af0 which uses Brimcap release tagged `v1.0.4`.

As a user, here I'm following the steps in the [Custom Brimcap Config](https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config) wiki article to create my own Brimcap YAML config that points to my own custom Zeek & Suricata installs on a Linux VM. I'm at the point where I've tested out the YAML config successfully with one pcap at the shell and now I've configured my Brim app to point to the same YAML. In the attached video, I try to import yet another pcap [ifconfig.pcapng.gz](https://github.com/brimdata/brim/files/7132523/ifconfig.pcapng.gz) (after uncompressing), and there's a failure that's short on detail, saying only "No records in request."

https://user-images.githubusercontent.com/5934157/132603803-6566af84-4970-4af5-b15d-ffd7a704e119.mp4

If we drop to the shell and run a functionally equivalent set of commands to analyze & load, we see there was more detailed error info returned in the `brimcap analyze` phase that we missed out on in the app.

```
$ brimcap -version
Version: v1.0.4

$ brimcap analyze -config zeek-suricata.yml ifconfig.pcapng | zed api load -
100.0% 2280B/2280B records=0 
zeek-wrapper.sh exited with code 1
stdout: (no output)
stderr:
fatal error in <command line>, line 3: problem with trace file - (invalid interface capture length 524288, bigger than maximum of 262144)

status code 400: no records in request
```

This ends up being an error unrelated to us, I think due to the pcap libraries being too old on this VM. But since that error info didn't make it into the app, the user doesn't have much to go on for debug.

mason-fish:
fixes #1820 

When we load a pool with the brimcap plugin, we pass each specific load operation the output stream from a forked `brimcap analyze` process. If the analyze process fails the load operation will fail too, but we have not been handling the load and/or analyze errors properly. This PR aims to fix that: now if load fails, we will check for an analyze error and report that if it exists, otherwise we report the load error.

I think this is also a bit cleaner to use a single error flag, `analyzeErr` that is a dedicated error bucket for the analyze operation, as opposed to a general error bucket for the analyze, load, and index operations.

### DIFF
--- a/plugins/brimcap/brimcap-plugin.ts
+++ b/plugins/brimcap/brimcap-plugin.ts
@@ -294,9 +294,9 @@ export default class BrimcapPlugin {
       const p = this.cli.analyze(pcapFilePath, cliOpts, signal)
       this.processes[p.pid] = p
 
-      let brimcapErr
+      let analyzeErr
       p.on("error", (err) => {
-        brimcapErr = err
+        analyzeErr = err
       })
 
       const handleRespMsg = async (jsonMsg) => {
@@ -310,7 +310,7 @@ export default class BrimcapPlugin {
             if (status.warning) onWarning(status.warning)
             break
           case "error":
-            if (status.error) brimcapErr = status.error
+            if (status.error) analyzeErr = status.error
             break
         }
       }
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Got it, I understand the issue. Looks like this will fix it.

PR #1852, opened 2021-09-28
Fix bug with primitive type values

### LINKED ISSUE #1819: typeof() result not shown for primitive types
Repro is with Brim commit af1f358 using test pcap https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz. The issue was originally reported by @mccanne.

As shown in the attached video, the value of `stringType` is not being shown in the app for queries that return primitive types, such as:

```
_path=="conn" | head 1 | cut stringType := typeof(_path)
_path=="conn" | head 1 | cut stringType := typeof(orig_bytes)
```

However, it _does_ print the output for an complex types, such as:

```
_path=="conn" | head 1 | cut stringType := typeof(id)
_path=="dns" | head 1 | cut stringType := typeof(answers)
```

When equivalent queries are issued via `zapi` against the same pool, the `stringType` value is output in all these cases.

https://user-images.githubusercontent.com/5934157/132591280-f1f7f7b1-7fd7-4c21-a529-01b7ee50b194.mp4

jameskerr:
Fixes #1819

Check if the value of a type is one of the primitive keywords. If so, use the type for that primitive as the value. Here is the repro example now fixed.

image

image

### DIFF
--- a/zealot/zed/types/type-type.ts
+++ b/zealot/zed/types/type-type.ts
@@ -1,6 +1,7 @@
 import {isNull} from "../utils"
 import {TypeValue} from "../values/type"
 import {BasePrimitive} from "./base-primitive"
+import primitives from "./type-primitives"
 
 export class TypeOfType extends BasePrimitive<TypeValue> {
   name = "type"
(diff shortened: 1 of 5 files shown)


zealot/zed/types/type-type.ts
```diff
@@ -9,7 +9,11 @@ export class TypeOfType extends BasePrimitive<TypeValue> {
     if (isNull(value)) {
       return new TypeValue(null)
     } else {
-      return new TypeValue(typedefs[value])
+      if (isPrimitiveName(value)) {
```
mason-fish, line 12: 👍

mason-fish (approved):

PR #1854, opened 2021-09-28
Prevent Next Page Search When Not Needed

### LINKED ISSUE #1825: ZSON containing 2 records shows up as 3 after import
Repro is with Brim commit 7a72af0. I stumbled onto this one by accident while doing other testing.

As indicated by the name, the attached [two_flows.zson.gz](https://github.com/brimdata/brim/files/7138945/two_flows.zson.gz) test data is ZSON containing two records.

```
$ zq 'count()' two_flows.zson.gz
{count:2(uint64)}
```

As shown in the attached video, when I drag it into the app, the initial splash of events makes it look like _three_ records. However, if I query for `count()` it shows the correct count of two, and if I clear my search bar and re-execute the default query, now I see just the two events I was originally expecting.

https://user-images.githubusercontent.com/5934157/132743090-953d613e-4232-47c9-bbd5-8f40e8623503.mp4

jameskerr:
The `Viewer.endStatus` enum is used to determine if we should fetch the next page of results. It used to default to "Incomplete" which was the signal to fetch the next page. Now it defaults to "INIT" so that we don't fetch the next page unnecessarily.

Fixes #1825

This was the source of the problem. isIncomplete was set to true every time we clear the viewer.

```js
// ResultsTable.tsx

  function onLastChunk() {
    if (props.isIncomplete && !props.isFetching) {
      props.dispatch(nextPageViewerSearch())
    }
  }
```

### DIFF
--- a/src/js/state/Viewer/reducer.ts
+++ b/src/js/state/Viewer/reducer.ts
@@ -6,7 +6,7 @@ import {createSelection} from "./helpers/selection"
 
 const init = (): ViewerState => ({
   records: [],
-  endStatus: "INCOMPLETE",
+  endStatus: "INIT",
   status: "INIT",
   columns: {},
   scrollPos: {x: 0, y: 0},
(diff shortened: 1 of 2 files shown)


mason-fish (approved):
👍

PR #1856, opened 2021-09-29
Handle nested and quoted zql field paths

### LINKED ISSUE #1828: Data not shown for a quoted field name
Repro is with Brim commit 7a72af0. This issue was originally found by @mccanne.

Consider the following simplified CSV data.

```
$ cat dotted.csv 
Malic.acid
1.71
```

When turned into ZSON, the field name is quoted, presumably because the dot would normally signify a field in a nested record, but that's not the case here.

```
$ zq -version
Version: v0.29.0-512-gbcb16977

$ zq -z -i csv dotted.csv | tee dotted.zson
{"Malic.acid":1.71}
```

If this ZSON is dragged into the app, the column header appears without the quotes and the value is not displayed at all.

### LINKED ISSUE #1838: Field references added via right-clicks aren't quoted to handle special characters
Repro is with Brim commit 0dcf10c.

A community user had been working with some challenging field names, such as ones that contain space characters. brimdata/zed#1291 first added support for these at the Zed layer, and here's some examples of working with them at the CLI.

```
$ zq -version
Version: v0.30.0-7-g4f67f3ee

$ cat data.ndjson 
{"one field": "some text"}
{"another": {"in nested record": "more text"}}
{"a": {"b": {"c": "more text"}}}

$ zq -z 'this["one field"]=="some text"' data.ndjson 
{"one field":"some text"}

$ zq -z 'another["in nested record"]=="more text"' data.ndjson 
{another:{"in nested record":"more text"}}

$ zq -z 'a["b"]["c"]=="more text"' data.ndjson 
{a:{b:{c:"more text"}}}
```

As the attached video shows, right-click operations like "Filter = value" or "Count by field" are not yet able to identify when field names would need this kind of handling and wrap them with the necessary `["..."]` syntax.

https://user-images.githubusercontent.com/5934157/133533694-3b71c994-e820-4b80-ad37-0e12859ba275.mp4

mason-fish:
fixes #1838 
fixes #1828 

@jameskerr has some overlapping changes he's working on relating to the `zed.Field` that we've coordinated on. We will collaborate within this PR to address each of our tickets.

### DIFF
--- a/src/js/zql/toZql.ts
+++ b/src/js/zql/toZql.ts
@@ -4,7 +4,6 @@ import {isStringy} from "zealot/zed"
 
 const needsQuotes = (fieldName: string) => !/^[a-zA-Z_$][\w]*$/.test(fieldName)
 
-// TODO: Mason - make field type a 'zed.Field' once it has path getter
 export const toFieldPath = (arg: string | string[] | zed.Field) => {
   const result = []
   const path =
--- a/src/js/zql/toZql.ts
+++ b/src/js/zql/toZql.ts
@@ -2,7 +2,28 @@ import isString from "lodash/isString"
 import {zed} from "zealot"
 import {isStringy} from "zealot/zed"
 
+const needsQuotes = (fieldName: string) => !/^[a-zA-Z_$][\w]*$/.test(fieldName)
+
+// TODO: Mason - make field type a 'zed.Field' once it has path getter
+export const toFieldPath = (field: any) => {
+  const result = []
+  field.path.forEach((path, i) => {
+    if (needsQuotes(path)) {
+      // if first path needs quoting, use 'this' as the bracket parent
+      if (i === 0) result.push("this")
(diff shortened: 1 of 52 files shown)


src/js/zql/toZql.ts
```diff
@@ -2,7 +2,28 @@ import isString from "lodash/isString"
 import {zed} from "zealot"
 import {isStringy} from "zealot/zed"
 
+const needsQuotes = (fieldName: string) => !/^[a-zA-Z_$][\w]*$/.test(fieldName)
+
+// TODO: Mason - make field type a 'zed.Field' once it has path getter
+export const toFieldPath = (field: any) => {
```
mason-fish, line 8: Once the `zed.Field` has been modified, I'm planning on changing the test I included to test `toZql` instead and no longer export this function. I just needed a fast way to prove this works.

nwt (approved):

PR #1859, opened 2021-09-30
Convert WindowManager to Class

jameskerr:
Typescript likes classes way more that functions that return
objects with methods.


src/js/electron/brim.ts
```diff
@@ -20,21 +20,21 @@ import isDev from "./isDev"
...
-  windows?: $WindowManager
+  windows?: WindowManager
   store?: Store
   session?: Session
   lake?: Lake
 }
 
 export class BrimMain {
-  readonly windows: $WindowManager
+  readonly windows: WindowManager
```
mason-fish, line 37: Maybe we should name this member `winManager` to be more consistent its value type, sort of like we do in the tests?

mason-fish (approved):
Looks great! This also enables our IDEs to facilitate autocomplete and click-ins/usage much better 🎉 Just left one naming idea which is not a blocker

src/js/electron/brim.ts
jameskerr: Totally. I've gone back and forth on that. In the end, I like the readability of `brim.windows` more than `brim.winManager` when using the brim object.

PR #1860, opened 2021-09-30
Get Devtools Back

jameskerr:
We were in the dark ages for too long. Not being able to use redux
or react devtools. Winter is over. By upgrading electron a few minor versions
and passing the allowFileAccess: true options to our extensions,
the devtools have been resurrected.

image

image


mason-fish (approved):
Awesome! Glad it was so simple!

PR #1861, opened 2021-10-01
Fix Union Encoding

### LINKED ISSUE #1855: Brim App UX Error when using fuse command
I used shaper to take in a json file to load into Brim via zed.  Once the data was loaded into Brim I used the fuse command to show all of the header fields.  I happened to click on a field and the result was the screenshot provided.  I haven't ran into this issue with any other types of data that I've loaded into Brim, whether it's from Zed or loaded into Brim directly.

**To Reproduce**
Load the provided results.zng file into Brim and use the fuse command.  Click on an ip address and it should show the error provided in the screenshot.  There are other values that will produce the same result, not just the IP addresses.

Screen Shot 2021-09-29 at 5 33 37 AM

**ZNG File**
[results.zng.zip](https://github.com/brimdata/brim/files/7253059/results.zng.zip)

**Desktop Info**
 - OS: MacOS
 - Brim Version: 0.25.0

jameskerr:
fixes #1855 

When encoding a union value back into zjson, we were calling value.toString() instead of value.serialize(). The resulted in the string representation of the value rather than the json value. 

In the original repro, the tags field was a union of "string", "string array", and "null". When we received the data the value was: `["CobaltStrike", "HKKFGL-AS-AP HK Kwaifong Group Limited"]`. When we encoded it back it became `"[CobaltStrike,HKKFGL-AS-AP HK Kwaifong Group Limited]"`. This resulted in bad zjson being saved. When it was read and decoded, it choked on that string value and caused the error.

A test has been added that replicated the error and confirmed that it was fixed.

### DIFF
--- a/zealot/zed/types/type-map.ts
+++ b/zealot/zed/types/type-map.ts
@@ -13,15 +13,17 @@ export class TypeMap implements ContainerTypeInterface {
     return `|{` + typeId(keyType) + ":" + typeId(valType) + "}|"
   }
 
-  create(value: [Value, Value][], typedefs) {
+  create(value: [Value, Value][] | null, typedefs) {
     return new ZedMap(
       this,
-      new Map(
-        value.map((entry) => [
-          this.keyType.create(entry[0], typedefs),
-          this.valType.create(entry[1], typedefs)
-        ])
-      )
+      isNull(value)
+        ? null
+        : new Map(
+            value.map((entry) => [
+              this.keyType.create(entry[0], typedefs),
+              this.valType.create(entry[1], typedefs)
+            ])
+          )
     )
   }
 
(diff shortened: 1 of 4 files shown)


zealot/zed/types/type-map.ts
```diff
@@ -13,15 +13,17 @@ export class TypeMap implements ContainerTypeInterface {
     return `|{` + typeId(keyType) + ":" + typeId(valType) + "}|"
   }
 
-  create(value: [Value, Value][], typedefs) {
+  create(value: [Value, Value][] | null, typedefs) {
```
jameskerr, line 16: Values can always be null. This was another bug the test caught. The Map and Union types are more rare and have not been battle tested as much so it makes sense these bugs are here.

zealot/zed/values/union.ts
```diff
@@ -18,7 +18,7 @@ export class Union implements ZedValueInterface {
 
   serialize() {
     if (isNull(this.index) || isNull(this.value)) return null
-    return [this.index.toString(), this.value.toString()]
```
jameskerr, line 21: The source of the bug.

mason-fish (approved):
👍

PR #1919, opened 2021-11-01
Make sure you have the stats for a pool before rendering search components

### LINKED ISSUE #1918: Clicking large pool makes app un-responsive
In the attached video, you'll see that when clicking on the 4GB demo.pcap pool, the app becomes unresponsive.

It looks like the query to populate the chart is not working correctly. It decided to do a `count() by every 1s` over a month of data which resulted in over 12,000 data points. This is what is bringing the app down.

The code that decides what time interval to use relies on the total size of span being searched. It appears that the first time we click a pool in a window, the span is not ready before we decide on the time interval.

I'll do a binary search and find the commit that caused the slowness.

https://user-images.githubusercontent.com/3460638/139727006-fef1ee8a-0be5-46e6-a627-bfff9bd3951d.mp4

jameskerr:
The search page relies on the pool stats. Do not render it until we have them. This fixes a bug where we issue the histogram query assuming the time span is 0 because we don't have the stats yet. The histogram then tries to count every one second resulting in thousands of results that bring the app to its knees.

fixes #1918


app/search/home.tsx
```diff
@@ -60,7 +61,7 @@ export default function SearchHome() {
   const exportAction = useExport()
   const columns = useColumns()
   const pluginButtons = usePluginToolbarItems("search")
-
+  const poolId = useLakeId()
```
mason-fish, line 64: we should probably rename this hook to `usePoolId()`

mason-fish (approved):
Nice catch!

app/search/home.tsx
jameskerr: Lots of places need to be renamed. I'll make a ticket.

jameskerr: https://github.com/brimdata/brim/issues/1922

PR #1923, opened 2021-11-02
Fix qlib folder context menu

### LINKED ISSUE #1907: sub-menu bug from read-only item to regular item
**Describe the bug**
If I have a read-only query item selected then I select a regular item and then activate the sub-menu, the sub-menu initially does not let me rename/edit/delete. If I close the sub-menu and reopen it then allows me to rename/edit/delete.

**Screenshots**
read-only bug

**Desktop Info**
 - OS: macOS
 - Brim Version: v0.26.0-69-gafd04fbc

mason-fish:
fixes #1907 

Disables `Edit`, `Run Query`, and `Copy Query` if a given item is a group/folder

image

### DIFF
--- a/src/js/components/SideBar/Item.tsx
+++ b/src/js/components/SideBar/Item.tsx
@@ -129,6 +129,7 @@ export default function Item({innerRef, styles, data, state, handlers, tree}) {
   const currentPool = useSelector(Current.getPool)
   const {isEditing, isSelected} = state
   const {value, id} = data
+  const isGroup = "items" in data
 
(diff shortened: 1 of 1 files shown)


jameskerr (approved):

jameskerr:
Well, now that I'm looking at it. I think those other menu items should not even be included. The only options should be the relevant ones. I think there is a "hidden" true option that you can pass to those menu builders.

jameskerr:
Yep, it's called "visible" https://www.electronjs.org/docs/latest/api/menu-item#menuitemvisible

PR #1932, opened 2021-11-02
Relative date bug

### LINKED ISSUE #1929: Typing in relative dates breaks the app
Typing in the time range two relative times breaks the app with an error:

https://user-images.githubusercontent.com/3460638/139928575-6c4c30a2-7732-40b4-984b-d9f605901b19.mp4

Here is a screenshot of the error:

Screen Shot 2021-11-02 at 12 01 15 PM

jameskerr:
fixes #1929 

The date constructor was being given a relative string time which it should be able to handle.


mason-fish (approved):
lgtm after that little lint fix 👍

PR #1945, opened 2021-11-08
lake service events endpoint changes

mattnibs:
Support changes to lake service events endpoint.

Depends on brimdata/zed#3246


mason-fish (approved):
I checked this out and can confirm we have commit notifications back, thanks @mattnibs !

Issue #1921, opened 2021-11-01
"More data available" popup not appearing


jameskerr:
To reproduce, create a new pool with zapi, click on that pool in the app (it will be empty), then load data into that pool with zapi load.

```
zapi create bigdata
zapi use -p bigdata
zapi load any-data.zng
```

You would expect to see a popup, but not appears.

This was working in version v0.25.0.

![image](https://user-images.githubusercontent.com/3460638/139753693-59f95a8a-2d0a-4f05-9b6b-1bb6341713dd.png)

PR #1949, opened 2021-11-08
Request zjson format from query endpoint 

### LINKED ISSUE #1946: Request zjson format for pool list/get
We've decided to embrace our own model more by using zjson instead of json as the response format for requests through the `/query` endpoint when listing and getting pool metadata. We can then continue to use zealot to unpack those values for use within the app.

mason-fish:
fixes #1946 
adds integration tests to ensure #1921 is resolved but depends on https://github.com/brimdata/brim/pull/1945

The itest is not related to this specific format change other than how the commit-notification breakage (which it tests) encouraged us to re-examine the requested response format change that is included here. That changes consists of requesting and unpacking zjson into js pool config objects instead of unpacking json.


jameskerr (approved):
Awesome. That test is easy to read.

PR #1956, opened 2021-11-12
Update auth0

mason-fish:
The main changes involved here are the following:

* we no longer assume or create defaults for protocol (`http` vs `https`) or ports.
* the `/events` zealot method is now facilitated with a fetcher.source which we wrap inside the `createBrimFetcher` layer, which is where we set auth credentials to requests based on workspace/lake
* the /auth/method endpoint wraps it's response in a describe block, so we needed to pull out the contents from it's `value` key. I expect this will not be necessary once some of the new zealot client behaviors (`resp.js()`) are introduced.
* with the protocol assumption removal mentioned above, we also run a migration to prepend all workspace hosts with "http://", since zealot no longer automatically does this.


zealot/util/utils.ts
```diff
@@ -8,7 +8,7 @@ function hasOwnProperty<X extends {}, Y extends PropertyKey>(
 }
 
 export function url(host: string, path: string) {
-  return `http://${host}${path}`
+  return `${host}${path}`
```
nwt, line 11: Maybe change `host` to `baseUrl` or some other URL-based name?

src/js/components/WorkspaceModals/WorkspaceForm.tsx
```diff
@@ -94,15 +94,15 @@ const WorkspaceForm = ({onClose, workspace}: Props) => {
...
+      label: "Lake URL",
       check: (value) => {
         if (isEmpty(value)) return [false, "must not be blank"]
         let isValid = true
         if (!isNewWorkspace && isDefaultWorkspace(workspace)) {
           const {host, port} = workspace
           isValid = value === host || value === [host, port].join(":")
         }
-        return [isValid, "cannot change host of default workspace"]
+        return [isValid, "cannot change lake url of default workspace"]
```
nwt, line 105: s/url/URL/ since this is user-facing.

zealot/zealot.ts
```diff
@@ -39,19 +34,16 @@ async function responseToPoolConfigs(
 }
 
 export function createZealot(
-  hostUrl: string,
+  host: string,
```
nwt, line 37: Would something like `baseUrl` would be a better name?

zealot/fetcher/fetcher.ts
```diff
@@ -1,72 +1,61 @@
...
+        headers
+      })
       if (!resp.ok) {
         const content = await parseContentType(resp)
         return Promise.reject(createError(content))
       }
       const iterator = createIterator(resp, args)
       return createStream(iterator, resp)
     },
-    async upload(args: FetchArgs): Promise<ZResponse> {
```
jameskerr, line 37: Can this be removed now? How do the progress updates happen now? The events endpoint?

mason-fish: Yeah we haven't been using this since I think we switched to use the load endpoint. There, we use nodeFetch and attach the request body as a stream then listen to the backend for progress updates.

src/js/state/getPersistable.ts
```diff
@@ -24,6 +24,7 @@ export function getWindowPersistable(state: State) {
     delete draft.workspaceStatuses
     delete draft.toolbars
     delete draft.configs
+    delete draft.modal
```
mason-fish, line 27: I threw this in bc I found a bad RSOD (red screen of death) bug when quitting and starting up again or refreshing the app if a modal is open. There is probably another solution (the error had to do with being unable to locate the modal dom element by it's #id) but I figured that we probably don't care to persist modals anyway.

zealot/fetcher/fetcher.ts
```diff
@@ -1,72 +1,108 @@
...
-export function createFetcher(host: string) {
+const fetchWithTimeout = async (
+  baseUrl: string,
+  args: FetchArgs
+): Promise<Response> => {
+  const {path, method, body, signal, headers, useNodeFetch} = args
+  const [wrappedSignal, clearTimeout] = useTimeoutSignal(signal)
+  if (body instanceof stream.Readable) {
+    body.once("data", () => clearTimeout())
+    body.once("start", () => clearTimeout())
```
mason-fish, line 30: This additional "start" listener is kind of a bummer, but was necessary for the brimcap stream pipeline where the stdout stream's data can take some time to begin to produce data, while the stderr stream's data progress gives us a quicker status report. Clearing the same timeout more than once has no realistic consequence.

nwt (approved):

zealot/fetcher/fetcher.ts
jameskerr: Sweet, thanks for the explanation.

jameskerr (approved):

Issue #1922, opened 2021-11-01
Rename "Workspace" and "Lake"


jameskerr:
We used to call "lakes" "workspaces" and we used to call "pools" "lakes". It's very confusing in certain files to see the names switched up. This ticket is a reminder to rename all those files, functions and variables. 

We should also write a migration to change all the URLs stored in the tab histories state, because the urls have the words "workspace" and "lake" in them too.

mccanne:
I also noticed that the list of workspaces goes away on restart.  They should be persistent.  I think of them like slack teams.  You shouldn't have to rejoin your slack teams every time you restart the app.

mccanne:
> I also noticed that the list of workspaces goes away on restart. They should be persistent. I think of them like slack teams. You shouldn't have to rejoin your slack teams every time you restart the app.

I realize the app implements the correct behavior in terms of keeping the connection list around but  I think it got removed when the service went down.  We should look into this.

jameskerr:
@mason-fish this is all done now?

mason-fish:
@jameskerr Yep!

PR #1957, opened 2021-11-13
change "Add Workspace..." to "Add Lake..."

mccanne:
This is an incomplete fix for issue #1922 but we are getting this
small change in now to immediately improve the cloud onboarding
terminology.  This will create some inconsistency in the docs that
refer to Workspace instead of Lake and throughout the code until
issue #1922 is addressed, but that should happen soon.


mason-fish (approved):
I changed a few more of the user facing instances of "workspace"

mattnibs:
Verified this works as of v0.28.0!

PR #1963, opened 2021-11-15
fix non-integer Ts.ns in dateToTs()

nwt:
The nanosecond calculation in dateToTs() sometimes returns a non-integer
value, leading to a RangeError when time().toBigInt() tries to convert
the value to a BigInt.  Fix by using Math.round() instead of
Number.prototype.toFixed() in the calculation.

Fixes CI failures like https://github.com/brimdata/brim/runs/4216942185?check_suite_focus=true#step:19:27.


jameskerr (approved):
Thanks @nwt

PR #1978, opened 2021-11-22
Add node version file for nodenv users

jameskerr:
In case anyone (me) uses nodenev, this file will make sure we're using the correct version of node.


mason-fish (approved):
👍

PR #1981, opened 2021-11-24
Zed Inspector

### LINKED ISSUE #1851: vertical values in log view
It should be easy to pivot between a horizontal "log like" view of data values (as is currently implemented in the log view) and a vertical "pretty -Z" style of data values.

The vertical value cells would have precisely the same formatting and styling as the horizontal value cells but would be displayed in an indented arrangement as with `zq -Z`.

There are many examples of JSON viewers out there that have collapsable views of embedded records. A lot of these UIs are poorly designed but the idea of collapsable elements is important especially for large, complex nested values. We could start out without any collapse-ability and add this functionality in a subsequent issue.

_Side note: we should have a new name for the "log view". Perhaps the "data view"? Or the "data browser"_

jameskerr:
The Zed Inspector is able to easily view heterogeneous data with complex and nested values. 

# Try It Out

This is a large feature that needs time to mature before we release it, so it's behind a feature flag. To run it, checkout this branch, then 
```
npm run start -- --feature-flags=zed-inspector
```

The zed data is passed to the `<Inspector />` component in the _values_ prop. That component transforms the array of zed values into an array of _RowData_. That row data is passed into the windowing component to be virtualized. The expanded state is kept in a JavaScript Map as _value => boolean_. 

**Collateral Changes:**

Some of this large diff are from these.

* All the Zed value classes had to have their TypeScript types changed from a union of concrete class types to an interface type.
* Moved all the Zed type colors into a single css file.
* Added Tab.**useState**(_key_, _defaultValue_) as a way to easily put temporary state variables in the Tab section of the redux store. It can be used just like React.useState.

**Known Bugs:**

* When the values array is large, the UI is sluggish. This is because the transformation from values into row data happens each time a value is expanded.
* Expanding a type expands all instances of that type in the results. This is because there is only ever one instance of a unique zed type and the expanded state is stored in a map where the zed value is the key. Since type values all share the same identity, expanding in one place expands them everywhere.

**To Do:** 
- [x] Auto switch to table view when the data shares a single shape.
- [x] Make sure each child has a unique key. We're still getting no-key errors.

Closes #1851 

https://user-images.githubusercontent.com/3460638/143324099-62287300-1e62-433c-8383-cd059b8e7aac.mp4


app/features/inspector/inspector.tsx
```diff
@@ -0,0 +1,102 @@
...
+      field: null,
+      key: null,
+      last: true,
+      // @ts-ignore
+      type: v.type
+    })
+  }
+
+  return (
+    <List
```
mason-fish, line 65: Would be nice to add an empty state here if there are no rows to render

mason-fish (approved):
Wow awesome work! Looks amazing 🔥 I like the code layout a lot: making those iterators with the generator functions is a good pattern.

Left a couple small comments. Perhaps in a later iteration, it would be nice to find a way to sync or at least remember scroll positions of each view so users can go back and forth without losing their place.

app/features/inspector/inspector.tsx
```diff
@@ -0,0 +1,102 @@
...
+    },
+    setExpanded: (value: zed.Value, bool: boolean) => {
+      const newMap = new Map(expanded.entries())
+      newMap.set(value, bool)
+      setExpanded(newMap)
+    }
+  }
+
+  for (const v of props.values) {
+    inspect({
```
mason-fish, line 53: This ultimately ends up building containers and annotating the ctx with them, but the verb "inspect" sounds more like a read-only action to me. Wdyt of a diff name like "inspectAndBuild", "generateContainers", or "annotateRows"?

app/features/inspector/inspector.tsx
jameskerr: Good point.

app/features/inspector/inspector.tsx
jameskerr: Hmm, let's talk more about this live.

PR #1984, opened 2021-12-03
Stricter Typescript

jameskerr:
Instead of fixing all 2000 errors that appear when enabling strict: true, let's do it in parts. This PR enables some of the easy strict options. The next PR will be strictNullChecks, and the last will be noImplicityAny.

Enables the following options in our typescript config:

* strictFunctionTypes
* strictBindCallApply
* noImplicitThis
* alwaysStrict

We had about 30 errors that needed to be fixed as a result of turning these one.


jameskerr:
Integration tests failed with what looks like a flaky test. Only on mac, this was the one integration test failure:

```
FAIL test/integration/tests/zed-events.test.ts (35.394 s)
  ● Handle Zed server events › pool-new/update/delete

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      42 |     await zealot.pools.delete(id)
      43 |     await brim.withImplicitTimeout(500, async () => {
    > 44 |       expect(await brim.isNotVisible(poolItem)).toBe(true)
         |                                                 ^
      45 |     })
      46 |   })
      47 | })

      at _callee3$ (tests/zed-events.test.ts:44:49)
      at tryCatch (../../node_modules/regenerator-runtime/runtime.js:45:40)
      at Generator.invoke [as _invoke] (../../node_modules/regenerator-runtime/runtime.js:281:22)
      at Generator.prototype.<computed> [as next] (../../node_modules/regenerator-runtime/runtime.js:106:21)
      at asyncGeneratorStep (tests/zed-events.test.ts:15:103)
      at _next (tests/zed-events.test.ts:17:194)
```

zealot-old/fetcher/callbacks.ts
```diff
@@ -18,7 +18,7 @@ export class Callbacks {
     this.callbacks = new Map()
   }
 
-  add(name: EventName, cb: (args?: ZealotPayloadValue) => void) {
+  add(name: EventName, cb: (args?: any) => void) {
```
mason-fish, line 21: Probably not too important here since we are going to be shifting to the rest of the new zealot client soon, but are we planning on adding strictness to no-any as we tighten up our typescript?

src/js/electron/tron/migrations.ts
```diff
@@ -15,62 +15,56 @@ type VersionedData = {
...
-  args: Args = {from: 0}
-): Promise<Migrations> {
-  let cv = parseInt(args.from.toString())
-
-  const files = await lib.file(dir).contents()
-  const migrations = files
-    .filter(onlyMigrations)
-    .map(build)
-    .sort((a, b) => a.version - b.version)
+export class Migrations {
```
mason-fish, line 18: Classy 😎

mason-fish (approved):

zealot-old/fetcher/callbacks.ts
jameskerr: Yeah that's why I thought it was find to be any here. We are planning to switch to no "implicit" any. An explicit any like this will still be allowed, but not advised.

PR #1985, opened 2021-12-07
Upgrade Electron to v14. Replace Spectron with Playwright.

### LINKED ISSUE #1934: electron upgrade

mason-fish:
fixes #1934

Lot's of changes here:

Replacing our integration test framework from Spectron to Playwright: Includes 3 of our original spectron test suites (queries, ingest, and zed-events). Unfortunately there is still a lot of inconsistency with running this on Windows (only in CI, I have not seen a single failure when running locally on my PC) but since we are so desperate to upgrade electron (which requires ditching Spectron) then we can test Windows manually in our pre-release cadence for now and tackle the CI issues more later.

Upgrading electron from 11 to 14: There are a few small changes involved here but the most noteworthy is that the `remote` module has been extracted out of the `electron` package and is now it's own npm scoped module at `@electron/remote`. I followed their provided migration instructions for this transition and the biggest blocker I hit was changing the way we mock that in our unit tests.

There are very few changes to the app itself other than the upgrade-related api changes. I did end up synchronizing/awaiting the `window.loadFile()` uses and re-order the launch of the hidden window. I did this so that the app would open more deterministically which then allowed the new integration tests to perform more consistently. I also re-worked some of the file input logic in the ingest flow so that we use the provided `File` object instead of the filepath. This was also for testing purposes since Playwright does not currently set the filepath on `File` objects the way that Electron does (and for this reason, in tests, we do NOT run or test the `brimcap index` sub-operation facilitated by the brimcap plugin). Finally, we also have upgraded to node 14! I found no issues with that aspect of the upgrade.


.github/workflows/ci.yml
```diff
@@ -65,68 +65,12 @@ jobs:
...
-        with:
-          name: linux-release
-          path: linux-release
-      - name: Integration Tests (Centos 8)
-        run: |
-          dnf install -y ./linux-release/Brim*.rpm
-          xvfb-run -d -s "-screen 0 1280x1024x24" npm run itest -- --ci --forceExit
-
-  integration_test:
+  playwright_integration_test:
```
jameskerr, line 68: Are we still testing centos?

app/core/env.ts
```diff
@@ -1,5 +1,5 @@
-import {app as electronApp, remote} from "electron"
-const app = electronApp || (remote && remote.app)
+import {app as electronApp} from "electron"
+const app = electronApp
```
jameskerr, line 2: This can run from the renderer process, so we'll need a way to tell the renderer what the environment looks like. We can make an ipc call and set a global.

src/js/electron/isDev.ts
```diff
@@ -1,6 +1,6 @@
-import {app as electronApp, remote} from "electron"
+import {app as electronApp} from "electron"
 
-const app = electronApp || (remote && remote.app)
+const app = electronApp
```
jameskerr, line 3: This also may run from the renderer process.

test/shared/data/index.ts
```diff
@@ -1,5 +1,6 @@
 import path from "path"
 import fs from "fs-extra"
+import {ReadableStream} from "web-streams-polyfill/ponyfill"
```
jameskerr, line 3: I added a web-file-polyfill for this in my own tests. You'll see it in the upcoming PR.

test/unit/__mocks__/@electron/remote.ts
```diff
@@ -0,0 +1,6 @@
+import {app as mockApp} from "../electron"
+export const app = mockApp
+
+export const dialog = {
+  showMessageBox: jest.fn()
+}
```
jameskerr, line 6: Perfect.

jameskerr (approved):
Let's go!!!

.github/workflows/ci.yml
mason-fish: No, but I can play around with adding it back in separately

PR #1991, opened 2021-12-12
Upgrade electron-builder and keytar

mason-fish:
Having bumped the electron version, we see this error while building windows releases:

image

So I've bumped both the `electron-builder` and `keytar` packages


package-lock.json
```diff
@@ -20302,9 +20793,9 @@
       "dev": true
     },
     "node-abi": {
-      "version": "2.19.3",
-      "resolved": "https://registry.npmjs.org/node-abi/-/node-abi-2.19.3.tgz",
-      "integrity": "sha512-9xZrlyfvKhWme2EXFKQhZRp1yNWT/uI1luYPr3sFl+H4keYY4xR+1jO7mvTTijIsHf1M+QDe9uWuKeEpLInIlg==",
+      "version": "2.30.1",
```
mason-fish, line 20796: This is what we really needed to upgrade according to the error in the PR description's screenshot

nwt (approved):

PR #1998, opened 2021-12-21
rename src/js/state/Workspaces to Lakes

nwt:
For #1922.


mason-fish (approved):
Looks like there will still be some code to change but this makes a huge dent, thanks @nwt !

PR #2002, opened 2021-12-24
Fix zapi link

philrz:
I enjoy keeping up with the project notifications, and I keep seeing a CI alert every morning because of the broken link in the CHANGELOG, so I figured I'd suggest a fix. :smile:


philrz:
That revealed more broken links, which in turn revealed other old stuff (e.g., lingering refs to `zqd` and other pre-lake stuff). I guess the docs need more attention than this quick fix I was attempting, so I hereby retreat.

jameskerr:
Thanks Phil. Those broken links have been bothering me too. I have a branch (the zed-inspector branch) that re-hauls the contributing doc and removes the outdated words.

PR #2005, opened 2021-12-27
fast inspector

### LINKED ISSUE #2007: Improve Speed on Zed Inspector
A synchronous loop in the Zed Inspector to build the rows makes it slow on large data sets (3-4 seconds). It was  slow when expanding and collapsing nested values if the number of zed values was large.

jameskerr:
This iteration of the inspector component removes the synchronous loop over all the zed values to create the rows. It instead just performs a count of how many rows there "will be". It does this by looking at the zed types and the "isExpanded" state. This was fast in my tests at 40ms for 10,000 values. It then looks at the scrollTop and height of the scroll container to determine which rows to actually create. To boost it further, it saves the result of the built rows in a cache when scrolling so that when it needs to re-render on a scroll, it only creates the rows that haven't been built in previous renders.

The old inspector was slow when expanding and collapsing nested values if the number of zed values was large. This is not a problem anymore.

This PR needs a few more things before its done:

- [x] Fix Infinite Scrolling
- [x] Account for type values
- [x] Remove the "InspectorContext" class
- [x] Fix right clicking the values
- [x] Remove the local tab state idea
- [x] Remove rootIndexValue and maybe the other added prop to InspectArgs
- [x] Correctly switch to the view when the viewer search completes

Fixes #2007


app/search/flows/next-page-viewer-search.ts
```diff
@@ -46,7 +46,7 @@ function nextPageArgs(
           .time(prevTs)
           .add(1, "ms")
           .toDate()
-        spliceIndex = index + 1
+        spliceIndex = index
```
jameskerr, line 49: I think this is right. I saw a bug where the last record was duplicated. Need to confirm.

src/js/components/SearchBar/MenuAction.tsx
```diff
@@ -2,18 +2,18 @@ import {useBrimApi} from "app/core/context"
...
 import Modal from "../../state/Modal"
-import Tab from "../../state/Tab"
 import PopMenuPointy from "../PopMenu/PopMenuPointy"
 import InputAction from "./InputAction"
 
 export default function MenuAction() {
   const dispatch = useDispatch()
   const api = useBrimApi()
-  const isFetching = useSelector(Tab.isFetching)
+  const isFetching = useSelector(Viewer.isFetching)
```
jameskerr, line 16: I decided to only care about the viewer search instead of the viewer and the chart.

src/js/flows/config.ts
```diff
@@ -1,2 +1,2 @@
-export const PER_PAGE = 100
+export const PER_PAGE = 500
```
jameskerr, line 1: This is what it used to be. Changing it back to 500

src/js/initializers/initStore.ts
```diff
@@ -6,6 +6,9 @@ import ipc from "../electron/ipc"
 import rootReducer from "../state/rootReducer"
 import {configureStore} from "@reduxjs/toolkit"
 import BrimApi from "../api"
+import {enableMapSet} from "immer"
+
+enableMapSet()
```
jameskerr, line 11: To use maps with immer.

src/js/state/getPersistable.ts
```diff
@@ -30,7 +30,7 @@ export function getWindowPersistable(state: State) {
       delete tab.viewer
       delete tab.chart
       delete tab.logDetails
-      delete tab.localState
```
jameskerr, line 33: This local state thing was not a good move. Keep state explicit.

src/js/state/createGlobalStore.ts
```diff
@@ -2,5 +2,15 @@ import {configureStore} from "@reduxjs/toolkit"
...
 
 export default function(initState: GlobalState | undefined) {
-  return configureStore({reducer: globalReducer, preloadedState: initState})
+  return configureStore({
+    reducer: globalReducer,
+    preloadedState: initState,
+    middleware: (getDefaultMiddleware) => {
+      return getDefaultMiddleware({
+        thunk: true,
+        immutableCheck: false,
```
mason-fish, line 11: These checks seem useful, can we turn them on or do we have immutability/serializable issues still?

mason-fish (approved):
What a huge optimization, nice work! 🔥

src/js/state/createGlobalStore.ts
jameskerr: I was just getting warnings when running the tests that the middleware was taking over 30ms and should maybe be disabled. So I disabled it.

PR #2030, opened 2022-01-07
Remove feature flag for zed inspector

jameskerr:
Remove the Feature Flag that held back the Inspector View.

SIDE FIXES
1. I added the three peer dependencies that our app required, so we won't see those warnings each time we install.

2. Added the [lint-staged](https://github.com/okonet/lint-staged) package that will run prettier, tsc, and eslint in a pre-commit hook. It uses [husky](https://github.com/typicode/husky) under the hook for this.

Because of this, I think we can disable the prettier check on CI. That will save us a little bit of time.


.github/workflows/ci.yml
```diff
@@ -48,4 +48,6 @@ jobs:
         if: failure() && steps.playwright.outcome == 'failure'
         with:
           name: artifacts-${{ matrix.os }}
-          path: run/playwright-itest
+          path: |
+            run/playwright-itest
+            !run/**/SS
```
jameskerr, line 53: This got our artifact uploads working again. We ignore the symlinked files that electron created.

mason-fish (approved):
Awesome! Excited to see this released into the wild. I dig the side changes too 💯

Issue #2033, opened 2022-01-07
Flaky Test: Playwright ingest.spec.ts


jameskerr:
This is a very similar failure to the query.spec.ts #2026 .

I think the problem is that we run these test blocks in a loop. If we only ingested one file, then checked the output, I'd bet we'd see more consistent things.

```js
import path from "path"
import {testDataDir} from "../helpers/env"
import TestApp from "../helpers/test-app"

describe("Ingest tests", () => {
  const app = new TestApp("Ingest tests")

  beforeAll(async () => {
    await app.init()
  })

  afterAll(async () => {
    await app.shutdown()
  })

  const searchZql =
    '_path=="conn" proto=="tcp" | cut ts, id.orig_h, id.orig_p, id.resp_h, id.resp_p, proto | sort ts'
  const sampleFiles = [
    "sample.pcap",
    "sample.pcapng",
    "sample.tsv",
    "sample.ndjson",
    "sample.zng"
  ].map((f) => [path.normalize(path.join(testDataDir(), f)), f])

  sampleFiles.forEach(([path, name]) => {
    test(`ingest of ${name}`, async () => {
      await app.ingestFiles([path])
      await app.search(searchZql)
      const results = await app.getViewerResults()
      expect(results).toMatchSnapshot(name)
    })
  })
})
```

This is what the failure looks like on CI. Passes locally.

```
2022-01-07T18:01:34.4721050Z FAIL browser: Electron (chromium) test/playwright/tests/ingest.spec.ts (61.816 s)
2022-01-07T18:01:34.4722550Z   ● Ingest tests › ingest of sample.zng
2022-01-07T18:01:34.4723240Z 
2022-01-07T18:01:34.4724160Z     expect(received).toMatchSnapshot(hint)
2022-01-07T18:01:34.4724830Z 
2022-01-07T18:01:34.4725830Z     Snapshot name: `Ingest tests ingest of sample.zng: sample.zng 1`
2022-01-07T18:01:34.4726580Z 
2022-01-07T18:01:34.4727580Z     - Snapshot  -  51
2022-01-07T18:01:34.4728030Z     + Received  + 889
2022-01-07T18:01:34.4728280Z 
2022-01-07T18:01:34.4728610Z       Array [
2022-01-07T18:01:34.4729360Z     -   "ts",
2022-01-07T18:01:34.4730110Z     -   "id › orig_h",
2022-01-07T18:01:34.4730890Z     -   "id › orig_p",
[...]

PR #2034, opened 2022-01-07
Remove Flaky Test

jameskerr:
Removes the flaky test in the playwright suite called ingest.spec.ts. The issue that documents it is found here. #2033


mason-fish (approved):
I included my flake theory for this in the new issue

PR #2036, opened 2022-01-10
remote-queries v0

mason-fish:
Here is v0 of the remote-queries feature 🎉 

A remote query is simply a named and described zed value (the same as a query from our local query lib) except that it is persisted in a zed lake instead of on a local file system. This is all facilitated by the app by creating a `_remote-queries` pool, and loading query records in for each create/update/delete operation. We include a timestamp on each query record operation so that we know which record represents the most current state. When a query is deleted, a new query record will include a `tombstone: true` value and will be the most current record. We then filter those deleted records out when retrieving them from the lake.

NOTE: There are several aspects of this feature which we intend to improve but are constrained by the current pool-centric UI. Very soon we will break ground on the new query-centric UI, which should greatly enable these remote queries to be a seamless workflow for third party integrations such as Observable. The primary pieces we are missing, to be added after the UI re-design, are:

* Updating current local/remote query within search page/on search
* Remote query tags (for now we are using the tag view as the means to access the remote queries and would need another ui element/dropdown if we wanted to show a remote query tag filter)
* Remote Query Folders

Select to see remote queries through the queries dropdown in upper right of section header
image

Create new remote query with "+" button in header after selecting the remote view from first pic
image

Fill out query, as usual (note there are no tags for remote queries yet)
image

Run the query just like a local query
image

### DIFF
--- a/src/js/components/LeftPane/QueriesSection.tsx
+++ b/src/js/components/LeftPane/QueriesSection.tsx
@@ -1,6 +1,6 @@
 import {nanoid} from "@reduxjs/toolkit"
 import {useBrimApi} from "app/core/context"
-import {includes} from "lodash"
+import {find, includes} from "lodash"
 import React, {ChangeEvent, useEffect, useRef, useState} from "react"
 import {Tree, TreeApi} from "react-arborist"
 import {useDrop} from "react-dnd"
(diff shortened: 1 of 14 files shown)


jameskerr (approved):
This works for me! Thanks @mason-fish !

PR #2043, opened 2022-01-11
Fix about window commit

### LINKED ISSUE #2037: "About" window no longer shows commit hash in Dev mode
Long ago I became accustomed to cutting & pasting the commit hash in the "About" window when filing bugs against a Brim app running in Dev mode. However, when trying this with the most recent Brim, I noticed that it's now showing just the `0.28.0` that corresponds with the last GA release, as shown in the screenshot below. I did a binary search and confirmed that the commit hashes stopped being shown as of Brim commit f0429b6 which is associated with #1985 (cc: @mason-fish).

image

mason-fish:
fixes #2037 

When we migrated to the new "@electron/remote" module, I had tried to find a way to access it the way we had been in `/app/core/env.ts` and ` src/js/electron/isDev.ts ` but could not puzzle out at the time why that caused our playwright tests to fail. Looking at it again this time, I can now see that the playwright test setup is the source, importing the `env.ts` package *before* it has launched electron, and that js test environment is not suitable to import the electron remote module.

The fix is to conditionally import the remote module based on the env method called, and the playwright setup does not call those methods. I think this is fine for now since the playwright launch setup is the only offender and is unlikely to change much. Later, and if needed, we can extend the conditional import to only happen after first inspecting a new env var to indicate what js environment we are in.

I also deleted the old about modal.

image


jameskerr (approved):
Cool with me!

PR #2048, opened 2022-01-13
Npm Workspaces

### LINKED ISSUE #2047: Use Npm Workspaces
- [x] Require everyone to install npm version 8
- [ ] Move zealot into the workspaces

jameskerr:
Closes #2047


jameskerr:
I was hoping to use the npm workspaces feature to have both the brim app the zealot client as two packages in the same repo. That requires us to upgrade to npm 7. Turns out npm 7 has a show-stopping bug preventing us from being able to use it. When it downloads a git dependency (for us that’s zed and brimcap ), it generates an incorrect integrity sha. This means each time you npm install zed, you’ll get a different sha for those deps in your package-lock.json. It also means that npm install  on ci will always fail with an EINTEGRITY errors. This didn’t happen in npm version 6. The sha was the same every time. There are issues in npm’s issue tracker for this. They plan to drop integrity checks on git dependencies.

But this makes using npm workspaces, which came out in npm 7, impossible without some very hacky workarounds that need to run on every npm install (local or ci). The alternatives are to switch our package manager to yarn, or pnpm or use lerna and stay on npm 6. I’m looking into what to do.

PR #2049, opened 2022-01-13
New Sidebar (feature flag: 'query-flow')

mason-fish:
New sidebar, ready for review! This is the first step toward what I think of as the query based ux/flow that we will be building for the app. The work here focuses only on the sidebar and will be hidden behind the 'query-flow' feature flag until the search bar and pool detail page have been implemented.

There is a lot of additive work here, and much of it is net new (with only a little bit of duplication from the old sidebar). Once we remove the ff then there will be a lot of old code to remove, but this collection of components are all self contained and will not need to be touched for that removal.

I ended up doing a lot of refactoring, and there's still more to do (would still love to use the "arborist" tree for the history's tree view), but I'm pretty happy with the overall improvement and the new ux design is really gonna take the app to the next level imo.

More than the reskin and refactoring, we also get a couple new features here that made sense to implement at the same time. Namely, search filters for every section, and inline renaming for all section items where applicable.


src/js/components/LeftSidebar/HistorySection/HistoryLinnear.tsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <SectionContents ref={resizeRef}>
+      <Tree {...defaults} data={{id: "root", items: entries}}>
+        {HistoryItem}
+      </Tree>
+    </SectionContents>
+  )
+}
+
+export default HistoryLinear
```
jameskerr, line 135: The name of this file has a type.

src/js/components/LeftSidebar/PoolsSection/pool-item.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ) : (
+          <Name>{pool.name}</Name>
+        )}
+        {progress}
+      </StyledPoolItem>
+    </ItemBG>
+  )
+}
+
+export default PoolItem
```
jameskerr, line 99: It would be great if the names of these new folders followed our desired convention of kabab-case everything. I also think all the new code should be in `app/features/sidebar/`.

src/js/state/Appearance/index.ts
```diff
@@ -1,17 +1,24 @@
 import {createSlice, PayloadAction} from "@reduxjs/toolkit"
 import {SectionData} from "src/pkg/sectional"
 import {State} from "../types"
+import {HistoryView, QueriesView, SectionName} from "./types"
 
 const init = () => ({
   sidebarIsOpen: true,
   sidebarWidth: 230,
-  sidebarSections: [{id: "pools"}, {id: "queries"}, {id: "history"}]
+  sidebarSections: [{id: "pools"}, {id: "queries"}, {id: "history"}],
```
jameskerr, line 9: This was a bad idea on my part. I think this should just be hard coded in the jsx.

src/js/state/Appearance/index.ts
```diff
@@ -1,17 +1,24 @@
...
+  currentSectionName: "pools" as SectionName,
+  queriesView: "local" as QueriesView,
+  historyView: "linear" as HistoryView
```
jameskerr, line 12: Nice!

jameskerr (approved):
I dig it! It's looking feeling really good.  Let's address some of these, but then merge it and keep iterating.

Here are my notes:

- [ ] Improve side bar item hover and active state, I'll give you some designs, but basically, the whole line should not change background, just the part with the rounded edges. 
image

- [ ] The linear history view has a flex issue 
image
- [ ] In the image above, make the search bar width fill up the space 100% (minus the padding)
- [ ] Add a hover state to the lake title.
- [ ] Decrease the size of the carets 
image
- [ ] Change the stroke and fill of the pools icon white when it's selected (right now it had some black in there) 
image
- [x] Add icon to the search bar
- [x] Have the item's overflow a little sooner (I'll design this as well)
- [x] Is the color of the history icon the same as the others?

jameskerr:
Would it make sense to have a single component called "SidebarItem" that could be used in the tab headers, the pool items, and the query items?

src/js/state/Appearance/index.ts
mason-fish: Yeah the new sidebar doesn't use those, I'll remove when we remove the ff

mason-fish:
I explored that <SidebarItem /> a bit but there ended up being a bit too much conditional variety to manage in one abstraction between each of the section items' contents. I opted for a single shared styled component to at least unify their appearance in the end.

PR #2055, opened 2022-01-14
Switch to pnpm

jameskerr:
https://pnpm.io/

This looks like it will be faster and supports workspaces. I looked into yarn version 2 but they are dead set on this no install plug and play system that differs significantly from the normal node_modules structure we're used to. You can opt out, but it seems like we'd be a second-class citizen. Also, electron doesn't work with "plug n play". (its pretty cool though).

Pnpm seems like the easiest drop in replacement for npm. And it's the fastest of all package managers.

And, it supports workspaces.


.github/workflows/ci.yml
```diff
@@ -18,18 +18,21 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: "1.16"
+      - uses: pnpm/action-setup@v2.0.1
```
nwt, line 21: Better to pin just the major version so we get any bug fixes or performance improvements that appear in a minor or patch release.
```suggestion
      - uses: pnpm/action-setup@v2
```

jameskerr: Ok sounds good.

jameskerr:
pnpm does not work with electron

https://github.com/pnpm/pnpm/issues/3108

yarn it is

PR #2118, opened 2022-02-01
Query flow

mason-fish:
This is still hidden behind the "query-flow" feature flag (play with it by running `yarn start --feature-flags=query-flow`)

There are still several places where the app assumes you've already selected a pool somewhere, and I'd like to continue to think through and iterate on the `BrimQuery` abstraction and the api it exposes. But I wanted to check this in so I can work on this in more focused chunks now that we have a really solid beach head established. Here are some screenshots to get a feel for the changes:

image

image

image

image

image

This is just a placeholder since there was no design yet but I needed to have something to connect the ingest -> begin searching experience (so the routing to this page and that query button work, but the presentation needs some tlc)
image


jameskerr:
Looking great @mason-fish !!

PR #2124, opened 2022-02-03
Zealot Upgrade

jameskerr:
* Move zealot into its own yarn workspace.
* Remove zealot-old and use packages/zealot in all places in the app.
* Upgrade the decoder/encoder for zjson
* Remove the api tests and the response tests as they were coupled to the old zealot implementation

Todos:
- [x] Add a timeout to the client requests
- [x] Write the migration to remove the pools state and ignore the ingest state


packages/zealot/src/client/client.ts
```diff
@@ -0,0 +1,229 @@
...
+  CreatePoolOpts,
+  CreatePoolResp,
+  CrossFetch,
+  LoadOpts,
+  Pool,
+  QueryOpts,
+  ResponseFormat
+} from "./types"
+
+export class Client {
```
jameskerr, line 19: This is the new zealot client class. It's flat and contains duplication at the moment. I wanted to get it all out here like this first, then clean up the duplication.

packages/zealot/src/query/result-stream.ts
```diff
@@ -0,0 +1,118 @@
+import {isObject} from "lodash"
+import {Response as NodeResponse} from "node-fetch"
+import {zjson} from ".."
+import {eachLine} from "../ndjson/lines"
+import {JSOptions} from "../zed/values/types"
+import {Channel, Collector} from "./channel"
+
+type CrossResponse = Response | NodeResponse
+export class ResultStream {
```
jameskerr, line 9: This is the class that handles the stream of zed values coming back from the server when you issue a query. 

```
const res = await client.query("hello")
await res.js()
await res.collect(({rows, shapes}) => {
  setRows(rows)
})
```
The above example is for a single channel result.

If you have multiple channels, all the methods are the same, you just get the channel first.

```
await res.channel(3).collect(() => {})
await res.channel(2).js()
await res.channel(5).shapes
/// and so on
```

The `.promise` method returns a promise for when the whole thing is done.

mccanne:
Should we archive the old repo and add a note pointing to brim-zealot once this is merged?

app/features/import/import-files.ts
```diff
@@ -144,6 +138,15 @@ const executeLoader = (
         onDetailUpdate,
         abortCtl.signal
       )
+      // Wait for the pool to have some data before we signal that
+      // the ingest is done.
```
mason-fish, line 142: Can we not rely on there being data after `await l.load()` has completed? If not, then we should modify it so it does so consumers of it don't need to worry about this sort of retrying imo

app/features/sidebar/flows/remote-queries.ts
```diff
@@ -1,30 +1,18 @@
...
-  const value = r.get<zed.String>("value").toString()
-
-  return {
-    description,
-    id,
-    name,
-    tags: [],
-    value
-  }
+const recordToQuery = (r: Query): Query => {
```
mason-fish, line 14: I ended up just removing this in favor of the `.js()` 🔥

app/features/sidebar/flows/remote-queries.ts
```diff
@@ -77,27 +63,28 @@ export const setRemoteQueries = (
...
 
     try {
       const data = new Readable()
-      const loadPromise = zealot.pools.load(rqPoolId, "main", {
-        author: "brim",
-        body:
-          "automatic remote query load for id(s): " +
-          queries.map((q) => q.id).join(", "),
-        data
+      const loadPromise = zealot.load(data, {
```
mason-fish, line 79: I like the api change here 👏

app/features/sidebar/pools-section/pool-item.tsx
```diff
@@ -1,10 +1,9 @@
 import getPoolContextMenu from "app/pools/flows/get-pool-context-menu"
-import {lakeSearchPath} from "app/router/utils/paths"
+import {poolSearchPath} from "app/router/utils/paths"
```
mason-fish, line 2: shoot, I changed these paths in my query-flow work too, though we are using the same name for this one 👌

app/features/sidebar/pools-section/pool-item.tsx
```diff
@@ -45,38 +47,40 @@ const StyledPoolItem = styled(StyledItem)<{isSelected: boolean}>`
   }
 `
 
-const PoolItem = ({innerRef, styles, data: pool, state, handlers}) => {
+// Is this duplicated somewhere?
```
mason-fish, line 50: Shouldn't be duplicated (at least not even close to verbatim): this is the new PoolItem for the Pools section tab which is different both in style and code to the current pool list section (which doesn't use the arborist component)

app/features/sidebar/pools-section/pool-item.tsx
```diff
@@ -45,38 +47,40 @@ const StyledPoolItem = styled(StyledItem)<{isSelected: boolean}>`
...
+        spanArgs: pool.empty() ? undefined : pool.defaultSpanArgs()
       })
     )
   }
 
-  const progress = p.ingesting() && (
+  const progress = ingest && isNumber(ingest.progress) && (
     <div className="small-progress-bar">
-      <ProgressIndicator percent={p.ingestProgress()} />
+      <ProgressIndicator percent={ingest.progress} />
```
mason-fish, line 80: Got it, so ingest is no longer stored in the pool. That certainly makes updating that particular slice in redux simpler 👌

app/lakes/new.tsx
```diff
@@ -32,12 +36,18 @@ const SubmitWrap = styled.div`
...
 
   const onSubmit = (e) => {
-    dispatch(createPool({name})).catch((e) => {
-      dispatch(Notice.set(ErrorFactory.create(e)))
-    })
+    const lakeId = select(Current.getWorkspaceId)
     e.preventDefault()
+    dispatch(createPool({name}))
+      .then((id) => {
+        dispatch(tabHistory.push(lakePath(id, lakeId)))
```
mason-fish, line 46: Glad you moved this out of the former `createPool` flow 👍

app/legacy/space-migration/space-migration.tsx
```diff
@@ -83,7 +83,7 @@ function Modal({onClose}) {
 
     spaces
       .migrate(({total, count, space}) => {
-        dispatch(refreshPoolNames())
+        dispatch(syncPools())
```
mason-fish, line 86: Coming back to my naming comment, maybe `syncPoolNames` is my new favorite name suggestion

package.json
```diff
@@ -1,28 +1,32 @@
 {
-  "name": "Brim",
+  "name": "brim",
```
mason-fish, line 2: Just a mental note that we should double check this doesn't change the name of the app when we build a release version of it

packages/zealot/src/client/client.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    Array [
+      Object {
+        "name": "alice",
+      },
+    ]
+  `)
+})
+
+test("collect values", async () => {
+  // const resp = await client.query("from 'my-pool' | alice", {})
```
mason-fish, line 35: straggler

packages/zealot/src/client/client.ts
```diff
@@ -0,0 +1,229 @@
...
+    })
+
+    const resp = await this.fetch(this.baseURL + "/query", {
+      method: "POST",
+      body: JSON.stringify({query}),
+      // consolodate the creating of these fetch arguments
+      headers: {
+        Accept: getAcceptValue(options.format),
+        "Content-Type": "application/json",
+        Authorization: `Bearer ${this.auth}`
```
mason-fish, line 57: This is the only place I'm seeing us use auth and it looks like if you don't provide auth while instantiating the client then it will still set an Authorization header to `Bearer undefined`. Better to not set the header at all imo, and we def will want to include auth on most of the other endpoints (not needed for version, and authMethod)

packages/zealot/src/client/client.ts
```diff
@@ -0,0 +1,229 @@
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: getAcceptValue("zjson")
+      },
+      body: JSON.stringify({
+        name,
+        layout: {
+          order: options.order,
+          // @ts-ignore What's with all the array wrapping?
```
mason-fish, line 96: I have the same question :)

packages/zealot/src/query/result-stream.ts
mason-fish: This looks way cleaner to me now. Though I still wish the channel system could use strings instead of numbers to name and refer to each channel. Not sure how much we are in control of that though with the behavior of backend

packages/zealot/src/util/utils.ts
```diff
@@ -7,8 +7,8 @@ function hasOwnProperty<X extends {}, Y extends PropertyKey>(
   return Object.prototype.hasOwnProperty.call(obj, prop)
 }
 
-export function url(baseUrl: string, path: string) {
-  return `${baseUrl}${path}`
+export function url(host: string, path: string) {
+  return `http://${host}${path}`
```
mason-fish, line 11: I think we used to have it like this but needed to remove any hard-coded protocols ("http" in this case) so we could reach out to https-handled backends, no?

mason-fish (commented):
This looks great! I left mostly style-related nits and questions. I think I'm going to have a nightmare rebase to deal with after this lol, but in a lot of cases I think we had the same idea for changes so will see. I think my only major concerns are with regard to authentication (which looks like it's a wip but almost complete), the hard-coded "http" business, and just making sure these app + zealot workspaces all get packaged up properly in the release build. Had you given that a smoke test yet? 🤞

jameskerr:
Thanks @mason-fish. Yes, the e2e tests are running off the released build. The new workspace structure is working for those. Will address your other comments inline.

app/core/pools/pool.ts
```diff
@@ -0,0 +1,69 @@
...
+  hasStats() {
+    return !!this.stats
+  }
+
+  hasSpan() {
+    return this.hasStats() && this.stats.span !== null
+  }
+
+  hasTsKey() {
+    return isEqual(this.data.layout.keys, [["ts"]])
```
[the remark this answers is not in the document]
jameskerr: I think this is what we want to test for. Does the pool have the field ["ts"] as the pool key. If that turns out to be wrong down the line, we can update this.

app/core/pools/sync-pools.ts
```diff
@@ -0,0 +1,22 @@
+import {getZealot} from "src/js/flows/getZealot"
+import Current from "src/js/state/Current"
+import Lakes from "src/js/state/Lakes"
+import Pools from "src/js/state/Pools"
+import {Thunk} from "src/js/state/types"
+
+/**
+ * This only gets the list of names from the server.
+ */
+export default function syncPools(lakeId?: string): Thunk<Promise<void>> {
```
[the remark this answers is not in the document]
jameskerr: I'd rather not call all the poolStats endpoints in a loop in case there are hundreds of pools in someone's lake. I like the rename option.

app/features/import/import-files.ts
jameskerr: We unfortunately can't! I can open an issue on the backend. I spent a half a day debugging this. The load can finish and return, but then if you get the pool stats, the size can be 0 and the span can be null.

app/features/sidebar/pools-section/pool-item.tsx
jameskerr: Exactly! Although I need to rename the state slice to "Ingests". The reason is because you can name a variable "import". It's a keyword!

```
const import = Imports.get(id) // wont work!
```

package.json
jameskerr: Yes, I have that same mental note.

packages/zealot/src/client/client.ts
jameskerr: Yes, true. I'll need to do an auth pass on the client before merging this.

packages/zealot/src/client/client.ts
jameskerr: Haha, the backed uses arrays to designate field names. So [["ts"]] means the pool keys are a single field name "ts". If the key was a nested field, it would be `keys: [["id", "orig_h"]]`. If there were multiple keys it would be `keys: [["id"], ["ts"], ["my", "nested", "field"]]`.

packages/zealot/src/util/utils.ts
jameskerr: Oh yeah, thanks. This function is no longer used. Removed.

mason-fish (approved):

Issue #2156, opened 2022-02-11
Lake doesn't start


jameskerr:
When I install the pre-release, packed app and start it up. I get this.

<img width="1362" alt="Pasted Graphic" src="https://user-images.githubusercontent.com/3460638/153519606-7431fe95-c4ad-4cdc-a428-b3db29df3a1a.png">

The logs show that spawn was called. We need to add more logs to see what's happening.

```
[2022-02-10 16:23:45.179] [info] spawning zed serve: /Applications/Brim.app/Contents/Resources/app.asar.unpacked/zdeps/zed serve -l localhost:9867 -lake /Users/jkerr/Library/Application Support/brim/data/lake -log.level=info -log.filemode=rotate -log.path /Users/jkerr/Library/Application Support/brim/logs/zlake.log -brimfd=3
```

This is logged, but it actually start the lake. If I run that command myself, it does start the lake.

PR #2158, opened 2022-02-12
New Zed Root Directory

jameskerr:
1. Use the zed lake class in the zealot folder.
2. Use the zdeps binary for zed
3. Change the root directory for the lake to "userdata/lake".

This allows the app to start up without error.

Fixes: #2156


package.json
```diff
@@ -88,11 +88,11 @@
...
     "husky": "^7.0.4",
-    "jest": "^27.4.5",
-    "jest-circus": "^27.4.6",
-    "jest-environment-node": "^27.4.6",
+    "jest": "^28.0.0-alpha.0",
+    "jest-circus": "^28.0.0-alpha.0",
+    "jest-environment-node": "^28.0.0-alpha.0",
     "jest-playwright-preset": "^1.7.0",
-    "jest-runner": "^27.4.6",
+    "jest-runner": "^28.0.0-alpha.0",
```
jameskerr, line 95: I had to upgrade jest to this new version to import "@brimdata/zealot/node" in the tests. The package.json "exports" field is new and just getting support.

packages/zealot/package.json
```diff
@@ -12,10 +12,26 @@
...
+  "typesVersions": {
+    "*": {
+      "*": [
+        "dist/cjs/index.d.ts"
+      ],
+      "node": [
+        "./dist/cjs/node.d.ts"
+      ]
+    }
+  },
```
jameskerr, line 34: A hack to have typescript support the subfolder exports.

src/js/electron/main.ts
```diff
@@ -29,7 +29,7 @@ require("@electron/remote/main").initialize()
 
 export const mainDefaults = () => ({
   lakePort: 9867,
-  lakeRoot: join(app.getPath("userData"), "data", "lake"),
+  lakeRoot: join(app.getPath("userData"), "lake"),
```
jameskerr, line 32: This is the whole point of this PR

mason-fish (approved):
This looks good to me pending a green build.

One question though: if we are effectively dropping all existing lakes and starting a new one with this change, should we also include a migration to reset the rest of the app state?

packages/zealot/src/cmd/paths.ts
```diff
@@ -3,6 +3,13 @@ import os from "os"
...
+  let dir = "../../../../node_modules/zed/dist"
+  if (isBuilt()) dir = "../" + dir
+  return path.join(__dirname, dir, name)
+}
 
-  return path.join(__dirname, "../../../../node_modules/zed/dist", name)
+function isBuilt() {
+  const buildDir = path.normalize("dist/cjs/cmd")
+  const regex = new RegExp(`${buildDir}$`)
+  return regex.test(__dirname)
```
nwt, line 14:
```suggestion
  const buildDir = path.normalize("dist/cjs/cmd")
  return __dirname.endsWith(buildDir)
```

jameskerr:
@mason-fish good point. The app state will remain unchanged, but the data for the zed lake will be new. In the release notes we will provide a migration script for people to run that will bring the data from the old lake into the new one.

test/playwright/playwright.config.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
+  globalSetup: "./setup/global.ts",
+  testDir: "./tests",
   browsers: [
```
mason-fish, line 4: I don't think this browsers key is a playwright config options and can now be removed. It was only used with the jest setup to rename the test browser in each test's output (it's default output did not indicate electron at all despite the fact that it was driving our electron app)

mason-fish (commented):
@jameskerr The new test runner looks good! Part of why I didn't use it originally was that it wasn't resolving zealot properly, so this actually makes perfect sense to do now. And with this in place we can also begin to use their very nice [trace-on-failure](https://playwright.dev/docs/test-configuration#record-test-trace) features for debugging flakiness 👌

test/playwright/playwright.config.js
jameskerr: Awesome!

PR #2177, opened 2022-02-15
Always show queries

### LINKED ISSUE #1954: load queries when no pool selected
If no pool is selected, all of the capabilities of the query library panel are disabled.

It seems like you should be able to load your query library (e.g., from a JSON file) even if no pool is selected.  For that matter, you should be able to create/edit/export queries also when a pool is not selected.

jameskerr:
Show the query library even if no pool is selected. If you try to run a query you get a simple error saying no pool is selected. Closes #1954 

Screen Shot 2022-02-15 at 2 36 42 PM

Screen Shot 2022-02-15 at 2 36 49 PM


mason-fish (approved):
This is the direction we're headed in anyway, so :shipit:

PR #2193, opened 2022-02-19
Inspector Error Types

jameskerr:
Places to check for proper rendering of any value.

**Table View**
* Fluid Table

image

* Columnar Table

image

**Inspector View**
* When it is one line line in a container (record/array)

image

* When it's container is opened and it render's on it's own line

image

* If it's a container, when it's expanded and collapsed

image

This goes for the typeof's as well

image

image

image

If you try to render anything other than a value in the table view, you'll now get this message.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,10 @@ jobs:
               yarn electron-builder --linux=deb
               sudo apt install -y --no-install-recommends ./dist/installers/*.deb
               ;;
+            macOS )
+              yarn electron-builder --mac=dir
+              cp -r dist/installers/mac/Brim.app /Applications
+              ;;
           esac
       - name: End to end tests
         id: playwright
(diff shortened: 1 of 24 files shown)


.github/workflows/ci.yml
```diff
@@ -36,14 +36,10 @@ jobs:
               yarn electron-builder --linux=deb
               sudo apt install -y --no-install-recommends ./dist/installers/*.deb
               ;;
-            macOS )
-              yarn electron-builder --mac=dir
-              cp -r dist/installers/mac/Brim.app /Applications
-              ;;
```
nwt, line 42: Maybe leave this for the future and just change `if: runner.os != 'Windows'` to `if: runner.os == 'Linux'?

nwt (approved):

.github/workflows/ci.yml
jameskerr: Ok sounds good.

PR #2208, opened 2022-02-22
Remove Babel SVG Loader

### LINKED ISSUE #2206: Remove svg loader from babel
We may want to switch from babel to something much faster like esbuild. Let's start preparing for this by removing the special features that are exclusive to babel.

jameskerr:
This proved a minor convenience for the cost of not being able to move on from babel when the time comes.

Closes #2206


mason-fish (approved):

jameskerr:
Subsumed by #2211

PR #2211, opened 2022-02-24
Remove Babel

### LINKED ISSUE #2004: Look into esbuild to replace babel
It mike be an easy drop in replacement to increase our build time 10x

jameskerr:
This PR cleans up our repo by:

1. Moving all source code that needs to be compiled into `src`
2. Moving the playwright tests into `packages/e2e-tests`
3. Replacing babel with swc.

swc is about 100 times faster than babel. It is designed to be a drop in replacement, and it was for the most part. I chose it for that reason as apposed to esbuild which is designed to replace a bundler like webpack. We can use esbuild later to build certain packages.

Moving all the source code into the single folder `src` makes all our configuration files so much simpler.

Closes #2004


mason-fish (approved):
Woo! This is awesome, `swc` ftw 🔥

PR #2228, opened 2022-03-01
Use tagged zed version in docs urls

### LINKED ISSUE #2209: Fix syntax docs links
There are some links in the app that allow users to view the zed syntax docs. 

Right now they point to the main, but they need to point to the zed version that the app is bundled with. The main branch might change in a way that won't work with the current version of the app.

mason-fish:
fixes #2209 

This also updates some docs paths which had changed. If those paths change again we will continue to need to make those changes explicitly. But now the docs will correspond to the appropriate version of zed.

NOTE: It may be simpler to wait until the zed docs have settled more before merging this. I'll leave the PR up since that may happen very soon and I can tie up the handful of remaining broken links at that time.

### DIFF
--- a/docs/Joining-Data.md
+++ b/docs/Joining-Data.md
@@ -39,18 +39,15 @@ ultimately be combined. The Zed [`join` docs](https://github.com/brimdata/zed/tr
 show examples with the [Zed CLI tools](https://github.com/brimdata/zed#quick-start)
 that specify these inputs as named files or pools in a [Zed Lake](https://github.com/brimdata/zed/blob/main/docs/zed/README.md).
 
-Brim release `v0.25.0` introduced initial support for storing data in Zed Lakes.
-However, due to a current limitation ([brim/1618](https://github.com/brimdata/brim/issues/1618)),
-Zed queries issued from within Brim cannot yet access multiple pools
-simultaneously. Instead such queries are currently limited to accessing the
-data from whichever pool is currently selected from the **Pools** list.
+The `join` operator is still experimental and has somewhat hard-to-use syntax,
+though this should be improved soon in a subsequent release.
+The easiest way to join data within Brim is to join two subsets of data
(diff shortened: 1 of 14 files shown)


src/app/core/env.ts
```diff
@@ -1,4 +1,12 @@
 import {app as electronApp} from "electron"
+import * as brimPackage from "../../../package.json"
+
+export const toTaggedZedRepoUrl = (path: string, isDirectory = false) => {
```
mason-fish, line 4: Wasn't quite sure where the best logical home was for this util function; open to suggestions if you disagree with core/env

jameskerr: How about a new file called `app/core/links` ?

jameskerr (commented):
Right on! See the comment above.

docs/Joining-Data.md
```diff
@@ -35,16 +35,17 @@ shy!
...
-that specify these inputs as named files or pools in a [Zed Lake](https://github.com/brimdata/zed/blob/main/docs/lake/README.md).
+that specify these inputs as named files or pools in a [Zed Lake](https://github.com/brimdata/zed/blob/main/docs/zed/README.md).
 
 Brim release `v0.25.0` introduced initial support for storing data in Zed Lakes.
 However, due to a current limitation ([brim/1618](https://github.com/brimdata/brim/issues/1618)),
 Zed queries issued from within Brim cannot yet access multiple pools
 simultaneously. Instead such queries are currently limited to accessing the
 data from whichever pool is currently selected from the **Pools** list.
 
+// TODO
```
mason-fish, line 48: @mccanne this needs re-write

jameskerr (approved):
Looks good, however, we might want to re-think how to do tests that hit the network. Now we need an internet connection for the tests to pass. Probably fine for most of the time, but we'll keep an eye on it.

PR #2243, opened 2022-03-03
Inspector Update

### LINKED ISSUE #2194: Expanding Complex Types in the Inspector
When you expand a complex type in the inspector, all the other types with that same reference also get expanded. This is because we store the expanded state in a map where the key is the zed value and the value is a boolean. The cool thing about zed is that there is only 1 reference to each unique type no matter how many instances of it there are. But it bites us in this case.

We need to change the key for this to be unique for each type. Not sure how this will work yet. Maybe wrap the value in something?

Here's a video

https://user-images.githubusercontent.com/3460638/154776843-15f942a8-6f70-4d25-a7c5-7d41ca01be41.mov

jameskerr:
Fixes #2194 

The bug described above was due to the way we store if a given container value is expanded or not. We used to the zed data instance itself. This worked fine most of the time. However, a feature of zed is each unique zed type is a singleton across all values that it has. This lets you compare the types of things using strict equality `val1.type === val2.type`. Given this feature, when you expanded one type value, all the other rows with that type also expanded.

To fix, I changed form storing the instance to storing an "indexPath" to that object. The example below shows how index paths are created.

```
The inspector 

Row 0 {...}
Row 1 {...}
Row 2 {name: "james", interests: ["hiking", "swimming", "coding"]}
```

The index path of the "interests" array is `[2,1]`. The first number is the row index, the second number is the field index for its record container. The benefits of this are that each value can have a unique index path, and the index paths are created on demand as you expand containers, as opposed to a solution where we'd need to walk all the nested structures.

**Other improvements:**
* A better render preview of complex types. Before a record with an array would look like `{my_array: Array(5)}`, and now it looks like `{my_array: ["one", "two" ...+3 ]}`
* Much better code structure. When I needed to add a new data type before, I had to change at least three different files each with their own switch statement that checked the type. Now each data type has it's own class that defines how it's rendered. There is now a single switch statement that checks the value and creates the view for it. For example, this is the StringView.

```ts
export class StringView extends View<zed.String> {
  render() {
    return `"${this.args.value.toString()}"` // it adds quotes to the value
  }
}
```

Whereas this is the ArrayView.

```ts
export class ArrayView extends ContainerView<zed.Array> {
  name() {
    return "Array"
  }

  openToken() {
    return "["
  }

  closeToken() {
    return "]"
  }

  render() {
    const n = 2
    const l = this.value.items.length
    const trail = l > n ? l - n : null
    const nodes = []
    nodes.push(syntax(this.openToken()))
    nodes.push(Array.from(this.iterate(n)).map(field))
    if (trail) nodes.push(note(" …+" + trail + " "))
    nodes.push(syntax(this.closeToken()))
    return nodes
  }

  *iterate(n?: number) {
    const items = this.value.items
    const length = n ? Math.min(n, items.length) : items.length

    for (let i = 0; i < items.length; ++i) {
      const last = i === length - 1
      yield createView({
        ...this.args,
        value: items[i],
        last,
        key: this.isExpanded() ? i.toString() : null,
        indexPath: [...this.args.indexPath, i]
      })
      if (last) break
    }
  }
}

```

The base classes View and ContainerView handle all the logic that is the same among all data values, while the derived types only define what is unique to themselves. 

We now have a cookie cutter for any new types that come along.


mason-fish (approved):
Right on 👌

PR #2247, opened 2022-03-05
Update Tab Behavior

jameskerr:
**Todo:**
 - [x] Ensure that this doesn't break the current behavior without a feature flag

**New Behavior:**

1. Clicking the new tab button or cmd+t now calls the `newTab()` thunk. If there is not lake, go to the lakesPath, if there is but no pools, the import path, if there are some pools the new query page.

2. Clicking on a item in the sidebar now calls `Tabs.activateByUrl` finds a tab with the same location.pathname and activates it. If there is no tab, it makes a new one.

3. The submitSearch function does not run when you open a new draft query, or when you click on a query in the sidebar.

**Things to think through:**

1. When to automatically run a query, and when to not.
2. The role of a TabsHistory now


src/app/core/state/index.ts
```diff
@@ -0,0 +1,6 @@
+import {useDispatch as useReduxDispatch} from "react-redux"
+import {AppDispatch} from "src/js/state/types"
+
+export function useDispatch() {
+  return useReduxDispatch<AppDispatch>()
+}
```
jameskerr, line 6: We should use this for all `useDispatch`s. It adds return types to the Thunks.

src/app/features/sidebar/header.tsx
```diff
@@ -0,0 +1,21 @@
...
+  justify-content: space-between;
+  align-items: center;
+`
+
+export default function Header() {
+  return (
+    <Container>
+      <LakePicker />
+      <PlusButton />
+    </Container>
```
jameskerr, line 19: Moved some code around into new files.

src/app/features/sidebar/plus-button.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const onClick = () => {
+    const template: MenuItemConstructorOptions[] = [
+      {
+        label: "New Query",
+        click: () => {
+          const query = dispatch(newQuery())
+          dispatch(Tabs.new(lakeQueryPath(query.id, lakeId, {isDraft: true})))
+        }
+      },
```
jameskerr, line 55: I removed the "New Remote Query" because I don't think we'll eventually want to have that distinction here. But I can add it back if we need it for demos.

mason-fish (approved):
just left one comment about the add remote query option, but lgtm!

src/app/features/sidebar/plus-button.tsx
jameskerr: Yeah for sure, I'll add it back.

PR #2260, opened 2022-03-08
fix pool id source for log details

### LINKED ISSUE #2238: Fix record detail views
In the query flow, the detail window still has some assumptions about a current pool id being set. Also the hotkey for the right pane is not working (may be related).

mason-fish:
fixes #2238 

Fixes a few places where we were not retrieving the pool id from the correct source (need to use underlying query's from pin instead of looking for a pool id in the location path).

Since we are supporting two flows atm, the options for solving this involved either conditionally rendering a copied and slightly adjusted component tree, or conditionally fetching the pool id in the relevant thunks. I opted for the second in hopes that when we remove the search-flow we will easily be able to find all uses of the `featureIsEnabled("query-flow")` and remove the excess logic at that time.

I had also experimented with putting this feature flag based logic inside the pool related current selectors (`Current.getPoolId`, `Current.getPool`, `Current.mustGetPool`) but there were two problems I hit there: one, we DO actually use the location based pool "getting" for the new pool home page, and two in the case of page initializers the feature flag might not yet be turned on.


jameskerr:
@mason-fish I agree with your decision to sprinkle in some `isFeaturedEnabled` checks rather than coping over all of the detail window. However, I think there is a middle ground between those two solutions. We should make it easy on ourselves to delete the legacy code when the feature flag is removed. For example, in the md5Correlation function, you could add some duplication at the value of being able to easily delete the old code when the time comes.

```js
function md5Search() {
  if (!featureIsEnabled("query-flow") return legacyMd5Search()
  ...  
}

function legacyMd5Search() {
  // the old code copied
}
```

This will help us make less mistakes when we need to remove the feature flag.

jameskerr (commented):
Behavior looks good, just want to discuss the comment above.

jameskerr (approved):

PR #2261, opened 2022-03-08
fix export for query-flow

### LINKED ISSUE #2239: Fix export
query flow export is not working, likely is related to from pin usage

mason-fish:
fixes #2239 

Get poolId from query source instead of location during query-flow

### DIFF
--- a/src/js/flows/exportResults.ts
+++ b/src/js/flows/exportResults.ts
@@ -6,11 +6,11 @@ import brim from "../brim"
 import Columns from "../state/Columns"
 import Current from "../state/Current"
 import SearchBar from "../state/SearchBar"
-import SystemTest from "../state/SystemTest"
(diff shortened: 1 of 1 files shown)


src/js/flows/exportResults.ts
```diff
@@ -37,17 +38,26 @@ export default (
...
+    if (span) {
+      const dates = span.map(brim.time).map((t) => t.toDate())
+      from = dates[0]
+      to = dates[1]
+    }
+  }
 
-  dispatch(SystemTest.hook("export-start"))
   const query = annotateQuery(program, {from, to, poolId})
+  dispatch(SystemTest.hook("export-start"))
```
jameskerr, line 60: We can delete all these by the way.

jameskerr (approved):

PR #2262, opened 2022-03-08
Named Type Fix

### LINKED ISSUE #2251: Named type not shown in inspector
When the record is collapsed, the named type appears, when it is expanded, its gone.

jameskerr:
Add the type name (if there is one) in both expanded and collapsed containers. Shown here is the person type and the data type.

image

Fixes #2251


mason-fish (approved):
looks like https://github.com/brimdata/brim/pull/2263 builds off this, which I already approved 👍

Issue #2274, opened 2022-03-15
React arborist bugs


jameskerr:
- [x] Be able to click and right click the bottom of the list without any items
- [x] Add the selectOnClick boolean an object with a property name
- [x] Rename to isSelectedStart and isSelectedEnd

Issue #2223, opened 2022-02-28
Sidebar CSS Polish & Bug Fixes


jameskerr:
- [ ] Add an active state
- [ ] Make the click target the whole row
- [x] Move the carrot next to the lake address a little closer to the text. Have it move with the text
- [x] Lighten the mag glass color in the search bar
- [ ] Plus button overflow ![image.png](https://images.zenhubusercontent.com/5e8cea85d10cba8a99efeb34/06c3de69-7c07-4f9b-9d1a-fb33af7ab902)
- [x] Multi-selection hightlight state  ![Screen Shot 2022-03-14 at 10.50.38 AM.png](https://images.zenhubusercontent.com/5e8cea85d10cba8a99efeb34/f99b64e2-cd20-4f49-8839-eed582e3d64b)
- [x] Don't allow clicking the folder and submitting a query ![Screen Shot 2022-03-14 at 10.50.58 AM.png](https://images.zenhubusercontent.com/5e8cea85d10cba8a99efeb34/6454013c-cdb3-416d-b96a-491278a0c4ee)
-[ ] Fix scrollbars in menu section

jameskerr:
There shouldn't be scroll bars at that top part. 
<img width="315" alt="image" src="https://user-images.githubusercontent.com/3460638/156853723-f30e8e99-e46f-407d-9048-9cecec263813.png">

jameskerr:
![image.png](https://images.zenhubusercontent.com/5e8cea85d10cba8a99efeb34/f0053ead-fce1-4f2b-bd7f-90421e71c44e)

PR #2277, opened 2022-03-15
Sidebar Item Refactor

jameskerr:
Refactor the sidebar to use a single component <Item /> for all sidebar items. This component handles progress states, editing states, selected states, icons, text, folder dropdown arrows. Having it all in one file will make working with new item's easier. And it will make it easier to make updates to all at once.

This also fixes some bugs that are outlined here: #2223

- [x] Make a folder click target the whole row
- [x] Multi-selection highlight
- [x] Don't submit a blank query when a folder is clicked
- [x] Fix scrollbars in the sidebar nav menu

This refactor is preparing for the new right click options that will be added in the next pr.  In order to do that I need to update react arborist by fixing this ticket: #2274 That will come next.

image

image

image

image

image


mason-fish (approved):
Nice! The `<Item />` refactor in particular seems like a solid win 👍

PR #2278, opened 2022-03-17
Query Promotion

### LINKED ISSUE #2231: Draft promotion UX
We currently have no way to promote our draft queries to either a a remote or local saved query. The only way to create a saved query is to use the old query modal ux, which is an added step we don't need any more and currently does not even take the values you may have already provided from your current draft.

So let's come up with a ux for taking an existing draft query and saving it either remotely or locally.

mason-fish:
fixes #2231 

Allows the promotion of a draft query => local and then from local => remote. Also grants some helper/ease-of-use functionality for copying, duplicating, or moving a query back and forth from one source to another.

I'm opening now because it's good to go as is but there is an open question I'd like to discuss on tab behavior when a query is deleted.

And here are some screenshots:
new header, draft query
image

rename by clicking on save, or by clicking on the title:
image

Cancelling, or hitting "escape" will return the draft's name back to the default, unsaved. Clicking "save" or hitting "enter" will promote the query to a locally saved query, at the root level of your local lib. It will continue to use the same tab you were on for the draft.
image

Adjusting the search bar will adjust the queries status to indicate that the query is "Modified", and issuing the search will then adjust the status to "Saved".
(search hasn't been executed yet)
image

(post execution/save)
image

A local or remote query has a dropdown for further actions
image

Copying or moving to a remote query
image


src/app/query-home/toolbar/query-page-header.tsx
```diff
@@ -0,0 +1,301 @@
...
+        dupeQ.name += "-copy"
+        if (querySource === "local") {
+          dispatch(Queries.addItem(dupeQ, "root"))
+          dispatch(Tabs.new(lakeQueryPath(dupeQ.id, lakeId)))
+        }
+        if (querySource === "remote") {
+          dispatch(setRemoteQueries([dupeQ])).then(() => {
+            dispatch(Tabs.new(lakeQueryPath(dupeQ.id, lakeId)))
+          })
+        }
```
jameskerr, line 251: It would be nice to have a single Query api for the CRUD operations that would find the source and do the right thing under the hood.

src/app/query-home/toolbar/query-page-header.tsx
```diff
@@ -0,0 +1,301 @@
...
+    {
+      label: "Rename",
+      click: () => setIsEditing(true)
+    },
+    {
+      label: "Duplicate",
+      click: () => {
+        const dupeQ = query.serialize()
+        dupeQ.id = nanoid()
+        dupeQ.name += "-copy"
```
jameskerr, line 242:
```suggestion
        dupeQ.name += " (copy)"
```

src/app/query-home/utils/query-params.ts
```diff
@@ -1,29 +1,11 @@
...
-const encodeBool = (params, value, name) => {
-  if (value) {
-    params.set(name, "true")
-  }
-}
-
-const decodeBool = (params, name) => {
-  const value = params.get(name)
-  return value === "true"
+// Will begin using this again when we start to encode ?version= in the url
```
jameskerr, line 6: I think we should include the version in the URL. For example `/queries/:id/version/:hash`.

And the lack of a version portion simply means latest.

jameskerr (approved):
A few ideas and suggestions. But good to go as far as I'm concerned.

src/app/query-home/toolbar/query-page-header.tsx
mason-fish: Yeah, as this has grown I've had a similar thought

src/app/query-home/utils/query-params.ts
mason-fish: I have a slight preference for the query param, but can be convinced to move it in the path. Let's discuss more when we start working on the query version history.

src/app/query-home/toolbar/query-page-header.tsx
jameskerr: Maybe this is belongs in the api.

PR #2286, opened 2022-03-24
Multiline Editor

### LINKED ISSUE #1521: Implement a Multi Line Editor for Brim
As we prepare the app for supporting the ability to edit shapers for their data, we are going to want to provide an inline Z editor to aid them. Ideally we want syntax highlighting, autocomplete, and other inspection features to make this process as easy as possible.

There are a few existing tools that can assist here. As far as the editor itself, the [monaco](https://microsoft.github.io/monaco-editor/) project seems like it would be a fine fit. It is maintained by Microsoft, powers VSCode, and appears to provide an api for registering new/custom languages.

We can also use [tree-sitter](https://tree-sitter.github.io/tree-sitter/) to generate the parser that the editor will use for syntax highlighting and autocomplete. Since we are currently in the middle of changing up the Z grammar, and the corresponding tree-sitter work would be coupled to the Z grammar, then it may be best to wait on using this tool until the language has stabilized a bit more.

mason-fish:
fixes #1521 

At long last, this finally introduces a multi-line editor to the app! The query-flow searchbar is now adaptive based on how many lines the query's value is: if it is only one line, we style the editor to appear as a searchbar, and if it is more than one then it will expand up to 3 lines with scrolling. In both cases we use the CodeMirror editor under the hood as the input type. I imported several useful extensions, including a couple that we are not yet using in this PR but plan on using later on as we continue the implementation (syntax highlighting, code folding, keyboard shortcuts, search within editor).

Currently looks like this:
(single)
image

(multi)
image

image

image

TODO:
- [x] Fix unnecessary app re-rendering on each input change
- [x] Hide behind feature flag
- [x] add back in a run/execute button in both stylings of the editor
- [x] fix cursor bugs


jameskerr:
Updating the redux store just for the value of the search input should be fast enough. If that is re-rendering a bunch of things, we need to make sure that other components don't depend on that piece of state. I don't think anything else needs to depend on it. If they need to depend on the "last run" query, they can select the last ran version of the query.

jameskerr:
We should chat about where to put this UI state. I think we need a new reducer. I am wondering the same thing for where to store the pins UI state.

src/js/state/Current/selectors.ts
```diff
@@ -51,9 +52,17 @@ export const getQuery = (state: State): BrimQuery | null => {
     Queries.getQueryById(queryId)(state) ||
     RemoteQueries.getQueryById(queryId)(state)
   if (!query) return null
-  return new BrimQuery(query)
```
mason-fish, line 54: this was the culprit of all of the re-rendering, since it always creates a new object

jameskerr:
Awesome work @mason! Before I review the code, here are some visual things I'm noticing that could be improved.

It looks like the line-height is off if there are scrollbars . I have mine set to always on because I use that wacom tablet alot. When they are on it looks like it breaks the design. Set overflow: hidden on the container when there is only one line. That should remove the scroll bars.

Screen Shot 2022-03-29 at 12 14 40 PM

Before we merge, let's get the styles right on the editor. We'll want `font-family: var(--mono-font)`. Here are the styles for it in sketch: https://www.sketch.com/s/c5a874f4-53b3-4af2-82fe-1212c45acda6/a/7yMzE08

And finally, instead of just giving them three lines when they switch to editor view, let's give them 150px. Something like this:

Screen Shot 2022-03-29 at 12 21 46 PM

In another PR, let's add the ability to resize the editor as large as they want. In a very future PR, we should let them put the editor and the results side by side, but that's just a thought I had right now.

src/app/core/components/brim-editor.tsx
```diff
@@ -0,0 +1,120 @@
...
+      extensions,
+      doc: inputValue
+    })
+    const newView = new EditorView({
+      state,
+      parent: ref.current
+    })
+    setView(newView)
+    newView.focus()
+  }, [ref])
```
jameskerr, line 115:
```suggestion
  }, [])
```

This will just run on mount because that ref object never changes its identity while the component is mounted.

You could do `[ref.current]` which will change, but the effect will run when something else causes it to render and ref.current has changed. As I'm sure you know, changing ref.current won't cause a re-render.

jameskerr (commented):
Code looks good!

jameskerr (approved):
I really like where we landed with this. It's ready. If you don't mind, let's just wait for the release to go out before merging.

philrz:
In case anyone else stumbles onto this and wants to check it out while it's still hidden behind a feature flag, I discovered the way to enable it in Dev mode:  

```  
yarn start --feature-flags=query-flow  
```

PR #2293, opened 2022-03-25
A few more changes to the v0.29 Migration article

philrz:
I did another proofreading pass on the article and found a few more possible improvements.

### DIFF
--- a/docs/Migration-for-Version-0.29.md
+++ b/docs/Migration-for-Version-0.29.md
@@ -6,7 +6,7 @@
 4. Open the app and see your pools.
 
 The migration script can be executed while Brim v0.29 is already open. Click **View > Reload** from the Brim pull-down menu to see your migrated pools after the script completes.
-> **Windows Users**: See the section below titled "How do I run this on Windows?".
+> **Windows Users**: See the section below titled "[How do I run this on Windows?](/#how-do-i-run-this-on-windows)".
 
 ## Will it delete anything?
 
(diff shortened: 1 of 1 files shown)


docs/Migration-for-Version-0.29.md
```diff
@@ -6,23 +6,23 @@
...
+No. Only run the steps above if you care to keep your old pool data. If you prefer a fresh start, you can delete your old pools by following the steps [below](#how-do-i-delete-the-old-pools).
 
 ## What does the script do?
 
 The Zed lake service that powers Brim made breaking changes to its data storage format in this release. Your pools are currently stored in a `data/lake/` subdirectory under the Brim [user data](https://github.com/brimdata/brim/wiki/Filesystem-Paths#user-data) path. The new version expects the pools to be in a `lake/` subdirectory (i.e., one level higher) under the user data path. The script will read in the old data and write it to the new location. The script does not attempt to delete anything on your system. If you want to delete the old pools after this script finishes, see [below](#how-do-i-delete-the-old-pools).
 
 ## How do I delete the old pools?
 
-You can remove the old pools by deleting the directory `<os-apps-dir>/Brim/data/lake/`.
+You can remove the old pools by deleting the directory shown in the table below for your OS.
```
philrz, line 25: I'm on-board with the convenience factor of keeping the table of paths to make it easier for the user. My one remaining concern is that the [Filesystem Paths](https://github.com/brimdata/brim/wiki/Filesystem-Paths) article differentiates between the "Application Binaries" path and the "User Data" path, so I felt a shorthand like `<os-apps-dir>` (with "app" in it) might confuse. It seems like we can let the table do most of  the talking, especially since it includes the `/` and `\` being different between Unix and Windows).

jameskerr: Right on!

jameskerr (approved):

PR #2312, opened 2022-04-05
fix query sidebar context menus

### LINKED ISSUE #2242: Fix context menu items
There are several context menu items that are still assuming the current pool id is set (Use as "start"/"end" time, View in full context, ...). These should all be exercised and fixed if broken. Includes context menu for detail window (which currently doesn't even open for the same current pool id assumption problem).

mason-fish:
fixes #2242 

In addition to fixing the run/copy enabling and execution, this also includes a behavior change: moving forward, when you right click on an item the context menu's actions will refer to just that one item. The delete action (and in the future, any other action which can be applied to multiple items) will act solely on the right clicked item unless that item is "selected" (highlighted in blue) along with at least one other item. In that case, all selected/highlighted items will participate in the operation.

right click on an item that is NOT part of selection
image

right click on an item that IS part of a multi selection
image

right click on an item that is the only item selected
image

renaming, running, copying don't care about selection and just operate on whatever you clicked on
image

### DIFF
--- a/src/app/features/sidebar/hooks/index.ts
+++ b/src/app/features/sidebar/hooks/index.ts
@@ -4,23 +4,9 @@ import {NativeTypes} from "react-dnd-html5-backend"
 import {DragItem, DragProps} from "src/app/features/import/use-import-on-drop"
 import {useBrimApi} from "src/app/core/context"
 import {useMemo} from "react"
-import {useDispatch, useSelector} from "react-redux"
+import {useSelector} from "react-redux"
 import brim from "src/js/brim"
 import Investigation from "src/js/state/Investigation"
-import {AppDispatch} from "src/js/state/types"
-import Current from "src/js/state/Current"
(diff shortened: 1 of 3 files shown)


jameskerr:
It's unexpected to me that when selecting multiple items, we present options that only operate on one item. I would think that when I click multiple, I only see options that apply to them all.

src/app/features/sidebar/hooks/index.ts
```diff
@@ -67,35 +66,31 @@ export const useSearchHistory = () => {
   }, [historyEntries])
 }
 
-export const useQueryItemMenu = (data, tree, handlers) => {
+export const useQueryItemMenu = ({data, tree, handlers, lakeId}) => {
```
jameskerr, line 69: Would you mind turning this into a thunk, so that the component doesn't need to re-render whenever one of these useSelector fires?

jameskerr (approved):
Left some comments and approve so that you can merge as you see fit.

PR #2313, opened 2022-04-05
Lockable Queries

### LINKED ISSUE #2234: Lockable Queries
By adding the behavior of auto-update-on-run to each query, we have prioritized and optimized the workflow for our observable integrations, but we also need to be able to lock a query from being saved/published on every run. Particularly since we don't have the query history in place yet, this always-write behavior could be frustrating when your query workflow is meant to fork off of an initial query as you investigate the shape of your data.

So we should include a boolean `isLocked` in each saved query ("saved query" here refers to both the local and remote query models, and not draft queries), and then decide on the app behavior for when someone wants to manipulate a locked query's value.

mason-fish:
fixes #2234 

Putting this initial commit up as a draft for ease of discussion. Some of the other changes will come after https://github.com/brimdata/brim/pull/2312 gets merged since there was overlap.

We still need to decide how the search bar and pins' intractability will change based on this locked condition. One solution is to disable each of these elements (search bar grayed out, "no-access" cursor, pins can also change their background/pill color to gray and not allow edits) with corresponding tool-tips to explain why, but can still be executed as usual.  

placeholder lock for now, need svg from our own designs
image

image

unicode's open lock is barely different than the closed...
image


src/app/query-home/toolbar/query-page-header.tsx
```diff
@@ -199,8 +199,21 @@ const QueryPageHeader = () => {
   }
 
   const menu = usePopupMenu([
```
jameskerr, line 201: Would you mind making this a thunk too? To avoid the unnecessary renders.

jameskerr (approved):
I see, this PR doesn't disable any inputs yet. It just provides the mechanism to toggle the locked state. Looks good!

PR #2314, opened 2022-04-05
Update auto-detect docs to reflect CSV & JSON support

philrz:
I noticed the auto-detect expectations in the app and the Brim wiki didn't yet reflect the CSV and regular JSON support added in brimdata/zed#3277 and brimdata/zed#3124, respectively. This PR proposes bringing that current.

Note that when this merges, the link to [https://github.com/brimdata/brim/wiki/Importing-CSV%2C-Parquet%2C-and-ZST](https://github.com/brimdata/brim/wiki/Importing-CSV%2C-Parquet%2C-and-ZST) that's still out there in currently-installed Brim apps will technically be broken. However, such broken links thankfully redirect to the top-level [https://github.com/brimdata/brim/wiki/](https://github.com/brimdata/brim/wiki/) where the user could easily find the relevant article from among the titles, so that seems like not a bad compromise. If this was considered not seamless enough, an alternative would be to hold back on merging this change until right when the next GA release goes out.


docs/Importing-Parquet-and-ZST.md
```diff
@@ -21,7 +20,7 @@ that can be easily imported into Brim.
 
 # Example
 
-Preprocessing can be performed with the [`zq`](https://github.com/brimdata/zed/tree/main/cmd/zq)
+Preprocessing can be performed with the [`zq`](https://github.com/brimdata/zed/blob/main/docs/zq/README.md)
```
mason-fish, line 23:
```suggestion
Preprocessing can be performed with the [`zq`](https://github.com/brimdata/zed/blob/main/docs/zq/README.md)
```
```suggestion
Preprocessing can be performed with the [`zq`](https://github.com/brimdata/zed/blob/main/docs/commands/zq.md)
```

mason-fish (approved):
lgtm after fixing just one zed docs link that just got changed after you'd already put this PR up 👍

docs/Importing-Parquet-and-ZST.md
philrz: @mason-fish: Per our discussion on Slack, if the new Zed docs site is where things are converging, then instead of fixing the links to point into the `docs/` location within the GitHub repo it seems like it might make sense to instead point to the relevant page below [https://zed.brimdata.io/docs/.](https://zed.brimdata.io/docs/.) I've just pushed a commit to adjust three links in here to do that. If you agree the approach makes sense, feel free to merge, as I don't have the permissions to do that myself. Or if there's a desire to keep linking into the GItHub docs location for now, let me know and I can revert that commit and go with your suggestion instead.

philrz:
I think folks know this, but the Markdown link checker keeps breaking in general because there's a bunch of other articles that have similarly broken links. As I discussed with @mason-fish on Slack, I'm game to do another pass through all the wiki docs and fix the others once I've got confirmation that linking to the new Zed docs site is the way to go. But since I think this smaller PR is ready to go in as-is, it seems like it could be merged first to kick things off,.

PR #2321, opened 2022-04-07
remove space migration code and documentation

nwt:
This stuff is nearly a year old, so it's unlikely anyone will miss it.

After this is merged, I plan to remove the `brimcap migrate` command, on which space migration relies.

PR #2322, opened 2022-04-08
Fix broken links in Brim docs

philrz:
As pre-sold to @mason-fish and @nwt, this is the first of what I expect to be a few passes through the Brim docs. I'd initially envisioned doing it all in one big PR, but @mason-fish made a good argument to fix the broken links first so the link checker would stop freaking out.

In subsequent PRs, I expect to:

1.  Purge some docs coverage that are specific to older releases, similar in spirit to what @nwt did with #2321.
2.  Correct some of the Zed CLI invocations in the wiki articles to be in sync with newer command syntax.


mason-fish (approved):
Right on, thanks @philrz! We should have been version-pinning those links in the changelog all along.

PR #2326, opened 2022-04-19
Query Version History

### LINKED ISSUE #2235: Query History
Locally saved queries do not have any memory of themselves, and so if we want to support a history for these we need to first design it in their model.

Remote queries have a different storage mechanism, using a zed lake, and history is already a part of their model.

That leaves the UX. A way to display (do we also want linnear and tree view like we have for global app history?) and manipulate (selecting current entry, what happens if you now edit/save the query?). We had previously decided that this larger ux piece could fit nicely in the right sidebar, and if so then we would also need an accompanying right sidebar refactor/re-design since this would be the first time the right sidebar is used for more than 1 thing (currently only used for selected log detail)

mason-fish:
fixes #2235 

Query Versions! This deepens our query model to store queries' pins & values in an array of tagged versions. We determine what "current" version you are on based on the location param, otherwise default to latest. The new `right-pane` allows users to view and navigate through a given query's version history as a new "Versions" tab next to the existing "Detail" tab we use to look at record details.

The Search bar is also always in multiline mode (though it only shows line numbers if you have more than one line).

With this in place we can also fix the global history by saving copies of the query version as they are executed.

One important note is that there is still some work to do to make this work **without** the feature flag, but I was thinking that we should just turn it off in main and start to remove all the dead code that will produce. Second note is that I didn't yet include any migration for existing queries. I figured on handling that in a follow up PR. And finally, I removed the brim-queries plugin, a.k.a. the brim query lib. So for the next release's release notes we should explain that those can be re-imported by heading over to the brimcap repo (where we will move the json to).

image


src/app/query-home/search-area/pins/pins.tsx
```diff
@@ -29,11 +28,6 @@ function renderPin(pin: QueryPin, index: number) {
 
 export function Pins() {
   const pins = useSelector(Editor.getPins)
-  const dispatch = useDispatch()
```
mason-fish, line 32: Since `submitSearch()` now creates a new draft for the current query, this created an infinite loop. Can also fix it by comparing the editor pins to the query, but I thought executing a query after editing a pin didn't match with how we don't execute after editing the value so I removed. Happy to discuss more!

src/js/components/QueriesModals/QueryForm.tsx
```diff
@@ -163,35 +163,39 @@ const QueryForm = ({onClose, query, value, isRemote}: Props) => {
         pins,
         from,
       }
-      if (isRemote) {
+      if (!isRemote) {
```
mason-fish, line 166: This is gross but will go away when we turn on the query-flow feature flag

src/js/state/Current/selectors.ts
```diff
@@ -34,34 +35,40 @@ export const getLocation = (state: State) => {
...
-    Queries.getQueryById(queryId)(state) ||
-    RemoteQueries.getQueryById(queryId)(state)
-  if (!query) return null
-  return query
-}
-
-export const getQuery = createSelector<State, Query | null, BrimQuery>(
-  getRawQuery,
-  (query) => {
+export const getQueryById =
```
mason-fish, line 49: this is a generic query selector (i.e. it will select from either of the 3 query type slices (draft, local, remote), so I wasn't sure which selector code to put it and just settled on Current for now. Open to other suggestions.

src/app/features/right-pane/version-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+    &:hover {
+      background-color: var(--primary-color);
+    }
+  }
+`
+
+const useForcedRenderInterval = (interval = 60000) => {
+  const [renderTrigger, setRenderTrigger] = useState(true)
+  setTimeout(() => setRenderTrigger(!renderTrigger), interval)
+}
```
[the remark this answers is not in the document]
jameskerr:
```
const [render, setRender] = useState()

useEffect(() => {
  const id = setTimeout(() => setRender(r => !r), interval)
return () => { clearTimeout(id)}
}, [])
```

Wait ,do you need clear interval and set interval?

src/app/features/right-pane/versions-section.tsx
```diff
@@ -0,0 +1,36 @@
...
+  margin-top: 8px;
+`
+
+const VersionsSection = () => {
+  const {resizeRef, defaults} = useSectionTreeDefaults()
+  const query = useSelector(Current.getQuery)
+
+  return (
+    <SectionWrapper ref={resizeRef}>
+      <Tree
```
jameskerr, line 20: Are these nodes draggable?

jameskerr: It should probably be disabled.

src/app/query-home/flows/submit-search.ts
```diff
@@ -12,29 +11,27 @@ const submitSearch =
...
-      pins,
+    dispatch(
+      newQueryVersion(prev.id, {...prev.currentVersion(), value, pins})
+    ).then(() => {
+      const query = Current.getQuery(getState())
+      const error = query.checkSyntax()
+      if (error) {
+        dispatch(Results.error(error))
+        return
+      }
```
jameskerr, line 28: We should check the syntax before we save it.

src/app/query-home/search-area/pins/pins.tsx
jameskerr: Let's discuss! I think we should try to submit searches after every keystroke with a throttle!

src/app/routes/app-wrapper.tsx
```diff
@@ -82,7 +82,7 @@ export default function AppWrapper({children}) {
               >
                 {children}
               </ColumnLayout>
-              <XRightPane />
+              <FeatureFlag name="query-flow" on={null} off={<XRightPane />} />
```
jameskerr, line 85: Nifty!

src/app/search/flows/md5-search.ts
```diff
@@ -46,12 +45,7 @@ export const md5Search = (md5: string): Thunk<Promise<ResultStream>> => {
 
     return dispatch(
       querySearch({
-        query: new BrimQuery({
-          id: "",
-          name: "",
-          value: q,
-          pins: [{type: "from", value: poolId}],
-        }),
+        query: `from ${poolId} | ${q}`,
```
jameskerr, line 48: much nicer

src/js/api/index.ts
```diff
@@ -103,13 +109,19 @@ export default class BrimApi {
   }
 
   importQueries(file: File) {
-    const node = parseJSONLib(file.path)
-    this.dispatch(Queries.addItem(node, "root"))
-    this.toast.success(`Imported ${node.name}`)
+    const {libRoot, versions} = parseJSONLib(file.path)
+    this.dispatch(Queries.addItem(libRoot, "root"))
+    console.log("versions: ", versions)
```
jameskerr, line 114: Some console logs in here

src/js/state/Appearance/index.ts
```diff
@@ -1,13 +1,14 @@
...
 import {State} from "../types"
-import {HistoryView, QueriesView, SectionName} from "./types"
+import {HistoryView, PaneName, QueriesView, SectionName} from "./types"
 
 const init = () => ({
   sidebarIsOpen: true,
   sidebarWidth: 230,
   sidebarSections: [{id: "pools"}, {id: "queries"}, {id: "history"}],
   currentSectionName: "pools" as SectionName,
+  currentPaneName: "versions" as PaneName,
```
jameskerr, line 11: I think should should be in the Tab state. Instead of the window state.

src/js/state/QueryVersions/index.ts
```diff
@@ -0,0 +1,71 @@
...
+import {last} from "lodash"
+import {QueryPin} from "../Editor/types"
+
+const versionAdapter = createEntityAdapter<QueryVersion>({
+  selectId: (queryVersion) => queryVersion.version,
+  sortComparer: (a, b) => (a.ts > b.ts ? 1 : -1),
+})
+const initialState = versionAdapter.getInitialState()
+const versionSlice = createSlice({
+  name: "$version",
```
jameskerr, line 11: maybe $QUERY_VERSIONS would be a better description.

src/js/state/QueryVersions/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  raw: (state) => state.queryVersions,
+  getByQueryId: (queryId) => (state) => state.queryVersions[queryId],
+  getLatestByQueryId: (queryId) => (state) => {
+    const q = state.queryVersions[queryId]
+    return q?.entities[last<string>(q?.ids)]
+  },
+  getByVersion: (queryId, version) => (state) =>
+    versionAdapter
+      .getSelectors()
+      .selectById(state.queryVersions[queryId], version),
```
jameskerr, line 70: Would you move these to a selectors file.

src/js/state/QueryVersions/index.ts
```diff
@@ -0,0 +1,71 @@
...
+export type QueryVersionsState = {
+  [queryId: string]: VersionsState
+}
+
+export type QueryVersion = {
+  version: string
+  ts: Date
+  value: string
+  pins?: QueryPin[]
+}
```
jameskerr, line 39: Would you move these to a types file.

src/app/query-home/flows/new-query-version.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    const source = dispatch(getQuerySource(queryId))
+    if (source === "remote") {
+      const query = RemoteQueries.getQueryById(queryId)(getState())
+      await dispatch(setRemoteQueries([{...query, ...version}]))
+    }
+
+    dispatch(QueryVersions.add({queryId, version}))
+    dispatch(tabHistory.push(lakeQueryPath(queryId, lakeId)))
```
jameskerr, line 32: We probably shouldn't change the location in this file. Whatever calls this should change the location.

src/app/query-home/flows/new-query-version.ts
```diff
@@ -0,0 +1,35 @@
+import {nanoid} from "@reduxjs/toolkit"
```
jameskerr, line 1: Let's move all these files into the src/js/State area of the codebase.

src/app/features/right-pane/version-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const query = useSelector(Current.getQuery)
+  const lakeId = useSelector(Current.getLakeId)
+
+  const onClick = (e) => {
+    e.preventDefault()
+    handlers.select(e, {selectOnClick: true})
+    dispatch(
+      tabHistory.push(lakeQueryPath(query.id, lakeId, queryVersion.version))
+    )
+    dispatch(Results.fetchFirstPage(query.toString()))
```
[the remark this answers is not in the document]
mason-fish: will replace with submitSearch, which will now have 2 save parameters to either save the new search to global history and/or a new version

jameskerr (approved):
You may merge when ready!

src/app/features/right-pane/versions-section.tsx
mason-fish: Oh, actually defaults already do disabledrag

PR #2327, opened 2022-04-19
Upgrade to React 18

mason-fish:
React 18 is here! 🎉 

To stay in sync with the new react 18 types and other react 18 changes, I have also upgraded:

* react testing lib packages
* react-dnd
* react-router (just a patch for now, the next major version will have the react type fixes we need but also has several breaking changes so let's do that as it's own ticket)
* react-redux, major version. Was straightforward, just needed some more explicit typing when we `useDispatch` and `useStore`
* react-window, patch
* styled-components, a couple of minor versions

To fix some of our install warnings I also upgraded prettier and added @babel/core to our deps. The only warning we get now is an upstream issue with react-hot-toast that I'm pretty sure we have no control over (other than submitting a PR there). The upgrade to prettier includes a some new style [defaults](https://prettier.io/blog/2020/03/21/2.0.0.html) which I tried to capture most of in my final commit to try and reduce noise for review.


jameskerr (commented):
As we discussed in the sync today, update the types first to version 18, then I think this can ship.

jameskerr (approved):
Let's do it!

PR #2331, opened 2022-04-24
Simplify Brim wiki

philrz:
This PR proposes a significant overhaul to the Brim wiki to update and simplify. I'll put some inline comments to explain some changes whose justification may not be immediately obvious.


CHANGELOG.md
```diff
@@ -116,12 +116,12 @@ particular we'd like to bring to your attention first.
      before `v0.25.0` will start. Auto-update notifications for Windows will
      resume on the
      next release. See the
-     [Installation](https://github.com/brimdata/brim/wiki/Installation#windows-installation-v0250)
+     [Installation](https://github.com/brimdata/brim/blob/v0.25.0/docs/Installation.md#windows-installation-v0250)
```
philrz, line 119: For many of these old wiki docs referenced from the changelog, I've changed them to point to tagged references in the `docs/` directory. One of the good/bad properties of the GitHub wiki is that links to pages that don't exist end up being redirected to the top-level wiki home. That's arguably better than a 404, but if there was any reason for a user to actually reference some old doc because it's mentioned in the changelog, I figure actually pointing them at it is preferable to starting them at the home page and them not seeing the doc. For docs that still live at their original location referenced in the changelog, I've left those links as is.

docs/Adding-Migrations.md
```diff
@@ -2,7 +2,7 @@ Because we persist state on a user's computer, we must migrate that state each t
 
 ### Creating a Migration
 
-Let's say we've added a property called `nickname` to each zqd connection in the app. We'll use the migration generator tool to create the necessary files. From the root directory, run `bin/gen migration <name>`. For example:
+Let's say we've added a property called `nickname` to each lake connection in the app. We'll use the migration generator tool to create the necessary files. From the root directory, run `bin/gen migration <name>`. For example:
```
philrz, line 5: I actually don't know if the contents of this "Creating a Migration" doc are still accurate, as I didn't write it. However, I figured `zqd` references are fair game to delete regardless. :smile:

docs/Home.md
```diff
@@ -5,24 +5,21 @@ effective use of the Brim desktop application and related tools.
...
 - [[Installation]]
 - [[Filesystem Paths]]
 - [[Supported Platforms]]
-- [[Microsoft Windows beta limitations]]
 - [[Troubleshooting]]
 
 ## User Documentation
 
 - [[Migration for Version 0.29]]
-- [[Linux RPM Upgrade]]
```
philrz, line 25: Here's a little history about this RPM article and an explanation of why I propose getting rid of it.

The new installer that arrived with Brim `v0.25.0` marked a point where users of the RPM packages had to manually uninstall the `v0.24.0` that used the old installer before the new version would install successfully, since otherwise there'd be errors about "conflicts". At the time I mistakenly thought that was a one-time problem specific to the point of transition between installers. However, testing with more recent releases has shown me that the problem still shows up between more recent releases that all use the new installer. Therefore, in this PR I've moved the uninstall guidance to the main Installation article so the standalone RPM article can be removed.

Furthermore, I took the opportunity to take a closer look at the experience with the newer installer and DEB packages. While this doesn't have the "conflict" problem that the RPM package has and the install appears to complete successfully, if I do the install from the CLI, it explains that this effectively installs the new version _atop_ the older one. This probably causes no real problems, but it seems messy because it has the potential for "droppings" to be left behind in old installs that are no longer referenced in newer ones. That might create confusion in debug scenarios.

Therefore, I chose to keep it simple in the Installation article and recommend an uninstall before upgrade on all Linux platforms.

docs/Joining-Data.md
```diff
@@ -8,9 +8,8 @@
 # Summary
 
 Similar to the well-known [SQL construct](<https://en.wikipedia.org/wiki/Join_(SQL)>),
-the Zed language provides a `join` operator that can be used to combine data
-sources. This cookbook describes how it can be used with the Brim application
-and discusses its limitations.
+the Zed language provides a [`join`](https://zed.brimdata.io/docs/language/operators/join/) operator that can be used to combine data
+sources. This cookbook describes how it can be used with the Brim application.
```
philrz, line 12: Since multiple pools can now be referenced in a Zed query in Brim, a case could have been made to drop this Join article altogether. However, since the multi-line editor is not yet here, and app-heavy users may not have yet read the Zed docs closely yet, I figured a few targeted refs to help the user successfully run through the Zed language Join tutorial might still be worthwhile. Maybe a few users have it bookmarked, right?

docs/Remote-Lakes.md
```diff
@@ -43,8 +41,9 @@ Before diving into the specifics of what's possible, here's an overview of
 some rough edges you may encounter as you work through the configurations
 described in this article.
 
-1. While **logs** can be imported from your local Brim app directly to a remote
-   Lake, **packet captures** currently cannot ([brim/1730](https://github.com/brimdata/brim/issues/1730)).
+1. If a user imports a packet capture while connected to a remote lake, the
+   generated logs are stored remotely but the pcap file and corresponding
+   index for flow extraction remain local to the user's workstation.
```
philrz, line 46: I was actually quite shocked to find this has been already working for so long! See [https://github.com/brimdata/brim/issues/1730#issuecomment-1106699704](https://github.com/brimdata/brim/issues/1730#issuecomment-1106699704) for details.

docs/Supported-Platforms.md
```diff
@@ -57,7 +57,7 @@ on releases older than Windows 8.1.
 
 Brim's [test automation](#automated-testing) runs on Catalina 10.15 and
 therefore this is the macOS version on which we are best able to ensure quality
-and prevent regressions. Several Brim developers also run macOS Big Sur 11.0
+and prevent regressions. Several Brim developers also run macOS Monterey 12.3
```
philrz, line 60: I can't speak for the official Brim Dev team, but I know when I worked with y'all, folks generally jumped to the new macOS releases as they came out, so here I'm assuming that's still the case.

docs/Supported-Platforms.md
```diff
@@ -92,10 +92,11 @@ builds ([brim/1266](https://github.com/brimdata/brim/issues/1266)).
 ## Linux
 
 Brim's [test automation](#automated-testing) runs on Ubuntu 18.04 (`.deb`
-package) and CentOS 8 (`.rpm` package) and therefore these are the platforms
-on which we are best able to ensure quality and prevent regressions.
+package) and therefore this is the Linux distribution on which we are best able
+to ensure quality and prevent regressions.
```
philrz, line 96: I corrected this statement to accurately reflect that the CI no longer tests on CentOS. See [https://github.com/brimdata/brim/pull/1985#discussion_r766045844](https://github.com/brimdata/brim/pull/1985#discussion_r766045844) .

docs/Filesystem-Paths.md
```diff
@@ -25,9 +25,9 @@ only modified when a newer version of the app is being installed.
 Since Brim is most often launched directly from the GUI environment in your OS,
 you'll typically not need to access the application binaries directly. One
 exception is if you wish to make use of the command line binaries that are
-bundled with the app, such as [`zq`](https://github.com/brimdata/zed/tree/main/cmd/zq),
-[`zed`](https://github.com/brimdata/zed#quick-start),
-[`brimcap`](https://github.com/brimdata/brimcap), and so forth. These can be
+bundled with the app, such as [`zq`](https://zed.brimdata.io/docs/commands/zq/),
+[`zed`](https://zed.brimdata.io/docs/commands/zed/),
+[`brimcap`](https://github.com/brimdata/brimcap/blob/main/README.md), and so forth. These can be
```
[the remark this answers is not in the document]
philrz: @nwt: Ok, personal preference I guess. I suggested linking to the Brimcap README because the links to `zq` and `zed` were pointing to "docs that describe how to use them" rather than "their repos", and in this regard I find a direct link to the Brimcap README a bit more apples-to-apples. But it ultimately lands at a page with the same content, so I'm happy to go with your changes.

nwt (approved):

PR #2340, opened 2022-05-02
Tab Previews

### LINKED ISSUE #2273: Preview Tab
If you run through and click every single query in the sidebar, you'll get a new tab for each one. This might not be what you want if you are just clicking through the queries to see what they are. Let's make it so that.

1. a single click creates a tab in a preview mode. The tab will be reused (or destroyed and recreated) if the user clicks another item in the sidebar. We can indicate it's in preview mode by making the title italic.
2. A double click opens a tab normally like today

jameskerr:
Fixes #2273 

When using the query-flow feature flag, single clicking on an item in the sidebar will open that item in a preview tab.

Preview Tabs:
1. Have an italic title
1. Remain in preview mode until a user clicks or performs a mousedown in the main pane (there may be a smarter way to determine when to switch to normal tab mode that I can't think of at the moment)
2. Are reused if the user previews something else
3. Only one preview tab can exist at a time

To open a url in a preview tab, call `Tabs.previewUrl(url)`.
To open a url in a normal tab, call `Tabs.activateUrl(url)`

The onClick handler for the sidebar items creates preview tabs, while the onDoubleClick handler creates normal tabs.

Other Changes:

* Renamed `Tabs.activateByUrl` to `Tabs.activateUrl`
* Renamed `Tabs.new` to `Tabs.create` because new is a keyword in JS
* Rewrote the `Tabs.reducer` to use `@reduxjs/toolkit`

@mason-fish please pull this branch and test it out to make sure it makes sense to you.

https://user-images.githubusercontent.com/3460638/166326372-fe4475f7-6cfc-41d7-9ff5-902b18336699.mov

Here's a video of the action.


src/js/state/TabHistories/index.ts
```diff
@@ -12,12 +12,11 @@ const slice = createSlice({
     save: adapter.setAll,
   },
   extraReducers: {
-    TABS_ADD: (state, action: TABS_ADD) => {
-      global.tabHistories.create(action.id, [action.url])
-      return state
-    },
```
jameskerr, line 18: I moved the creation of the tabHistory to `Tab.init(url)`.

jameskerr:
@mason-fish I think this is good to go know. I cant think of anything else that I like to indicate that a tab is a "preview" other than the italics. Let's see if this behavior is confusing to others and go from there.

mason-fish (approved):
Right on, love this feature! Hopefully I'm the only one who doesn't notice the italics (or cares to have an indication of the preview tab)

PR #2341, opened 2022-05-02
This is a function

jameskerr:
Instead of being a getter, this is actually a function.


mason-fish (approved):
Ah nice catch, though why not just make it a getter instead?

jameskerr:
I think this but snuck in because I thought it _was_ a getter. When do you think we should use a getter vs a plain function?

PR #2342, opened 2022-05-03
Electron 15, Node 16

mason-fish:
Upgrades electron to `v15` and node to `v16.5.0`. There were actually no changes to the code required. The one potentially breaking change in this version bump is the new default to set `nativeWindowOpen: true` ([docs](https://www.electronjs.org/blog/electron-15-0#breaking-changes)). I decided to keep this default and see if everything would work with the new native windows (which are the new default because they perform better than the previous `BrowserWindowProxy`) and it appears that this actually doesn't break anything for us.

The main additional changes required then, were to upgrade to node `v16.5.0`, and also to replace the deprecated [node-sass-chokidar](https://github.com/michaelwayman/node-sass-chokidar#deprecated) (which depended on node 14) with [sass](https://github.com/sass/dart-sass) and tweak its corresponding build script a bit.

Just looking ahead at the breaking changes documented for electron `v16`, `v17`, and `v18`, I suspect we might be able to get to latest very easily. I only see [one](https://www.electronjs.org/blog/electron-16-0#building-native-modules) issue that may apply to us in v16. 🤞


nwt:
@mason-fish: Don't forget to update the Node.js version in `.nvmrc`!

jameskerr (approved):
Thanks @mason-fish ! If you wouldn't mind updating the .node-version file as well. Thanks!

PR #2343, opened 2022-05-03
Work around docs.github.com link checker problem

philrz:
See brimdata/zed#3871 for background.

In addition to putting back a link I'd removed in a docs review, I figure having the workaround in place could be helpful for future generations since someone who innocently adds a link to docs.github.com might otherwise be left scratching their head.


jameskerr (approved):
Makes sense to me.

PR #2344, opened 2022-05-04
simplify commit message in .github/workflows/advance-zed.yml

nwt:
The commit log is very noisy because of the level of detail in these
commits messages.


jameskerr (approved):
I can dig it.

mason-fish (approved):

PR #2345, opened 2022-05-04
Zast: JavaScript Zed Ast Manipulator

jameskerr:
There are many places in the app where we need to modify the query string. This is hard stuff. For a while we've been hacking around by doing a best guess regex replace. This draft PR is a minimal attempt at what AST manipulation might look like.

Questions I have:

 - [ ] Does the language seem stable enough to justify working on this? Can we commit to maintaining the peg parse and this javascript library.
 - [ ] Will white-space be able to carry over? Right now, I don't see whitespace nodes in the AST. It seems like those will be  needed if we don't want to reduce the user's query to a single line every time they use the right click options.

### DIFF
--- a/src/app/core/zast/binary-expr-node.ts
+++ b/src/app/core/zast/binary-expr-node.ts
@@ -1,0 +1,16 @@
+import {build, ZastNode} from "./build"
+import {BinaryExpr} from "./types"
+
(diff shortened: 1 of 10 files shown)


jameskerr:
Closing this. The right time will come to build something like this.

PR #2347, opened 2022-05-04
Fix about window

### LINKED ISSUE #2332: "About" window no longer works (regression at 798e664)
Repro is with commit 798e664b3438ac64a0f6f43a6103f93fae8ca274.

As shown in the attached video, selecting Electron -> About pull-down menu no longer renders the window with version info. I don't see any error message in the DevTools console.  I've confirmed via binary search that this problem started appearing with the changes at commit 798e664b3438ac64a0f6f43a6103f93fae8ca274, which is associated with #2327 (cc: @mason-fish).

https://user-images.githubusercontent.com/5934157/165404666-f807a654-8fe7-4d8e-9c85-74872458bae0.mp4

jameskerr:
The about window was not working because we were looking for the wrong div id in the html.

Fixes  #2332

### DIFF
--- a/src/js/about.tsx
+++ b/src/js/about.tsx
@@ -4,6 +4,6 @@ import AboutWindow from "./components/AboutWindow"
 import lib from "./lib"
 import {createRoot} from "react-dom/client"
 
-const container = lib.doc.id("app-root")
+const container = lib.doc.id("about-root")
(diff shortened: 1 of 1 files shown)


mason-fish (approved):
Doh! Nice catch

PR #2353, opened 2022-05-09
Do not redirect to the same place

### LINKED ISSUE #2352: Switching Tabs Erases Editor State
In Brim, if you switch tabs the editor state gets blown away:

ZHxv0M67oq

jameskerr:
This redirect was causing the location to change twice each time
the tab rendered. This caused the editor state to be cleared.

Fixes #2352


src/app/pools/show.tsx
```diff
@@ -27,14 +27,9 @@ export default function PoolShow() {
...
           }
           off={
-            <>
-              <Route path={lakePoolSearch.path}>
-                <SearchHome />
-              </Route>
-              <Route default>
-                <Redirect to={`${match.url}/search`} />
-              </Route>
-            </>
```
jameskerr, line 37: It didn't seem like this was needed.

mason-fish (approved):
Hmm, I thought we needed this for some special case but I just checked this out and the routing seems to work as expected so 🤷 👍

jameskerr:
@mason-fish I know what you mean. I hope this didn't cause a breakage somewhere else. I couldn't find any place, so fingers crossed.

PR #2354, opened 2022-05-10
Fix pivot to logs when it has a star

jameskerr:
This fixes the problem Steve encountered today with pivot to logs.

The problem was the "*" in the filter. There was a regex that looks for a star that clears the filter. But it should only look for a "lone" star, not a star that's within other stuff.


mason-fish (approved):
⭐️

PR #2355, opened 2022-05-13
Fix Infinite Scroll

jameskerr:
**Results State**
I created a Results state slice that is responsible for the main results. It looks like this:

```
{
    values: [] as zed.Value[],
    status: "INIT" as ResultsStatus,
    page: 1,
    perPage: 500,
    aggregationLimit: 2000,
    aggregation: false,
    key: "",
    query: "*",
    error: null as null | any,
  }
```

The Viewer state slice will now only store data related to the Table view, just like the Inspector only stores data related to the Inspector view.

**Unused Icons**

I discovered some old icons that could be deleted.

**Viewer Multi Selection**

We had lots of unused code that was supposed to be for multi selection in the viewer. I removed it. I wrote that type of code again in react-arborist and it's a little better. So when we need it in the app, we'll copy it from react-arborist.

**Infinite Scroll Algorithm**

The infinite scroll now takes the head of perPage * pageNumber, then takes the tail of perPage. For aggregations we simply say you only get 2000 results.


src/app/query-home/results/results-table/index.tsx
```diff
@@ -18,26 +18,26 @@ import getEndMessage from "./get-end-message"
...
   height: number
   width: number
-  multiSelect: boolean
 }
 
 const ResultsTable = (props: Props) => {
   const queryValue = useSelector(Current.getQuery)?.value
-  const isFetching = useSelector(Viewer.getStatus) === "FETCHING"
-  const isIncomplete = useSelector(Viewer.getEndStatus) === "INCOMPLETE"
+  const isFetching = useSelector(Results.isFetching)
```
mason-fish, line 31: 👍

src/js/brim/program.test.ts
```diff
@@ -243,7 +243,7 @@ describe("#hasAnalytics()", () => {
       brim
         .program("* | fork ( => cut uid, _path => cut uid ) | tail 1")
         .hasAnalytics()
-    ).toBe(true)
+    ).toBe(false)
```
mason-fish, line 246: why did this flip?

src/js/components/SearchResults/getEndMessage.ts
```diff
@@ -1,16 +1,18 @@
-import {ANALYTIC_MAX_RESULTS} from "../../flows/config"
-import {withCommas} from "../../lib/fmt"
-import brim from "../../brim"
+// import {ANALYTIC_MAX_RESULTS} from "../../flows/config"
```
mason-fish, line 1: Seems like you can indeed remove

src/js/state/Results/reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+      s.page = 1
+      s.status = "FETCHING"
+      s.error = null
+    },
+    nextPage(s) {
+      s.page += 1
+      s.status = "FETCHING"
+    },
+    setValues: {
+      prepare: (values: zed.Value[], tabId: string) => ({
```
mason-fish, line 34: nifty, didn't know about `prepare`

src/js/state/Viewer/reducer.ts
```diff
@@ -1,15 +1,11 @@
 import produce from "immer"
 
 import {ViewerAction, ViewerState} from "./types"
-import {concat, splice} from "../../lib/Array"
-import {createSelection} from "./helpers/selection"
-import {zed} from "@brimdata/zealot"
+
+// import {createSelection} from "./helpers/selection"
```
mason-fish, line 5: 👍

mason-fish (approved):
Nice work! There is some overlap with the former `initialViewerSearch` api with what I'm doing in the query version work, but I think it will be easy to reconcile.

src/js/brim/program.test.ts
jameskerr: Previously, we need to have the "ts" field be included in the results to perform infinite scroll. If a query happened to use "cut", it probably cut out the ts field, so we were unable to get the next page. Our solution in this PR is more robust and doesn't care if you cut fields or not. 

The hasAnalytics check only needs to check the AST for a "Summarize" proc. It used to check for "cut" and "sort". It doesn't need to anymore. Hopefully that makes sense.

If the query has analytics, then we limit the results to 2000. If not, we infinite scroll using the head / tail query trick.

PR #2361, opened 2022-05-25
Location Change Fetches Results

jameskerr:
As discussed, this change introduces a loadRoute function which is called whenever this route is freshly activated.

It syncs the editor and it fetches the first page of results.


src/app/features/sidebar/pools-section/pool-item.tsx
```diff
@@ -24,18 +25,22 @@ const PoolItem = ({styles, data, state, handlers}) => {
         handlers.edit()
       },
     },
+    {
+      label: "Get Info",
+      click: () => {
+        dispatch(Tabs.activateUrl(lakePoolPath(pool.id, lakeId)))
+      },
+    },
```
jameskerr, line 33: I also made the clicking of a pool create a draft query with that pool as a pin. Otherwise, you can right click and click "get info" to pull up the pool page.

src/app/query-home/loader.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function fetchData(location) {
+  return (dispatch, getState) => {
+    const key = Results.getKey(getState())
+    const query = Current.getQuery(getState())
+
+    if (key === location.key) return
+
+    startTransition(() => {
```
mason-fish, line 49: 👌

mason-fish (commented):
Since updating and merging my versions PR this now also needs a merge, but before I did that I thought this was looking pretty good!

PR #2362, opened 2022-05-26
Turn on query flow

mason-fish:
Removes non-query-flow code path so the app's query-flow is now the default.


jameskerr (approved):

jameskerr:
Let's do it!

PR #2364, opened 2022-05-26
Query flow fixes part 2

mason-fish:
* Turns off History section in sidebar for some upcoming user testing experiments
* refactors the QueryVersion slice's code to separate files
* moves the right pane's "current" pane view ("detail" or "versions") into tab state
* defaults the right pane to already be open
* refactors a little bit of the BrimQueries class to be a bit simpler
* several null check style typeError protections added, with defaults
* fixes a bug where copying queries would give versions all new timestamps and caused them to lose their original order
* fixes a bug where copying or duplicating queries would not create new versions, so multiple queries would point to the same group of versions (very confusing UI experience...)
* fixes a bug where we were not updating the pool list upon creating the _remote-queries pool for the first time, this then prevented a **second** remote query from being created until you refreshed the page


jameskerr (approved):
Looks good. I think we need a more descriptive name for the paneName state. But merge it and we'll figure it out later.

PR #2365, opened 2022-05-26
More query-flow fixes

mason-fish:
* Remove the "add remote query" from plus menu since it was still using the old modal flow which would cause a breakage.
* Fixed BrimQuery .toString() typeErrors that occured in special situations
* Display errors in results

The error we get when there is no `from <pool> |` from the backend was not getting displayed. Now it does (though we should make this more helpful!)
image

### DIFF
--- a/src/app/query-home/utils/brim-query.ts
+++ b/src/app/query-home/utils/brim-query.ts
@@ -1,5 +1,5 @@
 import {Query} from "src/js/state/Queries/types"
-import {last} from "lodash"
+import {isEmpty, last} from "lodash"
 import {QueryPin, QueryPinInterface} from "../../../js/state/Editor/types"
 import {isNumber} from "lodash"
 import brim from "src/js/brim"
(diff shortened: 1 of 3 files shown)


src/app/query-home/utils/brim-query.ts
```diff
@@ -99,16 +99,19 @@ export class BrimQuery implements Query {
...
-      .filter((p) => !p.disabled)
-      .map<QueryPinInterface>(buildPin)
-      .map((p) => p.toZed())
+    let pinS = []
+    if (!isEmpty(this.current?.pins))
+      pinS = this.current.pins
+        .filter((p) => !p.disabled)
+        .map<QueryPinInterface>(buildPin)
+        .map((p) => p.toZed())
+    let s = pinS
```
jameskerr, line 108: Did this change anything? Doesn't look like it.

jameskerr (approved):

src/app/query-home/utils/brim-query.ts
mason-fish: It's just a little more careful with how it accesses pins by using `isEmpty()`. I a typeError and this fixed it

PR #2380, opened 2022-06-06
Skip Release Notes on First Launch

jameskerr:
If this is the first time ever opening the app, do not display the release notes. Otherwise, only show the release notes if it is the first time using this version.

I create a file called first-run after launching. If that file exists, it's not the first run ever.

First launch after nuking the "run" directory
Screen Shot 2022-06-06 at 2 14 49 PM

Launching again after changing the version in package.json to v0.30.2.

Screen Shot 2022-06-06 at 2 15 12 PM


mason-fish (approved):
👍

PR #2394, opened 2022-06-09
Resize editor

### LINKED ISSUE #2382: Resize Editor

mason-fish:
fixes #2382 

stores the height of the editor in tab state under the existing `editor` slice.

there is currently a bug where we are not properly persisting tabs which needs to be fixed before we can see the app remember that height.


jameskerr (approved):
image

image

This makes sense to me! Let's make it a little larger of a target area. I think 16px and -8px is nice.

Just make sure that the run button can be clicked without the cursor: row-resize appearing. You can either move it up a bit, or make sure it's layered on top of the drag area.

PR #2395, opened 2022-06-14
fix right pane bug

### LINKED ISSUE #2390: Fix right pane error.

mason-fish:
fixes #2390 

We have been storing Date instances in redux. This works fine until you reload the app, triggering the state to be serialized to json and then reloaded. After that the dates have been cast to strings and our components aren't expecting that.

The change here is to always use an iso formatted string for query version timestamps. We can still easily and reliably sort them in zed and js, and then consumers can convert to a Date instance as needed.

Also: I found some duplicate code that should have been removed when we turned the "query-flow" feature flag on


jameskerr:
This feels unfortunate. It might be confusing to have a field called "ts" is actually a string. Can we add a serialize, deserialize layer to the queries at the reducer level?

jameskerr (approved):

PR #2397, opened 2022-06-14
Default query name count func

### LINKED ISSUE #2375: Change Default Query Name to Query #{n+1}

mason-fish:
fixes #2375 

new counter function that determines next default query name's next count based on existing queries


jameskerr (approved):
I love those tests! Thanks!

PR #2404, opened 2022-06-16
Change events request accept header

mason-fish:
Will need to wait until the backend is updated accordingly to confirm, but this should help get our events subscription back in gear. I also updated the e2e test which was falsely passing.


jameskerr (approved):

nwt:
@mason-fish: Sorry for the extremely slow turnaround on this!

I just created brimdata/zed#3964 to fix the bad Content-Type response header bug described in brimdata/zed#3924. That gets the events mechanism in the app working for me, so once it's merged, you should be able to merge your fixes to `packages/e2e-tests/tests/zed-events.spec.ts`.

However, changing the Accept request header value needs to wait for either brimdata/zed#3965 or some other change to `GET /events` content negotiation that will make "Accept: text/event-stream" acceptable.

PR #2415, opened 2022-06-28
migrate old -> new queries

### LINKED ISSUE #2412: Saved Query Migrations
Ensure that old query state gets successfully migrated to the new queries state in Zui v1

mason-fish:
closes #2412 

Also removes the security-centric brimcap lib, which we decided users can choose to re-import from brimcap if they'd like.


jameskerr (approved):
Right on! I see the queries tree gets flattened, then for each one you take the value and create a single queryVersion from it. Then you mutate the query deleting the fields that are irrelevant now. This keeps the structure the same.

PR #2417, opened 2022-06-29
Careful Detail Pane

jameskerr:
- Stop decoding and encoding zed values
- Limit what gets rendered in the detail pane

We don't persist the log details state, but we were encoding the values before they went into the store. This was hurting performance when the zed value was yuge. Thankfully, it's easy to just stop encoding it since nothing depends on it being encoded.

Secondly, the detail pane was rendering the 50,000 item array to a string which also hurt performance. Now if it encounters an array, it displays the array bracks with a "..." and the number of items in the array. Same goes for sets and maps. Records are flattened in the detail view, so they don't appear in the value column.
Screen Shot 2022-06-29 at 10 20 01 AM

To be safe, I also limited the number of rows in the detail view to 500. If it reaches that limit, it says the user they can look in the inspector to see everything. I was testing it with a limit of 3, but it's now 500.

Screen Shot 2022-06-29 at 10 19 15 AM

### DIFF
--- a/src/app/core/format.test.ts
+++ b/src/app/core/format.test.ts
@@ -1,10 +1,10 @@
 import {createData} from "src/test/shared/factories/zed-factory"
 import {zed} from "@brimdata/zealot"
-import {formatPrimitive} from "./format"
+import {formatValue} from "./format"
 
 test("format a long number", () => {
   const num = createData(1394668388559068) as zed.Primitive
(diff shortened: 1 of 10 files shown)


mason-fish (approved):
Right on, this seems like a good fast fix. Agree on embedding the inspector in the future too!

PR #2428, opened 2022-07-11
brim => zui app data migration

### LINKED ISSUE #2409: Prepare Zui Data Migration Script
When Zui launches for the first time ever, write a script to go get the app state and the lake from the Brim directory.
Copy it into the Zui directory, then restart the app.

Test that permissions work.

mason-fish:
closes #2409 

On the first ever run of ZUI, this will look for the Brim app's data and copy it into ZUI. If there is no brim data or for some reason there is already ZUI data, then we do nothing.

To test, use the built release artifacts of this branch [here](https://github.com/brimdata/brim/actions/runs/2652844255).

To simulate a first run, you can remove any ZUI user data you may have (if you've run a ZUI release before) by nuking `~/Library/Application Support/Zui` (on macOS), and then running the app. In addition to seeing your Brim data in ZUI, you can also see the logs for details of what was copied.


src/js/electron/migrateBrimToZui.ts
```diff
@@ -0,0 +1,52 @@
+import {app} from "electron"
+import path from "path"
+import fs from "fs-extra"
+import log from "electron-log"
+
+export default () => {
+  const zuiPath = app.getPath("userData")
+  const zuiAppStatePath = path.join(zuiPath, "appState.json")
+  const zuiAppLakePath = path.join(zuiPath, "lake")
+  const zuiAppDataPath = path.join(zuiPath, "data")
```
jameskerr, line 10: We decided to check if this is zui insiders before running this.

jameskerr (approved):
Awesome, I like the one line check.

PR #2433, opened 2022-07-12
Status Bar & Layout Cleanup

jameskerr:
05DqT3HnZR

While I was doing this, I changed the AppWrapper class to use a css grid layout for the app. I added grid-areas to the main components of the page.

I also gutted the Pane class and renamed it "DraggablePane". 

I deleted some old css and component files.


mason-fish (commented):
I checked this out and the styling looks great to me. Couldn't find any issues with the new grid layout 👍 

I did find 2 other cases we should handle:
* If I issue a search, and then issue another before the first has finished. The status gets updated to "Completed" even though it is still fetching
* If I switch tabs after issuing a search, the completed status gets updated on the new current tab and the original tab remains in the same stale fetching status.

jameskerr:
@mason-fish you found a big bug! Results were always going to the "active" tab instead of the tab specified when the query began. I've fixed it in here.
image

image

image

I've added the number of shapes in the data as well.

packages/zealot/src/query/result-stream.ts
```diff
@@ -87,7 +87,7 @@ export class ResultStream {
           (e instanceof Error && e.message.match(/context canceled/))
         ) {
           this.status = "aborted"
-          resolve()
+          reject(e)
```
jameskerr, line 90: This fixes the first error you found. When aborting, it was still resulting in success. Not anymore.

src/js/state/Tabs/slice.ts
```diff
@@ -85,7 +85,7 @@ const slice = createSlice({
   },
   extraReducers: (builder) => {
     builder.addMatcher(isTabAction, (s, a) => {
-      const id = a.tabId || s.active
+      const id = a.tabId || a.payload?.tabId || s.active
```
jameskerr, line 88: This fixes the second problem you found. Where the results and the stats are updating the wrong tab.

mason-fish (approved):
Awesome!

Issue #2434, opened 2022-07-13
Switching tabs causes a partially-written query to disappear


philrz:
Repro is with Brim commit 1d75d8da.

As shown in the attached video, if I'm in the middle of writing a query, but stop short of running it, then click to another tab and back, the query text I had been writing has disappeared.

https://user-images.githubusercontent.com/5934157/178837615-8e744a69-764d-4651-8cbb-32dea57fb3c2.mp4

I bumped into this as a user when I was trying to check some query results in another window to use when composing my query.

philrz:
Verified in Zui commit a20f709.

As shown int he attached video, now when I click between tabs I no longer lose the query I was in the middle of typing.

https://user-images.githubusercontent.com/5934157/183481837-f05dd363-c4aa-4d2b-92dc-e39f717c0590.mp4

Thanks @jameskerr!

PR #2439, opened 2022-07-14
Session Flow

### LINKED ISSUE #2383: Rename "Versions" to "History"
Use abbreviations for time units. Keep that area small, so that as much of the query is visible.
Design a different selection state for the selected version.
Design a query version row with pins.

jameskerr:
**Main Feature**

This is the first implementation of the Session centered query experience. There is no ticket specifically for this.

**Bug Fixes**
Fixes #2383 

Renamed Versions to History

Fixes: #2434

The problem for #2434 was we were calling the "loadRoute" function every time a tab was activated. In the loadRoute function, we sync the editor state with the url. However, we should only do this the first time a new url is opened in a tab, not every time the tab is activated.

The solution is to keep a "lastLocationKey" property on the tab. Then when a tab is activated, it checks to see if the lastLocationKey is the same as the current location key. If it is, do nothing. Otherwise, load the route.


mason-fish:
Mason's To Do List:

- [ ] Decide on behavior of a session for query deletion (both tab history and session history will be affected). Implement solution.
- [ ] Refactor history section's logic for transforming session history to the props for each history entry component. It works now but should at least be moved to it's own thunk.

mason-fish (approved):

jameskerr:
The unfinished first-class sessions behavior is documented here: https://github.com/brimdata/brim/issues/2467

src/app/query-home/results/index.tsx
```diff
@@ -5,33 +5,31 @@ import {useResultsData} from "./data-hook"
...
+const BG = styled.div`
+  display: flex;
+  flex-direction: column;
+  border: none;
+  position: relative;
+  flex: 1;
+`
+
+const Body = styled.div`
+  flex: 1;
```
philrz, line 22: Adding a comment so I can link to it from https://github.com/brimdata/brim/pull/2513.

PR #2445, opened 2022-07-18
upgrade Zed to 834a63338e3864510538f542c21000c0033b838c

### LINKED ISSUE #2444: Explicitly request control messages when needed
The "Advance Zed" test has start failing (such as [here](https://github.com/brimdata/brim/runs/7383439851?check_suite_focus=true)), correlated with the merge of https://github.com/brimdata/zed/pull/4001. Looking at the failure, it seems pretty clear what's wrong. This code:

https://github.com/brimdata/brim/blob/0d3746775e5040f74595815fcc2d8c0cb9d3c68c/packages/zealot/src/query/result-stream.ts#L100-L123

...shows the app is still expecting to see the control messages in the query response, but https://github.com/brimdata/zed/pull/4001 flipped the logic such that a plain `/query` now returns only the query response. Therefore it seems this code would need to tack on the `?ctrl=T` option so it can get back the control messages.

On a related topic, I noticed there's a boolean `controlMessages` that appears several places in the code. Since the backend option for control messages went away and then came back with inverted logic, I assume these need to be attended to.

nwt:
This updates Zealot for brimdata/zed#4001 by adding the ctrl URL query
parameter to POST /query requests.

Closes #2444.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,7 +175,7 @@
     "use-resize-observer": "^8.0.0",
     "valid-url": "^1.0.9",
     "web-file-polyfill": "^1.0.4",
-    "zed": "brimdata/zed#5ee33b7afe4387f1a670eb9395d752a3a9a24fa0"
+    "zed": "brimdata/zed#834a63338e3864510538f542c21000c0033b838c"
   },
   "peerDependencies": {
     "react": "^18.0.0",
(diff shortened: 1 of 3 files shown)


jameskerr (approved):
Makes sense

PR #2455, opened 2022-07-20
Correlations

### LINKED ISSUE #2443: connCorrelation query misses Suricata alerts due to timestamps (regression at b819d7e)
Repro is possible with current Brim commit b282d7a, though a binary search has confirmed that the problem was a regression starting at commit b819d7e.

To repro, drag the common "all.pcap" test data into the app and query for `"1:IK50rOTNQcdxeBERZDmdl/D3z7c="`, which is a Community ID for a flow that's known to be associated with many Zeek events and Suricata alerts. The main results window does indeed show the many `alert` events alongside the Zeek `conn` log, but the Correlation visualization in the side Detail panel doesn't show any `alert` tiles.

image

By sniffing in Wireshark I could see the query being crafted by https://github.com/brimdata/brim/blob/main/src/js/searches/programs.ts and sent to the backend Zed API looks like:

```
{"query":"from 0x0f5fa8605ca618bf66b91542519d763d534ad2ac | uid==\"CKViDSjhtOHgqDUN1\" or \"CKViDSjhtOHgqDUN1\" in conn_uids or \"CKViDSjhtOHgqDUN1\" in uids or referenced_file.uid==\"CKViDSjhtOHgqDUN1\" or (community_id == \"1:IK50rOTNQcdxeBERZDmdl/D3z7c=\" and ts >= 1428413683.772 and ts < 1428413779.68) | head 100"}
```

The problem is with the `ts` timestamp comparisons. Brim commit b819d7e marked the point where it started using a Zed backend that had the change from https://github.com/brimdata/zed/pull/3809, which was when Zed `time` values that are converted from numbers started being interpreted as epoch nanoseconds rather than being scaled from epoch seconds to nanoseconds automatically. So to make this work, one could multiply the numeric times by 1-billion:

```
$ zed query -z 'from all.zng | uid=="CKViDSjhtOHgqDUN1" or "CKViDSjhtOHgqDUN1" in conn_uids or "CKViDSjhtOHgqDUN1" in uids or referenced_file.uid=="CKViDSjhtOHgqDUN1" or (community_id == "1:IK50rOTNQcdxeBERZDmdl/D3z7c=" and ts >= 1.428413683772e18 and ts < 1.42841377968e18) | has(event_type) | count()'
{count:20(uint64)}
```

Or express them as ISO8601 timestamps:

```
$ zed query -z 'from all.zng | uid=="CKViDSjhtOHgqDUN1" or "CKViDSjhtOHgqDUN1" in conn_uids or "CKViDSjhtOHgqDUN1" in uids or referenced_file.uid=="CKViDSjhtOHgqDUN1" or (community_id == "1:IK50rOTNQcdxeBERZDmdl/D3z7c=" and ts >= 2015-04-07T13:34:43.772Z and ts < 2015-04-07T13:36:19.68Z) | has(event_type) | count()'
{count:20(uint64)}
```

Related question: Would there be an easy way to audit the code and find other places where the app might still be generating queries with time values that depend on the old scaling behavior? We've been bitten by this Zed change in a couple places now and it would be a drag to keep having to find them all whack-a-mole style.

jameskerr:
fixes #2443 

This is exciting. We've heard from users who want to add custom correlation queries to brim. Before this PR, each correlation (zeek, suricata, md5) was implemented slightly differently and was hard-coded. This unifies the implementation and takes a huge step towards third-party correlations.

A significant change was extending the "Results" state slice. Now it can hold results for any number of queries, not just the main query. You provide an "id" whenever you set or select results.

I also created a simple `api.query` method on our BrimApi class. It cleans up after itself so that you don't have to think about it.

Finally, I've made a `correlations/run()` thunk that uses the above two features. Each correlation is given an id, the query is issued, and the results are saved to the Results state slice under that id.

Each correlation React component looks up the results by the relevant id. 

The last piece of the puzzle is converting the brimcap-plugin to a new system. It currently relies on a thunk `fetchCorrelation` that needs to be deleted. We need to decide if we should issue a new query, or wait for the zeek query to find the conn record.

# UPDATE:

1. All of the correlations now live in the plugins folder. 🎉
2. The brimcap plugin has been updated to not rely on the fetchCorrelation.

There is now a new api section called `correlations`.

It works like this:

```ts
// myplugin.ts

const md5Correlation: Correlation = {
  id: "zui-zeek/md5-correlation",
  when: hasMd5,
  query: (api) => {
    const md5 = api.current.value.get("md5").toJS()
    return zql`
        from ${api.current.poolName} 
        | md5==${md5} 
        | count() by md5 
        | sort -r 
        | head 5`
  },
}

api.correlations.add(md5Correlation)
```

Whenever the current value (selected) changes, we call `dispatch(runCorrelations())`. It will grab all the correlation objects off the api, loop through them, check if they should run with the `when` property, then run the query returned from the `query` property.

The results get saved to the results slice under the `id` correlation property.

## Next Steps

To allow third-party javascript to add their own correlations, we only need to do two things:
1. Pick a "plugins folder" for the app to look for this code
2. Update the UI to loop through the configured correlation configurations and create the components automatically based off the configuration. We'll need to add more properties to the correlation configs like, title, caption, type, and so on. At first we'll just have two types: table and timeline. Each one will have its own configuration properties.


src/js/state/Results/reducer.ts
```diff
@@ -1,81 +1,71 @@
...
-    error: {
-      prepare: (error: any, tabId?: string) => ({
-        payload: {error, tabId},
-      }),
-      reducer: (s, a: Pay<{error: any}>) => {
-        s.error = a.payload.error
-        s.status = "ERROR"
-      },
+    error: (s, a: Pay<{id: string; error: any; tabId: string}>) => {
+      console.log(a)
```
mason-fish, line 65: still want this?

src/plugins/brimcap/brimcap-plugin.ts
```diff
@@ -83,12 +84,15 @@ export default class BrimcapPlugin {
     this.updateSuricata()
   }
 
-  private async tryConn(detail: zed.Record, eventId: string) {
-    // TODO: dispatch is only temporarily public to plugins, so this won't always be needed
```
mason-fish, line 87: So happy this is now handled by this PR :)

src/js/api/index.ts
```diff
@@ -1,125 +1,57 @@
...
 
-  /*
-  TODO: store is made public only to make the initial plugin architecture
-    migration easier, it will eventually be private
-   */
-  public dispatch: AppDispatch
-  public getState: () => State
-
-  constructor() {}
+  private dispatch: AppDispatch
```
mason-fish, line 34: 👏

mason-fish (approved):
This is awesome, nice work!! 🔥

Issue #2457, opened 2022-07-22
Scrollbars appearing/disappearing when they shouldn't in left panel


philrz:
A community member first made us aware of this issue. Repro is with Brim commit ab46d88. Multiple symptoms are described in this issue on the assumption they have related root causes & fixes, but we can always break them off into more issues if they need to be handled separately.

Repro steps:
1. Have some test data with a long filename that will turn into a long pool name when imported
2. Import the test data repeatedly to create a screen full of pool names in the left panel
3. Begin resizing the window horizontally and vertically

As shown in the attached video, this reveals problems with both the scrollbars in the left panel.

1. The horizontal scrollbar is not visible at the beginning, but it probably should be because the pool names are wide enough that a user might want to "scroll right" to see their names.
2. When doing the horizontal window resize, we see the horizontal scrollbar appear and disappear, but ultimately it never sticks around.
3. When doing the vertical window resize, we see a redundant "inner scrollbar" appear and disappear repeatedly.

https://user-images.githubusercontent.com/5934157/180478135-c255eb80-05db-4496-8a52-970ec543cffa.mp4

In fact, these were second-order issues that were uncovered while trying to reproduce the user's original reported scrollbar issue, which I've only been able to reproduce on Windows. In this case I'm using Zui Insiders 0.30.1-26 to repro.

The repro steps 1 & 2 are the same. In this case however, a necessary ingredient is to have a secondary display, which in my case is the built-in laptop display. This secondary display is set to scale to 150%. A scaled up display also seems like a necessary ingredient for this repro.

![image](https://user-images.githubusercontent.com/5934157/180479945-38ca0453-7486-4a92-be4a-d0393aceb709.png)

I start with my Zui window like this on the main display.

![image](https://user-images.githubusercontent.com/5934157/180480237-1a2ea448-b428-4d75-a80a-e45baf2ce22c.png)

I then unplug the laptop, causing all open windows to move to the secondary/scaled display. At this point, the Zui window on the secondary display shows the redundant inner scrollbar. Unlike the repro on macOS shown in the video above, in this situation both scrollbars actually stick around,

![image](https://user-images.githubusercontent.com/5934157/180480495-dc14f4d7-43d7-49c4-87d3-e3314bfbd29c.png)

philrz:
Verified in Brim commit 3e0bfbc.

[...]

PR #2458, opened 2022-07-22
Fix Double Scroll Bar

jameskerr:
Fixes: #2457

Now there will be no flashing horizontal scrollbar, and no double vertical scroll bar as you resize or scale.

I made the choice for the sidebar not to be horizontally scrollable. That was a design choice that follows the pattern of other apps with sidebars. For example, here's a screenshot from slack with a long channel name. It also doesn't scroll horizontally.

Screen Shot 2022-07-22 at 10 38 04 AM 

I also tweeked the size of the drag anchors for the panes. And gave the section contents a min-width of 0 so that they will shrink as the window gets smaller.

I'd upload a video but I don't have good internet.

### DIFF
--- a/src/app/features/sidebar/common.tsx
+++ b/src/app/features/sidebar/common.tsx
@@ -13,9 +13,9 @@ export const StyledSection = styled.section`
 
 export const SectionContents = styled.div`
   flex: 1;
-  overflow-y: auto;
   position: relative;
   padding-top: 6px;
+  min-height: 0;
 
   &::before,
   &::after {
--- a/src/app/features/sidebar/item.tsx
+++ b/src/app/features/sidebar/item.tsx
@@ -205,7 +205,12 @@ type ItemProps = {
 
 export function Item(props: ItemProps) {
   return (
-    <Container tabIndex={0} style={props.styles.row} ref={props.innerRef}>
+    <Container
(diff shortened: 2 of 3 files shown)


philrz:
Thanks @jameskerr. Regarding the lack of horizontal scrollbar, I'm remembering that choice now. In contemplating the comparison to Slack, one small difference is that on Slack there's a hover tooltip that reveals the full channel name, whereas we don't have that yet for the long pool names in the left side panel. So right now the only ways I can see to observe the full channel name are to make one of the right-click selections (e.g., Get Info, or select Delete/Rename and then cancel out), or left-click to select the pool so I can see the "FROM" text in the main search area. All of these seem kind of heavyweight. This makes me think the tooltip might be a nice thing to add. If you agree and it justifies its own issue, I'd be happy to open... or if it's easy to add here, that's great too. :smile:

mason-fish (approved):

jameskerr:
@philrz done! 
image

jameskerr:
This is less sophisticated than Slack. I see that they sense if there is an overflow, then add the toolbar. Fancy. This just adds title text to all items.

philrz:
Looks great as is. Thanks @jameskerr!

PR #2479, opened 2022-08-10
Main Function Tech Debt Paydown

jameskerr:
As I was working on fixing the log detail window bug, I paid down some tech debt. This is a payment PR. 

## Main Process Lifecycle

The main processes' "main" function was big and messy with many dependencies and concerns. I've organized it into three steps:

1. Before boot
2. Boot
3. After boot

I've moved code that belongs in each step into the respective file. The main function now basically looks like this.
```ts
export async function main(args: Partial<MainArgs> = {}) {
  const err = await beforeBoot()
  if (err) return log.error(err)

  const brim = await boot(args)
  await afterBoot(brim)
  return brim
}

main()
```
 ## Initializers

Most of the code in the "boot" phase has been moved into an "initializers" directory. There's a function in the boot phase that automatically requires each file in that directory and runs it, passing in the BrimMain instance. This was heavily inspired by the way ruby on rails did things.

It's very nice. Now you can easily add more things that need to happen in the main boot phase, but you don't need to change other code, just add a file. And it's easy to split up the different types of things that need to be done. The current list of initializers now is:

```
src/js/electron/initializers
├── app-menu.ts
├── auto-update.ts
├── couriers.ts
├── custom-protocol.ts
├── legacy-ipc.ts
├── log-filters.ts
├── secure-web-contents.ts
├── shortcuts.ts
├── user-tasks.ts
└── window-events.ts
```

Here's an example of an initializer:

```ts
// custom-protocol.ts

export function initialize(main: BrimMain) {
  const brimCustomProtocol = "brim"
  app.setAsDefaultProtocolClient(brimCustomProtocol)
  app.on("second-instance", (e, argv) => {
    for (let arg of argv) {
      // handle custom protocol url handling for windows here
      if (arg.startsWith(`${brimCustomProtocol}://`)) return main.openUrl(arg)
    }
  })
}

```

I've also added a [log filter dep](https://github.com/sindresorhus/filter-console) that allows us to silence certain log messages from electron. I silenced the "ExtensionLoadWarning"s from the devtools that we cannot fix. I also rearranged the app paths log messages.

```ts
// log-filters.ts

export async function initialize() {
  // Since this is an ESM only module, we've got to do this...
  const filter = (await import("filter-console")).default

  filter([
    // These are annoying warnings about the devtools that we can't remove.
    /ExtensionLoadWarning/,
  ])
}
```

Now when you boot the app you get: 

CleanShot 2022-08-10 at 11 18 20@2x


mason-fish (approved):
This mostly looks really great to me!

The only reservation I have is that I don't like the pattern of implicitly running files just because they exist in a directory. I know we are already doing this for plugins but I think it makes more sense in that situation since we want to support running 3rd party code in the future. My 2 primary concerns with this are that the order of execution is not controlled, and it's just a little harder to trace how a given initializer gets run since we won't see an explicit call.

I know it is slightly more verbose, but I would personally prefer if we explicitly import each initializer into an array (where we can control the order) then loop-execute through that instead.

I'll approve anyway though because this is an awesome refactor and I realize my opinion falls on one side of a famously controversial paradigm (convention vs configuration) 😛

philrz:
I'm in no position to comment on the JavaScript-specific aspects of this, but I'll note that @mason-fish's concern about execution order has been addressed in other tools (e.g., Unix init tools that run scripts at boot time) by prefixing each file with some kind of numeric prefix. Not sure if that would be desirable here. But at a high level I do like the concept of obvious run order.

jameskerr:
Thanks for the input and I hear your concern @mason-fish. It seems to me that if code is going to be in an initializer, it's needs to not care whether other initializers have run or not. It looks like all our initializers are not order-dependent. 

On the other hand, the "beforeBoot" function contains only explicitly named functions in a specific order because it does matter there.

As for the magic folders, I must say that I am a big fan of the pattern. Whenever I can extend a system simply by adding a file in a special place, it feels like a win. I've done the dance many times of adding a file, then needing to require it somewhere else. It's fine, but it's a ceremony none the less (unless order matters like you pointed out).

Let's see how this holds up with time!

Issue #2480, opened 2022-08-10
Zui-fy the Brim wiki


philrz:
At the same time that we make the transition of the app name from "Brim" to "Zui", the wiki currently at https://github.com/brimdata/brim/wiki needs to mention the transition and also start referring to the app by its new name.

A stretch goal would be to simultaneously transition the wiki content into a separate docs site like we've done with https://zed.brimdata.io/. Once we've done this, we'll need to figure out how it fits in with what's already at that site, e.g., will there be a wholly separate page for https://zui.brimdata.io/ and a link between the two, or both land at the same page but with links to both sets of docs? Current consensus at a recent team meeting was something more like the latter, but final decision TBD.

philrz:
Note to self: When docs are at their post-GitHub-wiki home, link to them from the Welcome page (#2387). Also the link in #1497.

philrz:
Along with today's release of [Zui v1.0.0](https://github.com/brimdata/zui/releases/tag/v1.0.0), the former Brim wiki that's now at https://github.com/brimdata/zui/wiki has been replaced with a simple pointer to the new site https://zui.brimdata.io/docs. Regarding the other follow-on items mentioned in comments above, there are open issues for #1497 and #2704 that will now be pursued, and the New Issue templates also link to the new docs site.

Issue #1263, opened 2020-12-08
"Supported Platforms" update for the Suricata era


philrz:
Up until now, our [Supported Platforms](https://github.com/brimsec/brim/wiki/Supported-Platforms) statement has been primarily expressed in terms of where we run CI runs as well as the support statements from Go/Electron. However, the work on Suricata reminded us how there can be additional sensitivities in terms of shared libraries that may or may not be present in platforms and hence will determine whether Brim can install and operate out-of-the-box on them.

Now that the Suricata support is pretty well established, this issue will be used to capture a point-in-time snapshot of where we stand on the various platforms. The information can then be used as the basis for a revised support statement we can use going forward.

The approach I'm taking is largely a binary search: Establish the oldest & newest revisions in a given platform that seem to pass a smoke test, which can be combined with our CI statement.

philrz:
Smoke testing has been performed with the Brim `rc-v0.21.0-suricatav5.0.3-brimpre2` draft release candidate. Each smoke test consists of:

1. Install the base OS/distro while accepting the defaults on any config options. No additional software is added.
2. Install the Brim app using standard package install procedure for the OS/distro (e.g. double-clicking in some "file explorer" utility).
3. Import a [test pcap](https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz) and confirm the embedded Zeek and Suricata both generate events (`fuse | count() by _path,event_type`)

# Windows

Getting easy access to common desktop versions has been a challenge, as Google Cloud only offers Server versions. However, as part of this exercise I found https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/ where Microsoft makes dev VMs available from as far back as Windows 7 in the name of helping test IE compatibility. Therefore I used these in smoke testing.

## Windows 7: Fail

We'd had anecdotal evidence that some users had limited success running Brim on this EOL platform, though we also had some bugs reported that we couldn't really pursue due to lack of access to the platform. However, smoke testing with the `rc-v0.21.0-suricatav5.0.3-brimpre2` artifact it fails out-of-the-box, so it seems appropriate for us to claim no support for Windows 7 at this time.

![image](https://user-images.githubusercontent.com/5934157/101509762-f87a2c80-392d-11eb-848f-b05db6409250.png)

## Windows 8.1: Pass

[...]

PR #2481, opened 2022-08-10
Use macos-11 runners and update Suported Platforms article

### LINKED ISSUE #2463: Update build/CI automation to use macos-11
The following informational message was produced during a recent Actions run:

> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on August 30th, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

In reaction to this, I plan to:

1. Create test builds using the macos-11 Actions Runners
2. Smoke test it on older OS X releases
3. Make any necessary updates to our support statement at https://github.com/brimdata/brim/wiki/Supported-Platforms
4. Advance our Actions Workflow YAML configs to start using the macos-11 Actions Runners going forward

philrz:
As noted in #2463, GitHub is deprecating the `macos-10.15` platform in their hosted Actions Runners. @mason-fish recently made some Zui test binaries that were build with `macos-11` (artifacts are [here](https://github.com/brimdata/brim/actions/runs/2827034553), made from [this branch](https://github.com/brimdata/brim/compare/build/test-zui)), so this gave me an opportunity to smoke test on an older macOS release to check if they still run ok and, if not, update our support statement. Rather than run through the exhaustive set of tests described in #1263, I opted to just test on Mojave 10.14 since that was the oldest release Brim worked on the last time I went through this exercise, and assume if that's ok then newer releases would still be ok too. Thankfully the app still installed and ran ok on that.

Since it's been a couple years since the last time we did these tests, I went ahead and also smoke tested again on Windows 8.1, CentOS 8 1905, Debian 10.0.0, and Fedora 28, as these were also "oldest working" versions in the last exercise, and I figured we might as well make sure that none of the minor changes on the Actions Runners we we use might have introduced new problems. Thankfully these tested out ok as well.

Therefore, this PR advances the macOS platform to `macos-11` and updates the [Supported Platforms](https://github.com/brimdata/brim/wiki/Supported-Platforms) wiki article to reflect this. While doing this, I recognized that the "Brim wiki" is due for an overall transition to being the "Zui wiki" (and eventually a "Zui docs site" like we did for the Zed docs). But rather than just hit this one article or change everything over now before Zui is ready for GA, I've opened #2480 as a reminder to do this later, and for now this article is still Brim-centric with just a brief reference to Zui.

FIxes #2463

### DIFF
--- a/docs/Supported-Platforms.md
+++ b/docs/Supported-Platforms.md
@@ -23,7 +23,7 @@ recommendations on which to run Brim:
    * Windows 10 or newer
    * Windows Server 2019 or newer
 * macOS
-   * macOS Catalina 10.15.7 or newer (see [below](#hardware) for hardware considerations)
+   * macOS Big Sur 11.6.8 or newer (see [below](#hardware) for hardware considerations)
 * Linux
   * Ubuntu 18.04 or newer
   * CentOS 8.0 1905 or newer
@@ -55,9 +55,9 @@ on releases older than Windows 8.1.
 
 ### Software
 
-Brim's [test automation](#automated-testing) runs on Catalina 10.15 and
+Brim's [test automation](#automated-testing) runs on Big Sur 11 and
 therefore this is the macOS version on which we are best able to ensure quality
-and prevent regressions. Several Brim developers also run macOS Monterey 12.3
+and prevent regressions. Several Brim developers also run macOS Monterey 12.5
 and regularly perform ad hoc testing with it to reproduce reported issues.
 
 Basic [smoke testing](#smoke-testing) has also validated that Brim appears to
@@ -69,7 +69,7 @@ recommend attempting to run Brim on macOS releases older than macOS Mojave
 ### Hardware
 
 The build procedure for Brim's macOS releases creates binaries intended to
-run on the Intel-based Mac hardware that make up the majority of Macs in
+run on the Intel-based Mac hardware that make up many of the Macs in
 use today. Brim releases are not yet available that are built specifically for
 the recently-released [M1-based hardware](https://en.wikipedia.org/wiki/Apple_M1).
 However, Apple's [Rosetta 2](https://support.apple.com/en-us/HT211861) makes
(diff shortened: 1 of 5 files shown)


docs/Supported-Platforms.md
```diff
@@ -23,7 +23,7 @@ recommendations on which to run Brim:
    * Windows 10 or newer
    * Windows Server 2019 or newer
 * macOS
-   * macOS Catalina 10.15.7 or newer (see [below](#hardware) for hardware considerations)
+   * macOS Big Sur 11.6.8 or newer (see [below](#hardware) for hardware considerations)
```
philrz, line 26: FYI, while GItHub just claims "macOS 11" in their docs, I grabbed the full version string by executing `sw_vers` on a `macos-11` runner. I'm sure this will be subject to change as Apple keeps releasing minor updates.

docs/Supported-Platforms.md
```diff
@@ -105,7 +105,7 @@ Basic [smoke testing](#smoke-testing) has also validated the _oldest_
 release on which Brim appeared to work for each common distribution, as
 follows:
 
-* Ubuntu 10.04
+* Ubuntu 18.04
```
philrz, line 108: I only just noticed this dumb typo that I've fixed here. The results at #1263 definitely don't mention testing as far back as Ubuntu 10, and the OS list at the top of the article always showed the correct version 18.04. Therefore I imagine I just made a typo back in the day, perhaps due to having "Debian 10" on the brain.

docs/Supported-Platforms.md
```diff
@@ -128,8 +128,8 @@ strongly about trying to get Brim running on a [non-recommended platform](#non-r
 ## Development Tools
 
 There are two primary development tools on which Brim depends:
-[Electron](https://www.electronjs.org/docs/tutorial/support#supported-platforms)
-and [Go](https://golang.org/doc/install#requirements). Their support
+[Electron](https://www.electronjs.org/docs/latest/development/README)
```
philrz, line 131: Interestingly, Electron's content formerly at that "supported platforms" link is not only gone, but I can't for the life of me find a new page anywhere that divulges the same information. Their redirect page _claims_ that README link is the new place, but I don't see the detail there or anywhere below that. @jameskerr looked to and he can't find anything either. So I'm just pointing where Electron suggested and letting it be vague. 🤷

jameskerr (approved):
Looks good to me. Thank you @philrz for performing the tests!

PR #2494, opened 2022-08-17
Fix Detail Window

jameskerr:
The Log Detail window now works as intended. Double clicking a log in the results, or clicking the "open in new window" button in the detail right pane, will open the log detail window.

We used to serialize all of the state in the search window, then send it to the log detail window. Now we only send over the encoded zed value, and the url. 

When the detail window boots up, it will set the url and value, and the window will behave properly.

Note: I had to make the sessionQueries state slice a global state slice. The log detail window needs access to the query that it came from. It uses the url to find it. The session queries need to be available outside of their originating window, so I added a "$" to the reducer and we are good to go.

I also created a concept called Operation and SpecialOperation.

An Operation is an action performed on the main process and is requested through ipc from the client. Using electron's ipc apis does not give any type safety to the objects. This Operation wrapper gives us that and autocomplete.

```ts
export const greeting = createOperation(
  "greeting", 
  (main, event, arg: string) => {
    // the main process context is available as the first arg
    // the ipc event is second
    // and the argument that the renderer will send is third.
    // Typescript will infer the "string" type provided above and require that 
    // when invoking it in the renderer
    return `Hello, ${arg}!`
  }
)

// in the main process
greeting.listen(main)

// in the renderer
greeting.invoke("James") // Typescript will require a string here.
```

And finally, a special operation is one that can return one-off values when a condition is met. We do this frequently in the app. For example, a window boots up, then asks the main process for some data. The main process will kick off the new window, then it will wait for that window to request it's specific data. Or the main process ask a window for something and needs to wait for it to respond.

```
const numberAssigner = createSpecialOperation<number, id>("numberAssigner")

// in the main process
const win = createNewWindow()
numberAssigner.return(24).when(id => id === win.id)

// in the renderer
numberAssgner.invoke(thisWindowId)
```

I'd love feedback on the api design and the names of the functions/methods/classes.
```

Also Mason, I've created another magic folder. Any exported operations that live in the electron/ops folder will automatically have ".listen(main)" called on them. See `electron/run-op-listeners.ts`. 😅

### CODE REVIEW

File: src/js/state/createGlobalStore.ts [hunk 1]
```diff
@@ -1,9 +1,10 @@
 import {configureStore} from "@reduxjs/toolkit"
-import globalReducer, {GlobalState} from "./globalReducer"
+import rootReducer from "./rootReducer"
+import {State} from "./types"
 
-export default function (initState: GlobalState | undefined) {
+export default function (initState: Partial<State> | undefined) {
   return configureStore({
-    reducer: globalReducer,
+    reducer: rootReducer,
```
jameskerr, line 7: I realized that this doesn't need to be a different reducer. It's just the same reducer but we only use some of it in the global store.

scripts/start.js
```diff
@@ -1,7 +1,9 @@
 const sub = require("./util/sub")
 const log = require("./util/log")
+const fs = require("fs-extra")
 
 async function start() {
+  fs.removeSync("dist")
```
mason-fish, line 6: 👍

mason-fish (approved):
Right on!

PR #2495, opened 2022-08-18
Ingest flow integration tests

mason-fish:
Brings back the ingest flow integration tests. While I've modified them to be more stable, the plan moving forward is still to run these locally as a part of the release process and not in CI's inconsistent environment.

### DIFF
--- a/src/app/features/sidebar/plus-button.tsx
+++ b/src/app/features/sidebar/plus-button.tsx
@@ -72,7 +72,7 @@ export default function PlusButton() {
   }
   return (
     <>
-      <Button onClick={() => onClick()}>
+      <Button role="create" onClick={() => onClick()}>
         <Icon name="plus" size={18} fill="white" />
       </Button>
       <input
--- a/src/app/features/sidebar/plus-button.tsx
+++ b/src/app/features/sidebar/plus-button.tsx
@@ -72,7 +72,7 @@ export default function PlusButton() {
   }
   return (
     <>
-      <Button role="create" onClick={() => onClick()}>
+      <Button aria-label="create" onClick={() => onClick()}>
         <Icon name="plus" size={18} fill="white" />
       </Button>
       <input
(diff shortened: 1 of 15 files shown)


src/app/features/sidebar/plus-button.tsx
```diff
@@ -72,7 +72,7 @@ export default function PlusButton() {
   }
   return (
     <>
-      <Button onClick={() => onClick()}>
+      <Button role="create" onClick={() => onClick()}>
```
jameskerr, line 75: I don't think "create" is a valid ARIA Role. Is it?  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

You can do aria-label="create", that might be better.

src/app/query-home/search-area/editor/query-editor.tsx
```diff
@@ -34,7 +34,14 @@ const QueryEditor = ({value, disabled}: Props) => {
   }
   const ref = useEditorView({value, disabled, onChange})
 
-  return <EditorWrap isDisabled={disabled} ref={ref} onKeyDown={onKeyDown} />
+  return (
+    <EditorWrap
+      role="editor"
```
jameskerr, line 39: Looks like most of the roles specified should actually be aria-labels:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label.

src/js/components/BrimToaster.tsx
```diff
@@ -7,6 +7,7 @@ const BrimToaster = () => {
     <Toaster
       position="bottom-right"
       toastOptions={{
+        role: "status",
```
jameskerr, line 10: This one is legit: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role

jameskerr (commented):
Would you mind changing some of the roles to aria-labels before merging?

jameskerr:
How do I run these locally?

jameskerr:
It hangs here for me.

CleanShot 2022-08-23 at 11 57 02@2x

jameskerr:
CleanShot 2022-08-23 at 12 13 50@2x
Noice!

jameskerr (approved):

PR #2496, opened 2022-08-19
Autosave

### LINKED ISSUE #2039: Abrupt app kill causes loss of unsaved state (need autosave)
While testing with recent Brim (commit 00ad8da in Dev mode at the moment), I happened to stop the app by hitting Control-C in the window I launched it from _without_ having first politely closed it via Command-Q first in the app window. I noticed that this caused the loss of state changes that had accumulated in the app since I'd launched it, including connection config to a remote lake, imported queries, renamed entires in the Query Library, and query History entries. I found the same loss happened when running Brim GA `v0.28.0` if I killed it via `kill -9`.

I bounced this finding off @mason-fish and he confirmed this is expected behavior at the moment:

> The connection settings along with the rest of the state are written to file when you close the app the regular way (command-q, close from dock, or close through top menu). Ctrl-c or any other signal-based process ending will result in unsaved state.

While I respect that abrupt process deaths like `kill -9` should be very rare nowadays, in this era when common tools like Google Docs auto-save, I suspect users would take for granted that the time window for such state loss should be much narrower. Not sure if there's a big performance hit to doing state saves for every little operation, but perhaps there's a way to do it periodically, e.g., every X seconds and/or after Y changes have been made without a state save?

jameskerr:
## IPC

Every IPC call in the app has been converted to an operation. I've deleted all other ways we do ipc.

```
src/js/electron/ops
├── autosave-op.ts
├── close-window-op.ts
├── feature-flags-op.ts
├── get-app-meta-op.ts
├── get-global-state-op.ts
├── get-window-state-op.ts
├── global-dispatch-op.ts
├── main-args-op.ts
├── move-to-current-display-op.ts
├── open-about-window-op.ts
├── open-detail-window-op.ts
├── open-directory-op.ts
├── secrets.ts
├── show-preferences-op.ts
├── show-release-notes-op.ts
└── show-save-dialog-op.ts

0 directories, 16 files
```

## Window Manager Class

The window manager class has been refactored and now has an outline like this:

CleanShot 2022-08-23 at 16 23 55@2x

## BrowserWindow classes

I've made a class for each of our 4 window types (About, Search, Detail, and Hidden). They all inherit for ZuiWindow which contains all the common code. This is how the DetailWindow class looks:

```ts
import {WindowName} from "../windows/types"

export class DetailWindow extends ZuiWindow {
  name: WindowName = "detail"
  persistable: false
  options = {
    resizable: true,
    width: 680,
    height: 480,
    minWidth: 500,
    webPreferences: {
      nodeIntegration: true,
      contextIsolation: false,
    },
  }

  beforeLoad() {
    this.ref.setMenu(null)
  }
}
```

## Autosave

And finally autosave. The renderer will send a message to the main process everytime it's state changes. The main process will update that state in the window manager, the save everything to the appState. The functions are all throttled to prevent super fast updates from crashing things.

```ts
const saveSession = throttle((main) => {
  main.saveSession()
  log.debug("Session Autosaved")
}, 500)

export const autosaveOp = createOperation(
  "windows.autosave",
  async (main, e, args: {windowId: string; windowState: State}) => {
    main.windows.update(args.windowId, args.windowState)
    saveSession(main)
  }
```

## Logging

If you run the app with `--verbose`, you'll get more the debug logs about the ipc messages. Otherwise the app will only output info level logs.

Test these features as I've moved code around.

- [x] autosave-op.ts
- [x] close-window-op.ts
- [x] feature-flags-op.ts
- [x] get-app-meta-op.ts
- [x] get-global-state-op.ts
- [x] get-window-state-op.ts
- [x] global-dispatch-op.ts
- [x] main-args-op.ts
- [x] move-to-current-display-op.ts
- [x] open-about-window-op.ts
- [x] open-detail-window-op.ts
- [x] open-directory-op.ts
- [x] secrets.ts
- [x] show-preferences-op.ts
- [x] show-release-notes-op.ts
- [x] show-save-dialog-op.ts

Fixes #2039


jameskerr (commented):
I've self reviewed this and it appears good to go.

philrz:
At @jameskerr's suggestion, I gave this branch a test and the fix looks good. I followed the repro steps from #2039, which is to say that I started the app in Dev mode and made some changes like connecting to a remote lake, imported some queries, renamed entires in the Query Library, and populated some query History entries. Abruptly killing my app window via Control-C always makes those state changes get lost, but on this branch they were all still intact upon relaunch So it's a 👍 for me!

mason-fish (approved):
Awesome work, and what a nice feature to finally have! 👏

PR #2501, opened 2022-08-26
Export e2e test

### LINKED ISSUE #2432: Test coverage for Export
In a recent team meeting it was noted that we'd benefit from having an automated test that catches the kinds of Export failures identified in #2426. Note that #2426 actually describes issues with multiple Export formats that had different root causes, so a test that Exports in all of the available formats would seem ideal.

mason-fish:
fixed #2432 

I was flirting with the idea of switching our export flow to use an html file input so that we can use the same pattern for testing save dialogs as we do for the ingest flow (where Playwright has visibility and control of the "filechooser" event), but I ended up finding a very simple way to mock it within playwright instead. It's worth taking note of this particular pattern because it can be used for other electron features that aren't supported by playwright yet:

```js
      app.brim.evaluate(async ({dialog}, filePath) => {
        dialog.showSaveDialog = () =>
          Promise.resolve({canceled: false, filePath})
      }, file)
```

The tests go through all 4 currently supported export formats and verifies success by checking that the size of the output file is correct, which seemed good enough for me.

### DIFF
--- a/packages/e2e-tests/tests/export.spec.ts
+++ b/packages/e2e-tests/tests/export.spec.ts
@@ -1,0 +1,57 @@
+import {test, expect} from "@playwright/test"
+import path from "path"
+import {testDataDir} from "../helpers/env"
+import TestApp from "../helpers/test-app"
+import os from "os"
+import fsExtra from "fs-extra"
+
+const tempDir = os.tmpdir()
+const formats = [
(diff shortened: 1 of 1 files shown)


jameskerr (commented):
Excellent!

jameskerr (approved):

jameskerr:
Watching all these run was very satisfying

image

PR #2511, opened 2022-08-31
Point at new docs site

### LINKED ISSUE #2360: Update "Syntax docs" link in Brim app's query bar drop-down menu
Hi Brim team,

I was playing with the toy data in Brim yesterday and noticed that the "Syntax docs" link sent me to GitHub, but the new Zed website docs seemed more helpful.  The description below details things.

**Describe the bug**
Brim app's "Syntax docs" link directs to older GitHub documentation but the new Zed docs are more directly helpful.

Here is the current link: https://github.com/brimdata/zed/blob/v1.1.0/docs/language/README.md

**To Reproduce**
Open the Brim app and click the query bar's settings drop-down icon.

**Expected behavior**
Link to: https://zed.brimdata.io/docs/commands/zq/ (or something else on the new Zed website).

**Screenshots**
Screen Shot 2022-05-23 at 3 42 00 PM

**Desktop Info**
 - OS: MacOS
 - v0.30.0 downloaded via: https://www.brimdata.io/download/

philrz:
As described in #2360, the Zed Syntax Docs link from the app has been pointing at docs URLs in the Zed GitHub repo. The new docs site makes a better destination, though.

In the past, if the app was running a random Zed commit rather than a tagged GA version like `v1.2.0` the URL would point to that specific commit hash in the Zed repo. However, the new site doesn't have docs at every hash, so this PR turns those cases into version `next`, which is effectively the docs as of the tip of Zed `main`. I expect most users running non-tagged Zed with their Brim apps would be people hacking with Zui Insiders, so this destination seems to make sense anyway.

To test, I set the Zed dependency in `package.json` to each current possible variation.

When running Brim commit 0029841 that points at Zed commit `d07901b`, we land at https://zed.brimdata.io/docs/next/language.

image

After changing the Zed dependency to `v.1.2.0`, now we land at https://zed.brimdata.io/docs/language. Note the lack of the explicit version tag in the URL. This is due to the redirect config introduced in https://github.com/brimdata/zed-docs-site/pull/40.

image

Finally, if I change the Zed dependency to `v1.1.0`, now we land at https://zed.brimdata.io/docs/v1.1.0/language.

image

Since this PR also changes the Zed format links on the import page, I tested those as well. With the Zed dependency still set to `v.1.1.0`:

image

image

image

import pages

Fixes #2360

### DIFF
--- a/src/app/core/links.ts
+++ b/src/app/core/links.ts
@@ -1,11 +1,12 @@
 import * as brimPackage from "../../../package.json"
 
 const currentZedTag = brimPackage.dependencies.zed.split("#")[1] || "main"
+const zedDocsTag = currentZedTag.startsWith("v") ? currentZedTag : "next"
 
 export default {
-  ZED_DOCS_ROOT: `https://github.com/brimdata/zed/blob/${currentZedTag}/docs/commands/zed.md`,
-  ZED_DOCS_LANGUAGE: `https://github.com/brimdata/zed/blob/${currentZedTag}/docs/language/README.md`,
-  ZED_DOCS_FORMATS_ZNG: `https://github.com/brimdata/zed/blob/${currentZedTag}/docs/formats/zng.md`,
-  ZED_DOCS_FORMATS_ZSON: `https://github.com/brimdata/zed/blob/${currentZedTag}/docs/formats/zson.md`,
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Cool

PR #2513, opened 2022-09-01
Overflow hidden on the results

### LINKED ISSUE #2492: "Scroll right" makes column headers bleed into pools list
Repro is with Brim commit 290f138 and these steps as shown in the attached video:

1. Load some data such as the [sample.zng](https://github.com/brimdata/brim/blob/main/src/test/shared/data/sample.zng) form this repo
2. Set the **Column** headers to **On**
3. Scroll rightward in the main events panel

The column headers can be seen scrolling leftward into the pools list on the left side panel.

https://user-images.githubusercontent.com/5934157/185238528-a107b4ba-5919-44c5-9ee9-df9478d7ce09.mp4

jameskerr:
This must have happened in a merge conflict.

Fixes #2492

### DIFF
--- a/src/app/query-home/results/index.tsx
+++ b/src/app/query-home/results/index.tsx
@@ -21,6 +21,7 @@ const BG = styled.div`
 
 const Body = styled.div`
   flex: 1;
+  overflow: hidden;
 `
 
 const ResultsComponent = () => {


philrz (approved):
I just tested this out on the branch and the fix indeed appears effective. Regarding its origin, it looks like this started life without the `overflow: hidden;` in https://github.com/brimdata/brim/pull/2439/files#r961143278 when that file was otherwise going through lots of changes. 🤷‍♂️

Issue #1493, opened 2021-03-03
Manual Brim Test Cases


jameskerr:
These areas are not yet covered by integration tests. Some of them are too complex to be automated with our current tooling (Spectron), but others should be.

- [ ] Log detail window, right click search in new tab
- [ ] Zoom in to histogram with a single click and a double click
- [ ] Shutdown and reopen app with multiple windows
- [ ] Refresh the window
- [ ] Create a workspace
- [ ] Delete a workspace with that workspace active in another tab, then go to that tab
- [ ] Renaming a space when other tabs are open to that same space
- [ ] Press the add space + button
- [ ] Use the form to add a space
- [ ] Check the app after killing zqd
- [ ] Ingest auto refresh when doing nothing
- [ ] Ingest auto refresh when searching
- [ ] Infinite scrolling
- [ ] Relative dates appear in the url
- [ ] Relative dates work in the data
- [ ] Double click a log to open the log detail window
- [ ] Import a corelight suricata record and open the log detail pane
- [ ] Open the log detail pane on various sources of data
- [ ] Use the column picker to select certain columns, "auto", "on", and "off"

Issue #1345, opened 2021-01-09
Detail for "Import complete" pop up


philrz:
Repro is with Brim commit `f6c7ba3`.

As shown in the attached video, let's say I start a big pcap import on the first tab, a small pcap import on the second tab, and then switch focus back to the first tab. I'm shown an "Import complete" pop up regarding the completion of the small pcap import. As a user that may have many tabs open for different data sets, my expectation is that the messages I'm seeing on the current tab would be regarding the current tab. If we want to preserve this behavior so the user can be informed as activities occur on other tabs, I might suggest we include more info in the message to disambiguate it (e.g. the name of the Space for which import completed).

https://user-images.githubusercontent.com/5934157/104105350-3c7c8c00-5262-11eb-9926-bf549b40121b.mp4

philrz:
I checked with current Brim commit 91b323d to see whether this issue is still with us in the "pools" era, and it is. That said, the original sentiment of the pop-up being on the "wrong tab" no longer applies, since in Session Flow a tab is no longer assumed to be particular to any one pool. However, it's still true that a user may have simultaneous imports happening on multiple tabs, so it does seem it would still benefit the user to see some additional detail, e.g., the name of the pool that just completed import.

https://user-images.githubusercontent.com/5934157/187543826-ae3b2924-d70a-4150-b66a-b1c680ceccfe.mp4

philrz:
Verified in Zui commit a9d64be.

We addressed this one some time ago. As shown in the attached video, the pop-up notification that a pool load has completed now includes the name of the pool. This should remove any confusion if a user is on an unrelated tab when a load completes.

https://github.com/brimdata/zui/assets/5934157/da382a55-9aa1-4df3-b92b-5676afe82e6e

PR #2514, opened 2022-09-01
Increase toast loading duration to biggest number

### LINKED ISSUE #2497: Export continues long after "Exporting..." pop-up goes away (regression)
Repro is with Brim commit 9912a29. An issue with the exact same symptom was found once before in #1346 and ultimately fixed via #1424, so this represents a regression. Perhaps info from the prior fix will help quickly narrow down and fix this one also.

The attached video shows the initiation of a large ZNG export. After the "Exporting..." pop-up fades away, you can see the ZNG file continues to increase in size, indicating the export operation is still in-progress. As a user, a pop-up that communicates an in-progress operation (e.g. a verb ending in "ing"), my assumption is that when the pop-up goes away, the operation described has completed.

https://user-images.githubusercontent.com/5934157/186023455-bf4773a0-8c2d-412c-ac1a-9507926fb3b1.mp4

jameskerr:
Fixes #2497

I tested this by exporting 3.4 gigs of ndjson. The export took over a minute, did not disappear, and then displayed the "export complete" 
message.

### DIFF
--- a/src/js/components/BrimToaster.tsx
+++ b/src/js/components/BrimToaster.tsx
@@ -14,7 +14,7 @@ const BrimToaster = () => {
           duration: env.isTest ? 2 ** 31 - 1 : undefined,
         },
         loading: {
-          duration: env.isTest ? 2 ** 31 - 1 : undefined,
+          duration: 2 ** 31 - 1,
         },
         error: {
           duration: env.isTest ? 2 ** 31 - 1 : undefined,
--- a/src/js/components/BrimToaster.tsx
+++ b/src/js/components/BrimToaster.tsx
@@ -1,4 +1,3 @@
-import env from "src/app/core/env"
 import React from "react"
 import {Toaster} from "react-hot-toast"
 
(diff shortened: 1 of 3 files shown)


philrz:
@jameskerr: Because the repro involves large data sets and a long delay, am I guessing this one is too hairy to add a test for? I ask because we've regressed in this area after fixing it once before, so my usual instincts are to plead for a test so this doesn't keep happening. 🙁

jameskerr:
I know, I was trying to think of how to test this. I think there is something I can do. I'll add.

jameskerr:
I also added the file name to the export message:

image


In the future it would be sweet to add a little link to open the containing folder in the default file explorer.

philrz:
Oooo! I like the file name in the export message. It's orthogonal so not asking for this now, but it reminds me of the Import pop-up that could use similar detail per #1345.

philrz (approved):
My test on this branch was a successful, so that's a functional 👍 from me!

Also, I've added a checkbox to #1493 as a reminder that we may one day want to revisit this area to add more testing.

jameskerr:
Yup, that was on my mind so I added it here too. Alright!

PR #2518, opened 2022-09-06
Make the history button purpose more clear

### LINKED ISSUE #2476: Tooltip for clock icon
Repro is with Brim a20f709.

As a user, I'd forgotten what the clock icon was supposed to do. At first I smashed on it several times and nothing visibly changed, so at first I wondered if there might have been a bug. Then I remembered that it had something to do with the right panel. As shown in the attached video, I ultimately deduced that it seems to have the purpose "Open the History tab on right panel", which explains why it was acting like a no-op when I already had the panel open & on that tab. Assuming this is the intended functionality, I think adding a tooltip saying something to that effect would be helpful in the event users get stuck in the same spot I was.

https://user-images.githubusercontent.com/5934157/183488626-df257ac6-f4fe-4423-a960-64b931356066.mp4

jameskerr:
Fixes #2476 

The video tells it all!

https://user-images.githubusercontent.com/3460638/188738069-0792c0c2-ee78-47ff-b3e7-a186154853f3.mp4

### DIFF
--- a/src/app/commands/flash-element.ts
+++ b/src/app/commands/flash-element.ts
@@ -1,0 +1,15 @@
+import {createCommand} from "./command"
+
+export const flashElement = createCommand<[selector: string]>(
+  {
+    id: "flashElement",
+  },
+  (ctx, selector) => {
+    const el: HTMLElement = document.querySelector(selector)
+    if (!el) return
+    el.classList.add("flash-element")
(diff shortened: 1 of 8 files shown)


philrz (approved):
Tests out nicely for me on the branch, so that's a functional 👍 from me!

PR #2525, opened 2022-09-09
Don't crash if state file does not exist.

jameskerr:
If the JSON is appState.json is in an unexpected state, the app won't start.

In the example below (on main branch), I've changed app state.json to be what's on the left, and I was able to re-create Gilles issue.

CleanShot 2022-09-09 at 08 31 06@2x


jameskerr:
We are not going this route.

PR #2526, opened 2022-09-09
Don't work on state for about windows

jameskerr:
This will fix Gilles startup issues. We were trying to work on "state" from an about window, which has no state.


philrz (approved):
The fix in the branch was effective in my test, so that's a functional 👍 from me.

PR #2529, opened 2022-09-09
Pin Problem

### LINKED ISSUE #2528: "from" pin is dropped when saving query (regression at b08ceaf)
Repro is with Brim commit 9b2c9c4, though I've confirmed via binary search that this problem first showed up at commit b08ceaf, which is associated with the changes in #2472.

As shown in the attached video, when a "from" pin is present and the query is saved, the "from" pin is immediately dropped.

https://user-images.githubusercontent.com/5934157/189424106-c5b31f3d-92e0-4317-848a-cb60821302df.mp4

jameskerr:
I was overwriting the provided pins with the default. Face palm. I've added a test that will break if that happens again.

Fixes #2528

### DIFF
--- a/packages/e2e-tests/helpers/test-app.ts
+++ b/packages/e2e-tests/helpers/test-app.ts
@@ -51,7 +51,7 @@ export default class TestApp {
         })
       )
     )
-    this.mainWin = await this.getWindowByTitle("Brim")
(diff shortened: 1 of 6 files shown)


philrz (approved):

philrz:
The fix tests out nicely in the branch, so that's a functional 👍 from me.

PR #2530, opened 2022-09-12
Pin Design Update

### LINKED ISSUE #2386: Make pin management closer to the pins
Design needed

jameskerr:
The new pin button is moved right next to where the pins go. The "Run" button is moved to the Title Bar.

image

https://user-images.githubusercontent.com/3460638/189771941-85beb6c0-2398-4043-860f-2f920adb1099.mp4

Closes #2386


philrz (approved):
Yes, seems like a great improvement, so a functional 👍 from me!

This doesn't have to be something that's added right now, but seeing these changes made me think again of the yellow highlighting pulse you added on the word "History" when the user hit the clock icon when the History panel was already open (#2518). Would it be feasible to add a similar yellow pulse effect to the dotted/placeholder "from" pin area when an "Error: pool name missing" comes up? Granted, this is really only a concern for users getting the hang of this for the first time. I'm mostly just concerned about users that were accustomed to the old workflow and wanting to give them the easiest possible time understanding the importnace of the "from" pins now.

PR #2535, opened 2022-09-14
A Page for When the Pool is Missing

### LINKED ISSUE #2477: Help users pick a pool for their new Query Session
Repro is with Zui commit a20f709.

I was doing some testing as a user mostly familiar with the prior app workflows and new to the Session Flow introduced in #2439. As shown in the video below, it made sense to me that when clicking "Query Pool" from a pool info page that it started me out in a new Query Session with that pool already selected in the "FROM" pull-down. However, when I later clicked "+" to open a new Query Session tab, I was confused by the absence of a similar "FROM" pull-down. I thought that maybe some kind of "blank" pull-down might be shown to help me make a pool selection.

https://user-images.githubusercontent.com/5934157/183520365-5fda4898-8da8-44ff-96d0-615e77f09af0.mp4

I showed this to @jameskerr and he reminded me that that the "FROM" is actually a "Pin" query element, so the way to add it in a new tab would be to click the "Pins" pull-down as I do later in the video and select "New "From' Pin". This works as shown in the video, but here's some additional thoughts on how this could be made easier for me to have figured out on my own.

1. @jameskerr mentioned that there's already plans to move the "Pins" button closer to the actual pins. Indeed, if that had been the case (e.g., if "Pins" were just to the left of where the "FROM" was previously) I'd have made the connection to click it and see the 'From' option.

2. As shown in the video, when "New "From' Pin" _is_ clicked, it effectively creates a sort of "blank" pull-down selector. Therefore, when a user clicks to open a new Query Session, it might be helpful to do this for them. Granted, a user at this point could manually type the `from "dce_rpc.log.gz"` as part of a whole query, and this might indeed be desirable for some users that are pasting fully-formed Zed from CLI use cases. However, I suspect that more app users are going to be working mostly/entirely within the app and hence would appreciate saving the clicks of adding a pin.

3. Now that I knew all this, on a related topic I asked @jameskerr why clicking "Query Pool" doesn't currently auto-add a "Time Range" pin with the start/end times for the pool key already filled out, since that would seem to match what we're doing with the auto-added "FROM". @jameskerr acknowledged that this is something he's thinking about currently as he's working on bringing back the Zeek-centric histogram (#2472).

jameskerr:
Fixes #2477 

image


philrz:
@jameskerr: Seems great! I tested out on this branch at commit ee1ba93 and found one little rough edge. As shown in the video below, when I have a very small number of pools, the vertical real estate in the pool list seems excessively small such that it cuts off part of even just a single pool name when that's all I've got. So it looks like a little padding would help here.

https://user-images.githubusercontent.com/5934157/190274823-8487ee8f-dd20-46f0-9fe0-834183f0054e.mp4

jameskerr:
Now it looks like this with lots of pools:

image

And like this with few pools:

image

philrz (approved):
I've tested this branch out at commit 31eb00d and it's a functional 👍 from me. My understanding is that @jameskerr still has a test to add, but since i"m not going to be around later today to click Approve, I'm approving it now so it'll be easy to merge once the test is in and passing.

Issue #2551, opened 2022-09-22
Buggy multi-select pool delete operations


philrz:
Repro is with Brim commit b7ee21a which uses Zed commit `6737ae9`.

As shown in the attached video, it's currently possible to Shift-Click to multi-select many items in the pools list. This may give users the impression that a selection in the right-click menu will now operate on the multiple selected pools, but this is currently not the case.

For instance, if a regular Delete is selected, only the pool that the mouse pointer was on during the right-click is actually offered for deletion.

The Delete All is a bit more hazardous since if the user doesn't pause and fully take in the words in the pop-up notice, they might be thrown by all their pools being deleted rather than just the highlighted ones. Also, the video shows the unexpected "config journal unavailable after 10 attempts" error from the backend during this operation. I'm not sure if this is a sign that the frontend is hitting the API in a bad way or if it's a backend bug, but also seems worth looking closer at.

Of course, I expect we could make this all go away by just not allowing for multi-select on the pools list, which might be the expedient route. 😄 

https://user-images.githubusercontent.com/5934157/191823583-7076b7e4-f06f-4296-90ff-775f08758edb.mp4

jameskerr:
Yup, that's no good

Issue #2540, opened 2022-09-19
Dragging a query folder to the bottom makes it disappear & breaks further operations


philrz:
Repro is with Brim commit 24d658. This issue seems somewhat similar in spirit to #1906.

Repro steps as shown in the attached video:

1. Drag the [Brimcap Queries](https://github.com/brimdata/brimcap#brimcap-queries) folder into the app
2. Make a new folder and add a query into it
3. Drag the new folder to become the bottom entry under the Brimcap folder
4. Drag the new folder just below the bottom of the Brimcap folder entries

At that last step, the new folder disappears from the interface and the following error appears in DevTools:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'model')
    at /Users/phil/work/brim/dist/js/state/Queries/reducer.js:64
    at immerClass.ts:80
    at Immer.produce (immerClass.ts:96)
    at curriedProduce (immerClass.ts:80)
    at combination (/Users/phil/work/brim/node_modules/redux/lib/redux.js:465)
    at k (<anonymous>:2235:16)
    at D (<anonymous>:2251:13)
    at <anonymous>:2464:20
    at Object.dispatch (/Users/phil/work/brim/node_modules/redux/lib/redux.js:219)
    at e (<anonymous>:2494:20)
```

Further attempts to do operations in the query library do not succeed, and DevTools keeps getting spammed with errors like:

```
Uncaught Invariant Violation: Cannot call hover after drop.
    at invariant (/Users/phil/work/brim/node_modules/@react-dnd/invariant/dist/invariant.cjs.development.js:31:15)
    at checkInvariants (/Users/phil/work/brim/node_modules/dnd-core/dist/cjs/actions/dragDrop/hover.js:43:28)
    at DragDropManagerImpl.hover (/Users/phil/work/brim/node_modules/dnd-core/dist/cjs/actions/dragDrop/hover.js:23:5)
    at Object.hover (/Users/phil/work/brim/node_modules/dnd-core/dist/cjs/classes/DragDropManagerImpl.js:86:38)
    at /Users/phil/work/brim/node_modules/react-dnd-html5-backend/dist/cjs/HTML5BackendImpl.js:271:23
```

https://user-images.githubusercontent.com/5934157/191072309-ebee4519-d41d-4778-9b4b-bd33c6111893.mp4

jameskerr:
Thanks for reporting.

philrz:
Verified in Brim commit a6fc1a8.

[...]

Issue #2508, opened 2022-08-30
Rename-in-place


philrz:
Brim is at commit 91b323d as this issue is being filed.

Currently, renaming things like pools and saved queries requires a right-click and selecting a **Rename** option. This is pretty easy, but as @jameskerr and I agreed in https://github.com/brimdata/brim/issues/1477#issuecomment-782264316, tools like Finder and Windows Explorer remind us that there's even easier ways to activate a rename, such a user single-clicking the entry after they've already clicked it once to select it.

PR #2568, opened 2022-10-14
react arborist 2

### LINKED ISSUE #1924: Unable to deselect multiple queries
When using the new query library to select multiple queries, you can't deselect them by clicking the outside of the list. The listener should just be on the containing div.

image

### LINKED ISSUE #2199: Scroll to sidebar item after creation
When creating a new query, dragging a query library, create a pool or dragging a pool, we must scroll the sidebar to that item.

### LINKED ISSUE #2500: Clickable ways to add/replace a "from" pin
Users of the older Brim app may be accustomed to the workflows where a `from` pool was always pre-selected for them based on which pool they had highlighted in the left sidebar. In Session Flow as it stands today, this is now handled typically by either an explicit "From pin" or a typed-out `from` at the start of the Zed code in the editor. A "From pin" along with the pool it references is also saved/recalled with Saved Queries, which might not always be desirable (see #2478).

In a recent team discussion, @jameskerr proposed a couple ways that might make it easy for users to adapt to this in clickable ways.

1. If you're in the query editor and you Cmd-click or Alt-click a pool, it would be added as a "From pin" if one doesn't already exist, or replace the current "From pin" if one already exists.

2. Make it so if the user drags the name of a pool from the left sidebar into the editor, it would similarly add/replace the current "From pin"

### LINKED ISSUE #2560: Pool creation resets pool search
Repro is with Brim commit 6eaaa54.

To make it easy to observe the problem, we start a loop of new pools being created, in this case using the [Zed sample data](https://github.com/brimdata/zed-sample-data).

```
NUM=1; while true; do for file in *; do   zed create pool_00000_dev01_table_${file}_${NUM};   zed -use pool_00000_dev01_table_${file}_${NUM} load $file; done; NUM=`expr $NUM + 1`; done
```

As shown in the attached video, if we attempt to perform a search of the available pools, the trimmed set of results seems to be "reset" back to the full set each time the list is refreshed by the arrival of a new pool. Our results only "stick" when we stop the script from creating more pools. While a production environment is not expected to have pools being created this rapidly, in a shared environment, a user might be thrown by their searches suddenly being interrupted by the actions of some external user or automation that creates a pool.

https://user-images.githubusercontent.com/5934157/192845917-52ec726e-6b66-4df9-8d7b-7799c89105c6.mp4

### LINKED ISSUE #2564: False hits in search of saved queries
Repro is with Brim commit 6eaaa54.

In the attached video, I create a saved query called "tcp". I then start some searches of the saved queries by typing a single character in the "Search Local Queries..." box. As it shows, for some characters like "a" and "e", the query is still shown. This is not expected, since I'd have expected it to disappear if I type any character other than "t", "c", or "p".

https://user-images.githubusercontent.com/5934157/192894217-50e661e2-c849-484c-9f2e-a2d6a08bf01a.mp4

jameskerr:
All Tree's have been updated to React-Arborist 2.0.0

* Queries Tree (remote and local)
* Pools Tree
* Versions List
* History List

Features:

* You can alt+click pools to make them the from pin
* You can filter the pools and queries list
* You can create queries easier with the right click menu
* Selection is automatically synced and scrolled to (Fixes #2199)

Fixes #2564
Fixes #2560
Fixes #2500
Fixes #1924


philrz:
I'm testing the branch at commit 0a96a8b.

One little issue I've spotted is that the filtering seems to be case insensitive, but not completely. As shown in the attached video, entering a filter of lowercase "act" narrows down the query list to just some entries that include the characters "Act", so that's when I got the sense it was case insensitive. However, if I note that the existing entries start with a capital "Act" and I enter that instead, now the filter excludes all entries, which was unexpected.

https://user-images.githubusercontent.com/5934157/197407638-3958cbaa-af91-47ef-817a-0d719bbffc03.mp4

philrz:
Here's another finding at commit 0a96a8b.

In the attached video, I perform the following actions:

1. Create a Local saved query
2. Copy it to Remote and confirm it's there
3. Click back to my Local queries
4. Right-click on the Local query and attempt to delete it

The pop-up appears saying that deletion was successful, but the Local query still appears intact. Clicking to Remote, I see _that_ query is the one that got deleted, which was unexpected.

https://user-images.githubusercontent.com/5934157/197407876-28e703af-788b-4d6a-9010-1baef0b4ed8c.mp4

philrz:
In a positive finding when testing at commit 0a96a8b, I see that #2564 appears to be fixed by this PR. In the attached video after saving a query called "tcp", I enter each letter of the alphabet in the filter box and now the only ones where the query is still visible are the expected "t", "c", and "p", with no false hits anymore. I've edited the opening issue text so #2564 will be auto-closed when this PR merges.

https://user-images.githubusercontent.com/5934157/197409566-477ecd78-3149-4c41-9686-571c09e04009.mp4

philrz:
Testing at commit 0a96a8b also shows that #2560 appears to be fixed by this PR. As shown in the attached video, now the creation of new pools as triggered by the external script no longer resets the filtered results. Instead if I have an active filter like "ftp" or "dns", now only new pool entries that match those terms appear in the list. I've edited the opening issue text so #2560 will be auto-closed when this PR merges.

https://user-images.githubusercontent.com/5934157/197409919-ee5a8551-30b9-4d3e-a586-e62cb197bc42.mp4

philrz:
More testing at 0a96a8b. As shown in the attached video, it seems that issue #2551 is _mostly_ fixed by this PR.

The first problem highlighted seems to be completely addressed. Now when doing a multi-select of the same set of pools as in the original repro and then executing a right-click **Delete**, all the highlighted pools are deleted instead of the singleton that was deleted in the past. 👍

For the second problem highlighted in the repro, the **Delete All** option seems to be gone from the right-click menu, which seems fine. But acting as a user trying to do the equivalent with what's currently available, hitting Cmd-a to highlight everything and then selecting the right-click **Delete**, I still see a "config journal unavailable after 10 attempts" error as we saw before. I'm not necessarily advocating that we need to fix this problem in this PR, but wanted to point out that we're closer on that issue. If we end up merging this PR without addressing it, I'll update #2551 to reflect what's improved & what's left.

https://user-images.githubusercontent.com/5934157/197410504-e732f921-858a-4fe9-8c7e-052213eba6d0.mp4

philrz:
More testing at 0a96a8b.

I'm not sure if this one is appropriate to address here, but while checking if #2540 was addressed by this PR, I stumbled onto the issue in the attached video. What it shows that with a query folder expanded, at first a right-click **New Folder** operation places the created folder inside the existing query folder. However, the act of collapsing and expanding the existing query folder and then selecting **New Folder** again causes the created folder to appear _outside_ the existing query folder.

As a user, I'd have expected the new folders to always be _outside_ of the existing query folder, since I was right-clicking in open space below all existing queries/folders.

https://user-images.githubusercontent.com/5934157/197411548-48d2e7f9-1e17-4c3a-92ad-affa90f607ad.mp4

philrz:
More testing at 0a96a8b.

It looks like issue #2500 has been addressed by this PR, since per the opening comment from @jameskerr,  Alt-click now populates/replaces the `from` pin, which is the first of the two ideas floated in #2500. On further reflection, the second idea floated in #2500 probably doesn't make sense since single-clicking an entry in the pools list with intent to drag it into the editor would make the editor disappear due to the main window being populated with the pool info page. Therefore I've edited the opening issue text so #2500 will be auto-closed when this PR merges.

philrz:
More testing at 0a96a8b.

As shown in the attached video, it appears that #1924 is addressed by this PR. Therefore I've edited the opening issue text so #1924 will be auto-closed when this PR merges.

https://user-images.githubusercontent.com/5934157/197414837-2539164f-72fd-4e00-bbe5-f4864bd72d31.mp4

philrz:
@jameskerr: I noticed that #2508 is partially addressed by this PR in that hitting Enter on a highlighted entry now starts a Rename, which is definitely an improvement vs. where we were previously where the user could only right-click. Would it be easy to take it all the way by also making a single-left-click-after-a-brief-pause on an already-highlighted entry also start a Rename? That would seem to make the behavior match with tools like macOS Finder and Windows Explorer with which users are surely familiar.

philrz:
More testing at 0a96a8b.

Regarding the expected fixes for #2199, the attached video shows these mostly addressed. Of the items listed in #2199:

1. When creating a new query, the sidebar does indeed scroll to the new query entry ✅ 
2. When creating a new pool by clicking **New Pool** and clicking the **Choose Files** button, the sidebar does indeed scroll to the new pool entry ✅ 
3. When creating a new pool by dragging it into the **Pools** list, the sidebar does indeed scroll to the new pool entry ✅ 

Things I noticed that don't quite match up with expectations:

1. Dragging in a query library did not change what was highlighted in the sidebar. I'm not sure this is super important, but it might be a nice convenience to have the name of the newly-created folder highlighted so the user could immediately hit Enter and start a rename if they didn't like the built-in name.
2. [A comment in #2199](https://github.com/brimdata/brim/issues/2199#issuecomment-1067102406) speaks of intent to also include items created in the sidebar through right-click action, but right now right-clicking among the entries in a query folder doesn't offer an option that can create a new entry. I'm not sure if this was intended. FWIW, a right-click at the top level outside of any folder or right-clicking on a highlighted folder name (not shown in the video) _does_ create a new entry and effectively scrolls to it by starting a Rename, so if that's all that was intended, I don't think there's any action needed here.

https://user-images.githubusercontent.com/5934157/197416181-6ae01721-9841-4138-9458-dda31c8a5730.mp4

philrz (approved):
I've re-tested the fixes to all the loose ends I highlighted in recent comments and marked with a 🚀 the ones that checked out ok, which is all of them. It's therefore a functional 👍 from me at this point. Thanks @jameskerr!

PR #2571, opened 2022-10-25
Blank Content Type

jameskerr:
Send no content type for load method.


nwt (approved):
This makes tests pass with brimdata/zed@29557045038182ba6488d7eac73f5c2316559753 so :shipit:.

PR #2572, opened 2022-10-25
Pool Folders

jameskerr:
Pools can now be grouped by a specified delimiter. The default is the "/" character.

The groups can be collapsed and expanded.

When a group is renamed, all of the pools under that group get renamed one by one. This can be improved when the backend can handle multiple requests without error. Then we can send many rename requests at once.

When a group is deleted, all of the pools under that group get deleted one by one. 

- [x] Persist the openness state between renders.


philrz:
I tested this out on the branch at commit 1623df4. The only glitch I've been able to find is related to the new setting in **Preferences**. Starting out with it at the default `/`, I have this hierarchy:

image

I thought that maybe if I was a user that didn't want any kind of grouping, I could delete the value. But when I did that, I get:

image

I have a guess why it's doing that, but I imagine it'd make more sense if it turned off the functionality at that point.

philrz (approved):
I just re-tested at commit cfeb46a and it seems to be working great. Functional 👍 from me!

PR #2573, opened 2022-10-26
Run on Enter Preference

jameskerr:
If the setting is set to true, enter runs the query, shift+enter gives a new line.

If the setting is false, cmd+enter runs the query and enter gives a new line.

CleanShot 2022-10-25 at 17 21 02@2x


philrz (approved):
I've tested this branch at commit a470f02 and everything is working as I'd expect. On Windows and Linux, when the box was unchecked, Ctrl+Enter was now used to execute the queries, and on macOS it was Cmd+Enter. Therefore it's a functional 👍 from me.

PR #2580, opened 2022-11-01
Use replacement for deprecated set-output

philrz:
It was recently [announced](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) that the `set-output` syntax is being deprecated in Actions. I tested out the replacement syntax in a personal repo to get a feel for it and it works as described, so here's changes to our own Workflows to adopt the new approach.


jameskerr (approved):
Much cleaner.

PR #2581, opened 2022-11-03
Fix pool ids

### LINKED ISSUE #1805: Configurable organization/view of pools
A community user that's been helping with the beta testing had the following suggestion:

> When Brim displays the pools it would be nice to be able to rearrange them in a different order.  Then I thought, a directory structure would be even better than that.  If I had a 100+ pools it would be nice to be able to organize it all by some kind of structure.  This would probably be helpful on the CLI as well. 

Just yesterday @jameskerr was showing off a "tree view" he's been working on that allows for organizing items in the app into hierarchies, add/delete/rename, etc., so being able to re-use that component was the first thing that came to mind when hearing this. But it seems there's multiple angles to consider, so here's a dump of some of my initial thoughts.

As implied in the inquiry, it seems like an open question: "Where does the organization start?" Right now the set of Pools is flat in a Zed Lake, so depending on how users approach it, indeed I can see the potential to accumulate enough Pools that it becomes a challenge to browse them at both the CLI and app layers. This might make the case for allowing hierarchical organization at the Zed Lake layer and having that order be reflected "read only" in all clients. This would seem to imply an admin-defined organization of pools that's presumed to meet the needs of all users.

OTOH, users might appreciate the flexibility of being able to create their own personal view of how the pools are organized. What comes to mind is the "Quick Access" panel in Windows file explorer. This effectively consists of "pins" that point to commonly-accessed items in the large, hierarchically-organized filesystem. When browsing the filesystem, the user can right-click on any item and select "Pin to Quick Access". The entries already in Quick Access can be re-ordered or removed to suit the user's preferences. It looks like they can't be organized into a hierarchy, but we could still do it in ours if we wished. Of course, it may be seen as a drawback that every user is now responsible for maintaining this themselves.

I suppose we could get fancy and offer this at _both_ layers. Once again, borrowing from the Windows example, there could be pool hierarchy/order defined within the Zed Lake itself, and if the pool count and admin-defined organization is tidy enough that it satisfies all users, they'd be free to stick with just this approach. Then if they outgrow it, something more visualization-layer like "Quick Access" could be made available to offer another layer of user-specific organization.

jameskerr:
Use a stable id for the pool folders.

Previously, we used the names of the folder. That lead to an error where an id was "". Now we use the index of the pools and the index of the folders.

This also fixed an issue where the openness state changed when you renamed something. Now that the id is not dependent on the name, this is fixed.

https://user-images.githubusercontent.com/3460638/199813762-6ebed6c5-f019-4dba-b3b0-071d86f3b665.mp4

Closes #1805


philrz (approved):
I tested it out in the branch and the failure I reported in https://github.com/brimdata/brim/issues/1805#issuecomment-1302482799 no longer occurs if I happen to rename a pool such that it now starts with a `/`, so that's a functional 👍 from me.

PR #2584, opened 2022-11-08
Load More Data Into Pool

### LINKED ISSUE #1982: Dragging files to an existing pool to add more data
As of Brim `v0.28.0`, I see that dragging a file to the pools list always creates a newly-created pool to store the data. The only way I know of to add additional data to an existing pool is to drop to the shell and run a `zapi load` command line. But it'd be great if this could be accomplished right inside the app as a variation on the existing drag & drop UX. For instance, if my drag movement ends with my mouse hovering over the name of an existing pool, add the data to that pool. If the drag movement ends with my mouse over the "blank" area of the pools list (and maybe the "POOLS" header as well, just in case the pools list is full of pool names) it could maintain its current behavior of creating the new pool.

While browsing the open issues to see if this feature had already been requested, I came across #1067 that spoke of this in a pcap context. That text pre-dates the current Brimcap/pool constructs, but thinking about how I last recall the current app working, it seems like it might not be too difficult to cover this for the pcap case as well: Brimcap invocation effectively ends with a `zapi load` (right?), then Brimcap's pcap index could also be populated with the info for the new pcap, and I think you'd be covered. Whatever we end up doing something for #1042 also seems relevant.

jameskerr:
Allows loading more data into a pool.

Changes the "confirm" message you get when you try to quit the app while data is loading. That would be good to test out.

Closes #1982


philrz:
Testing on the branch at commit ac69d36.

When I try to load garbage data (e.g., my `/etc/passwd`) into a pool on the branch, the error message returned from the backend no longer has line breaks.

image

Meanwhile at current tip of `main`, which is commit 1db0181, it's still more readable:

image

philrz:
Testing the branch at commit ac69d36.

I spotted this error handling in the PR and figured I'd try witnessing it:

https://github.com/brimdata/brim/blob/ac69d364891e54020ea46a2ef4a311a71871893a/src/js/api/pools/create-and-load-files.ts#L26-L36

As shown in the attached video, dragging a directory seems to causes a "Loading data into pool..." message to appear that never goes away. An error message does show up in DevTools.

https://user-images.githubusercontent.com/5934157/202329318-507ee46a-cfd6-44df-ad72-29defd11b74c.mp4

For some reason I thought the app used to handle directory loads, but maybe that was a long, long time ago (or my memory is bad). FWIW, when I did a quick test of the same on current tip of `main` (commit 1db0181), it quietly does nothing, though a similar error shows up in DevTools.

https://user-images.githubusercontent.com/5934157/202330019-baaeaa2f-2b2a-42e6-b24d-0efb38ab3f85.mp4

So it seems like we already had an issue here, but the "Loading" message arguably makes things a little worse since it gives the user an impression that something is happening when it isn't.

philrz (approved):
I've re-tested on the current branch and the couple of issues I'd spotted previously seem to be fixed and everything else seems ok, so that's a functional 👍 from me!

PR #2586, opened 2022-11-08
Improve on the term "Generic Pin"

### LINKED ISSUE #2532: Improve wording for "Generic pins"
Repro is with Brim commit c36fada.

During a team meeting, it was spotted that the generic-ness of "Value" when creating a Generic Pin might be too general and hence confuse some users. Ideas for improvement were to call it a "Zed fragment" or a "Zed snippet". In addition to replacing the word "Value" in the pop-up, the same wording could also be used in place of "Generic" when describing the pin.

image

philrz:
In a sync meeting some feedback was given that the terms "Generic Pin" and "Value" were maybe a bit too... generic. Here's a take at some of the alternatives we discussed to try on for size.

image

image

Closes #2532


jameskerr (approved):
I'm into it.

PR #2596, opened 2022-11-21
Remove unreferenced test data

philrz:
I was starting to try my hand at writing an e2e test and thought I might need some special test data. To avoid unnecessary bloat, I figured I should first see if I could make use of any of the existing test data. I quickly found that some of them were old enough that modern Zed tooling doesn't even see them as valid formats anymore. That got me to grepping the repo, which revealed that the ones I've removed in this PR don't seem to be referenced in any tests anymore. I found that `yarn test` and `yarn e2e` both still run fine with all of these removed.

The one thing I'll point is that the filenames of a couple (`custom-sample.ndjson` and `types.tsv`) are mentioned by name in `src/test/unit/states/v0.20.0.json`, but the way they're referenced doesn't make me think they're actively _used_. But pointing it out just to be sure.


jameskerr:
Thank you @philrz ! These all look like they go. The references to the data in the state file are probably pool names. That state file only exists to help test migrations. So if the tests all pass, they are save to delete.

The way I've been using test data, as you may have seen in the code, is `getPath("my-test-file.json")`. And that getPath function is imported from `import {getPath} from "zui-test-data"`. The files themselves live in `packages/zui-test-data/data/`.

Feel free to add more files into that folder, then get at they with that `getPath` function.

Hope that is helpful. Thanks again.

jameskerr (approved):

PR #2597, opened 2022-11-22
Add ZNG and ZSON to file detection test

philrz:
While working on #2596 I noticed that `sample.zng` wasn't actually being used by any of our current tests. However, I've used that specific file in bug repros all the time and link to it often from issues, so I didn't want to propose removing it.

At the same time I found this "detect file types" test where I could put it to work. 😉  I threw in ZSON as well for good measure.


jameskerr (approved):
Thanks!

PR #2600, opened 2022-11-27
Watch for "Load Successful" pop-up after pool creation

philrz:
I was trying my hand at creating an e2e test by going off some existing tests and I was getting unexpected results. It turns out the example I was following may have had problems, as I found that if I replaced the `Import Complete.` in this diff with garbage text, the test still passed without complaint. I looked for other examples and found that changing to `expect(...).toBeVisible()` did what I think was intended, as now it chokes on garbage text but runs successfully when it sees the `Load Successful` pop-up. However, one snag remained in that the linter complained `error:  Expect must be inside of a test block`. I figured out how to adjust the linter rules to allow this, but if there's a better way to go about this, I'm all ears.

### DIFF
--- a/packages/e2e-tests/helpers/test-app.ts
+++ b/packages/e2e-tests/helpers/test-app.ts
@@ -8,7 +8,6 @@ import {
   Page,
   _electron as electron,
 } from "playwright-chromium"
-import {expect} from "@playwright/test"
 import env from "../../../src/app/core/env"
 import {itestDir} from "./env"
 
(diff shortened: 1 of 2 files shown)


packages/e2e-tests/helpers/test-app.ts
```diff
@@ -63,7 +64,7 @@ export default class TestApp {
     ])
 
     await chooser.setFiles(filepaths)
-    await this.mainWin.locator("text=Import Complete.").isVisible()
```
jameskerr, line 66: Looks like this might be better. You won't need expect.

```
await this.mainWin.getByText("Load Successful").waitFor()
```

https://playwright.dev/docs/api/class-locator#locator-wait-for

philrz: @jameskerr: Thanks for finding that! Indeed, I tried it out and it worked as advertised. I've pushed a commit to use that instead and backed out the linter changes I'd put in to permit expect.

jameskerr (approved):

PR #2601, opened 2022-11-29
New Pool Page

### LINKED ISSUE #2498: Allow specification of input format when the user knows it or auto-detect fails
In #1913, @jameskerr remarked:

> I'd love a dialog that asks the user which format their data is in to pass to `zed load`. I've got a csv that it thinks it's zson so it fails with a strange error.
>
> A possible workflow is:
>
> Import the file
> 1. If there is an auto detect error, prompt the user for the type of file
> 2. It would also be nice to give a summary of what the auto detect detected.

Since users may sometimes know their format in advance, I'd also recommend we offer them the ability to explicitly specify the format on the first try. For instance, https://github.com/brimdata/zed/issues/3865 describes the perils of trying to send certain JSON through the auto-detector. If a user works with this kind of JSON a lot, they'd surely appreciate a way to avoid the failure round-trip.

In conclusion, maybe the "prompt" in @jameskerr's list above could be some kind of pull-down list of the supported formats, with "Auto-detect" as the default. Then if the auto-detect fails, the same list could be offered again.

Also, the "summary of what the auto detect detected" exists today as something like this:

image

When looking at this while pretending to be a new user, I admit this could be somewhat confusing, since it's not immediately obvious that each line represents why it failed to auto-detect as that format.

jameskerr:
* The Load Data button on the toolbar has been removed. Instead users will use the dropzone form.
* You can create an empty pool.
* Buttons in the modals have been changed.
* Form elements in the preferences page have been changed. 
* Introduced the panes folder in the repo. 

CleanShot 2022-11-29 at 09 43 22@2x

CleanShot 2022-11-29 at 09 43 41@2x

Closes #2498


philrz:
While testing this branch at commit 93231e7 it seems i've found an issue. As shown in the attached video, now if I create and drop a pool from the CLI, the pool info page ends up showing a big-text "404". I went back to current `main` and with the same repro steps the pool info page gets replaced with the New Pool page, which was more what I'd expect.

https://user-images.githubusercontent.com/5934157/204646541-502d9bfe-c250-49fc-b88d-df525b4f74d0.mp4

philrz (approved):
I've tested out this branch and also pushed some commits of my own with the blessing of @jameskerr. To review the changes since I started doing that, look at https://github.com/brimdata/brim/compare/f2459bd...31a0e5b. Here's a summary:

* I caught this branch up with `main` a couple times to take advantage of new Zed backend functionality to handle load of certain formats via the API (https://github.com/brimdata/zed/pull/4229, https://github.com/brimdata/zed/pull/4235, https://github.com/brimdata/zed/pull/4246)
* "zeek" and "zjson" can now be selected as explicit input formats in which case they are wired via MIME type to specific readers in the Zed backend.
* I dropped "ndjson" as an explicit format in the load picker. Starting with https://github.com/brimdata/zed/pull/3555, Zed CLI tooling dropped explicit reference to NDJSON since both formats are supported under the name "JSON". While we could have kept the "ndjson" reference in the list and just wired it to the Zed JSON reader, I went with dropping it for consistency with `zed load`. (On a related topic, we could consider dropping "ndjson" from the list of Export formats as well, but the API _does_ recognize the "x-ndjson" MIME type for query responses, so I'm in favor of leaving that in place for user convenience.)
* I updated the text about which formats are auto-detectable and dropped the link to the doc that previously described importing Parquet and ZST from outside the app. With this branch, Parquet can now be imported by picking the format, and we can do the same for ZST once https://github.com/brimdata/zed/issues/4251 is completed. Therefore I'm planning to drop that doc in the new Zui docs site since it's no longer needed.

With all those changes in place, I've tested load of all the explicit formats and confirmed they all work as expected when selecting the matching format of input data file, and also that they respond with errors where expected when given mismatched input data. On the latter, some of the error messages from the backend could stand to be improved, but we can pursue that as a separate topic. My testing also confirmed that the picking of pool config like name, key, and sort order are working as expected.

As @jameskerr and I discussed when this PR was first opened, it'd be great to have some e2e tests here. That said, until #2605 gets addressed, e2e tests that trigger pop-ups (such as loading to pools as would happen here) are going to be glitchy. Therefore if we'd like to get these changes in front of the Zui Insiders users beta-style I'd be ok with getting it merged then circle back and add the e2e tests after #2605 is done.

src/pages/pools/new.tsx
```diff
@@ -1,71 +1,193 @@
...
+          <b>Auto-detectable formats</b>:{" "}
+          <Link href="https://tools.ietf.org/html/rfc4180">CSV</Link>,{" "}
+          <Link href="https://www.json.org/json-en.html">JSON</Link>,{" "}
+          <Link href="http://ndjson.org/">NDJSON</Link>,{" "}
+          <Link href="https://docs.zeek.org/en/current/log-formats.html#zeek-tsv-format-logs">
+            Zeek
+          </Link>
+          , <Link href={links.ZED_DOCS_FORMATS_ZJSON}>ZJSON</Link>,{" "}
+          <Link href={links.ZED_DOCS_FORMATS_ZNG}>ZNG</Link>, and{" "}
+          <Link href={links.ZED_DOCS_FORMATS_ZSON}>ZSON</Link>.<br></br>
```
philrz, line 185: @jameskerr: I'm guessing this use of `<br>` is inappropriate, so feel free to overhaul my formatting. 😂

jameskerr:
I've updated the 404 page to look like so:

CleanShot 2022-12-07 at 14 19 02@2x

jameskerr:
Phil, I made the default pool key "ts" on the new pool form. If you ingest zeek logs or a pcap and don't specify the pool key as "ts", you won't be able to see the histogram because the pool won't have a "span".

If you just drag a file into the left sidebar, the pool key will be set as "ts" and the histogram will appear.

Issue #2603, opened 2022-12-01
Node processes left behind after Cmd-q exit


philrz:
Repro is with Brim commit 0d33807.

Repro steps:
1. Start the app in Dev mode via `yarn start`
2. Quit the app via Cmd-q
3. Kill the `yarn start` via Ctrl-c in the Terminal window where it was launched (if applicable)
4. Check the process table via `ps` to look for Brim-related processes

The first video below was taken at commit 6b58531 and is a baseline of how things used to be when healthy. As you can see, after I hit Cmd-q, the Terminal window shows the `yarn start` is still running. I therefore hit Ctrl-c on it. Checking `ps` shows there's no more Brim-related processes running, as I'd expect.

https://user-images.githubusercontent.com/5934157/205143925-bad0f56e-c648-48f9-85a1-e701f24f1978.mp4

Next comes the repro video taken at the next commit 0d33807, which is associated with the changes in #2584. Here we see that when I hit Cmd-q, the `yarn start` I had running in the Terminal window exits automatically by itself. Checking `ps` shows Brim-related processes have been left running. This is is the bug.

https://user-images.githubusercontent.com/5934157/205144341-8a0df709-ec9f-4413-83ae-5c5beeabe210.mp4

Finally, at that same commit 0d33807, I've also observed that if I skip the Cmd-q and go immediately to hitting Ctrl-c in the Terminal window, there's no processes left behind.

https://user-images.githubusercontent.com/5934157/205144917-feedc846-4570-41e6-a157-96dfaa86c1cd.mp4

philrz:
Verified in Brim commit c2a1db4.

As shown in the attached video, now when we Cmd-q to quit the app, lingering processes are no longer visible in the `ps` output.

https://user-images.githubusercontent.com/5934157/206017633-d2ed9b1b-d557-4da2-bd8e-005af1010b44.mp4

Thanks @jameskerr!

PR #2607, opened 2022-12-06
Cleanup Dev Processes

jameskerr:
I thought that process.exit(0) would terminate any subprocesses that were started from the main process. That is not the case with node.

So now, each process that gets started in the start script, is killed if the app is quit.


philrz (approved):
I tested on the branch at commit 3ead6bb with the repro steps from #2603 and the processes were no longer left behind, so, that's a functional 👍 from me.

PR #2608, opened 2022-12-06
Add Zeek and ZJSON as export formats

philrz:
Since https://github.com/brimdata/zed/pull/4246 enabled Zeek as an exportable format over the API, I've added it to the set of permitted export formats in the app.

Since ZJSON was already working in the API and #2601 has it lined up to be an explicit import format, I added that here as an export format as well for consistency.

I updated some of the locators in the e2e test to the improved syntax @jameskerr showed me, which was in one case necessary to make the updated tests pass. I'll point it out in a PR comment.

### DIFF
--- a/packages/e2e-tests/tests/export.spec.ts
+++ b/packages/e2e-tests/tests/export.spec.ts
@@ -9,6 +9,8 @@ const tempDir = os.tmpdir()
 const formats = [
   {label: "zng", expectedSize: 3692},
   {label: "zson", expectedSize: 15137},
+  {label: "zjson", expectedSize: 18007},
+  {label: "zeek", expectedSize: 9697},
   {label: "json", expectedSize: 13659},
   {label: "ndjson", expectedSize: 13657},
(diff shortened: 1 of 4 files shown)


packages/e2e-tests/tests/export.spec.ts
```diff
@@ -45,8 +47,8 @@ test.describe("Export tests", () => {
         .locator('#app-root button:above(:text("Export"))')
         .first()
         .click()
-      await app.mainWin.locator(`text=${label}`).first().click()
```
philrz, line 48: This was the one that needed revision. The text selector's behavior of searching for substrings (per [docs](https://playwright.dev/docs/selectors#text-selector)) was causing the test to click the "zjson" entry when it was trying to find the one for "json".

jameskerr: much better

jameskerr (approved):
Thank you Phil!

PR #2610, opened 2022-12-07
Fix Those Buggy Windows

### LINKED ISSUE #2589: On Windows, "TypeError: Object has been destroyed" when repeatedly opening About
Repro is with [Zui Insiders v0.30.1-116](https://github.com/brimdata/zui-insiders/releases/tag/v0.30.1-116), though I found through a separate exercise that this problem seemed to first appear in [Zui Insiders v0.30.1-53](https://github.com/brimdata/zui-insiders/releases/tag/v0.30.1-53).

The attached video was taken on a fresh Windows 2019 Server VM on AWS. After opening Zui Insiders v0.30.1-116 for the first time, I open the **About** window, close it, then open it again. On the re-open the following error message pops up:

image

On some repros I had to close & re-open it an additional time to trigger the problem. As the video also shows, there's no sign of additional error messages in the DevTools console. Also, once the failure is triggered, I found my first attempt to **Exit** from the app seemed to do nothing, so I had to click **Exit** a second time to get the window to actually close. Finally, peeking in **Task Manager** showed that the Zed process and multiple Zui Insider processes were still running at that point, which is also unexpected.

https://user-images.githubusercontent.com/5934157/201191867-b1f9bbcb-d96c-4c29-a3b2-7ac3fd928561.mp4

I did a binary search that seemed to show [Zui Insiders v0.30.1-53](https://github.com/brimdata/zui-insiders/releases/tag/v0.30.1-53) was where this problem first appeared. With v0.30.1-52 and earlier I could open & close the **About** window 10 times in a row without complaint. Looking at the Brim commits that came shortly before the v0.30.1-53 release, #2496 happened to be a big PR that did include some specific changes related to the **About** window, but not sure.

I also confirmed this problem does _not_ seem reproducible on macOS or Linux.

jameskerr:
The bug:

When creating a new window, the app took these steps:
1. Load the window
2. Setup a "closed" listener to delete the reference to the window

If you closed the window quickly before it finished loading, the first step would throw an error. The second step would never happen. This lead to us keep around a reference to a destroyed window. 

Any other places in the code that tried to access anything on that window would get the "Object is Destroyed" message.

The fix is to do this:

1. Setup the "closed" listener
2. Load the window

Now, no matter what, when the window is closed, we will remove the reference to it. Just like the electron docs instruct: https://www.electronjs.org/docs/latest/api/browser-window#event-closed

If the load fails, it will throw an error. I've gone to all the places were windows are created and wrapped them in a try catch so that we get helpful error messages.

Fixes #2589

### DIFF
--- a/src/js/electron/windows/about-window.ts
+++ b/src/js/electron/windows/about-window.ts
@@ -2,6 +2,7 @@ import {WindowName} from "../windows/types"
 import {ZuiWindow} from "./zui-window"
 
 export class AboutWindow extends ZuiWindow {
+  persistable: false
   name: WindowName = "about"
   options = {
     width: 360,
--- a/src/js/electron/windows/window-manager.ts
+++ b/src/js/electron/windows/window-manager.ts
@@ -76,7 +76,7 @@ export class WindowManager {
     if (win) {
       win.state = state
     } else {
-      log.error("No Window Found with id: ", id)
+      log.error("window not found: ", id)
     }
   }
 
@@ -88,16 +88,15 @@ export class WindowManager {
 
   private async register(win: ZuiWindow) {
     this.windows[win.id] = win
-    win.ref.on("closed", () => {
-      console.log("THIS WINDOW WAS CLOSED")
-      this.unregister(win)
-    })
+    win.ref.on("closed", () => this.unregister(win))
     await win.load()
+    log.debug(`window registered:`, {id: win.id, name: win.name})
     return win
   }
 
   private unregister(win: ZuiWindow) {
     delete this.windows[win.id]
+    log.debug(`window unregistered:`, {id: win.id, name: win.name})
     // whenever a window is closed in Linux or Windows check if 'hidden' window is last
     // open, and if so tell it to close so the rest of the app will shutdown
     if (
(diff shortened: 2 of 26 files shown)


src/js/electron/windows/about-window.ts
```diff
@@ -2,6 +2,7 @@ import {WindowName} from "../windows/types"
 import {ZuiWindow} from "./zui-window"
 
 export class AboutWindow extends ZuiWindow {
+  persistable: false
```
jameskerr, line 5: Don't persist the about window

src/js/electron/windows/window-manager.ts
```diff
@@ -57,25 +76,27 @@ export class WindowManager {
...
+    return this.where((w) => w.persistable && !w.destroyed).map((w) =>
+      w.serialize()
     )
   }
 
   private async register(win: ZuiWindow) {
     this.windows[win.id] = win
-    await win.load()
     win.ref.on("closed", () => this.unregister(win))
+    await win.load()
```
jameskerr, line 92: This is the bug

philrz (approved):
I've tested this branch at commit 79b15e8 on a Windows 2019 Server VM on AWS. On this VM I consistently repro'ed #2589 before the changes in this branch, and now the issue can no longer be reproduced. Therefore a functional 👍  from me.

Issue #2569, opened 2022-10-23
how to build from sorce


amirseyedian:
Im using linux on an arm64 system and there is no binary supporting it. I assume I have to build the app from source unfortunately there is no proper guide on the wiki and readme.
there is no makefile either I think I can compile with yarn but running into unrelated issues.
A step by step guide on building from source code would be great.

philrz:
@amirseyedian: Maybe it didn't stand out well in the top-level [README](https://github.com/brimdata/brim/blob/main/README.md), but as linked from there, the [CONTRIBUTING](https://github.com/brimdata/brim/blob/main/CONTRIBUTING.md) doc has the info on developing/building from source. Knowing you're about to follow that guide, I just revisited the steps on a Linux VM and did find one small mistake in that doc if your goal is to make an installable package, which is that the `yarn release` shown there is currently wired to compile on macOS, but executing the underlying command `yarn electron-builder` instead will generate installers for Linux when running on Linux. When I did this just now I was admittedly on `x86_64` and not `arm64`, and I don't think I've heard reports confirming if others have tried building for that platform, so please do report back if you're successful or hit any snags and we can update the docs to reflect anything else learned along the way.

PR #2611, opened 2022-12-08
Improvements to CONTRIBUTING doc

philrz:
Community issue #2569 led me to revisit our user-facing docs for creating release artifacts. I thought they might be in some ways confusing or incomplete when viewed by an outsider. Specifically, the `yarn release` that was referenced previously was wired up to run on MacOS, which may have suited our internal needs but might confuse a user such as the one in #2569 running on a non-Mac OS. Since Electron Builder defaults to building for the detected OS, I've changed it over to that.

We also only had code signing details for MacOS (and the ones we had needed a little adjustment), so I've included the ones for Windows as well.

I fixed a couple typos, fixed some links, and improved some other wording while I was in there.

I'll add PR comments to further point out a couple specific changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,8 @@ Install these dependencies:
 
 1. [Node](https://nodejs.org/en/download/package-manager/) - the version specified in the `.node-version` file at the root folder.
 2. [Yarn](https://yarnpkg.com/) - a package manager for installing dependencies and starting Brim in dev mode.
-3. [Go](https://go.dev/doc/install) - to compile some Zed dependencies.
+3. [Go](https://go.dev/doc/install) - to compile some [Zed](https://zed.brimdata.io/) dependencies.
+4. Typical command line tools, such as `make`, `unzip`, and `curl`
 
 Then clone the repo, install the node modules, and start the app.
 
@@ -21,7 +22,7 @@ yarn
 yarn start
 ```
 
-When a file is changed, it will be recompiled it and reload the app.
+When a file is changed, it will be recompiled and the app will automatically reload.
 
 On subsequent updates, `git pull` and `yarn`.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -10,7 +10,8 @@ Install these dependencies:
 
 1. [Node](https://nodejs.org/en/download/package-manager/) - the version specified in the `.node-version` file at the root folder.
 2. [Yarn](https://yarnpkg.com/) - a package manager for installing dependencies and starting Brim in dev mode.
-3. [Go](https://go.dev/doc/install) - to compile some Zed dependencies.
+3. [Go](https://go.dev/doc/install) - to compile some [Zed](https://zed.brimdata.io/) dependencies.
+4. Typical command line tools, such as `make`, `unzip`, and `curl`
```
philrz, line 14: I called these out because I found that on a minimal install of Ubuntu or Windows, some of these tools were not present. The debug logs generated when the `yarn` commands fail are fairly clear in explaining what's missing, but I figure if we mention these up front it might save the reader some debug cycles.

package.json
```diff
@@ -30,7 +30,7 @@
     "e2e": "yarn workspace e2e-tests test",
     "format": "prettier '**/*.{js,ts,tsx,scss}' --write --loglevel warn",
     "prerelease": "yarn build",
-    "release": "electron-builder --mac",
+    "release": "electron-builder --publish never",
```
philrz, line 33: In addition to dropping the `--mac` so it builds for the detected OS, the `--publish never` prevents it from seeking out a GitHub token env var to publish with and failing when it's not found. Our Actions Workflows generally handle the job for the GitHub-published ones.

jameskerr (approved):
This is great. Thanks Phil.

PR #2612, opened 2022-12-08
Extend timeout on exports

jameskerr:
When exporting data, we don't get control messages. This means it could be a long time before the first byte is sent back. By default we have a 60 second timeout. A user recently hit this limit and could not export their data. This PR makes the timeout Infinity for exports.

### DIFF
--- a/packages/zealot/package.json
+++ b/packages/zealot/package.json
@@ -52,7 +52,7 @@
     "fkill": "7.2.1",
     "jest": "^27.3.1",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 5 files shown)


philrz:
I've tested this branch at commit 9405352 using the repro steps from #2613. As shown in the attached video, the **Export** now completes successfully even after a 5 minute delay.

https://user-images.githubusercontent.com/5934157/206544216-1ff008b6-37d6-4dc1-907b-a78656103cf5.mp4

That's a functional 👍 for me.

PR #2622, opened 2022-12-16
Tidy the format selections for load and export

philrz:
As I've been working a lot lately with adding/testing formats for load & export, I've noticed how the Zed docs have been pretty tidy in how formats are typically referenced in alphabetical order and presented with capital letters where appropriate. The one exception is the CLI tooling where they're specified in lowercase, but that seems to be par for the course when it comes to CLI tools. Therefore I'm proposing we go the ordered/caps route in the app so it's consistent with what users see in the docs.


jameskerr (approved):
Ok, sounds good to me.

philrz:
Thanks @jameskerr!

Issue #2606, opened 2022-12-05
No error message displayed on CSV load exception


philrz:
Repro is with Brim commit 96a0951 which uses Zed commit `22f640e`. This issue feels very similar to the existing JSON-related one in #2554.

A community zync user reported a problem reading CSV with Zui. In their own words:

> Zui did not report any problem. It just stopped the CSV import without any message at all. I noticed because I compared `count()` to the `wc`.
> maybe simply send a more comprehensive error message in Zed/Zui when CSV import fails before the EOF? line number and column number with last value parsed?

Indeed, as shown in the following video, reading the file [bad.csv](https://github.com/brimdata/brim/files/10158482/bad.csv) using CLI tooling does show an error, but the app just shows the data that parsed ok without any mention of the error. This is a data correctness hazard since it gives the user the impression they're working with all their data when it's actually truncated.

https://user-images.githubusercontent.com/5934157/205760538-de1072b7-47da-4c75-82c3-0a9f72d38488.mp4

As we can see in Wireshark, the error message was returned from the backend, but just didn't make it all the way to the user.

![image](https://user-images.githubusercontent.com/5934157/205761000-626242a5-5305-4479-9fee-f4577b8aba24.png)

philrz:
Verified in Zui commit a372438.

As shown in the attached video, now the user is shown a warning when this partial load occurs, then they can click this for further detail.

https://user-images.githubusercontent.com/5934157/217349576-0e249e38-acb5-4394-b426-ee3f60bb90e8.mp4

It's been pointed out that the "Load successful" pop-up does represent a contradictory message alongside the "Ingest failed with warnings" message, so #2660 has been opened to improve on that.

Thanks @nwt!

Issue #2591, opened 2022-11-16
Add Parquet as an Export format


philrz:
At the time of the opening of this issue, Brim is at commit 9d0a742.

Since the backend already supports it, like we did for ZSON with #1486, it should be trivial to add Parquet as an Export format. However, as I was about to impulsively do this, I remembered that the backend still has this limitation:

```
$ zq -version
Version: v1.2.0-130-g8dd27d98

$ zq -f parquet 'fuse' ./src/test/shared/data/sample.tsv 
union type unsupported
```

Since union types often show up in Zed formats (particularly as a side effect of things like applying `fuse`) my inclination is to wait on an improvement in the backend before we add it as an **Export** option.

For now, I've got a branch https://github.com/brimdata/brim/tree/parquet-export where I've got it working for the happy path. Separately I opened up https://github.com/brimdata/zed/issues/4215 to track what I'm observing in the API interactions that are currently preventing the app from reporting the failure scenarios to the user.

philrz:
Verified in Zui commit a28b421.

**Export Results** now offers Parquet as a format option, and it succeeds as long as the data fits within Parquet's limitations. The attached video shows the load and export of such a file and opening the exported file in a separate tool that supports Parquet. If the data for some reason cannot fit within Parquet's limitations, the user will see the appropriate error message when the export fails thanks to the changes linked to https://github.com/brimdata/zed/issues/4215 and the changes in #2751.

https://github.com/brimdata/zui/assets/5934157/a007fcd4-b4a9-4b8f-a771-a141e8d47496

Issue #2554, opened 2022-09-24
No error message displayed on json load exception (regression after GA Brim v0.28.0)


burpheart:
Before opening a new issue, please make sure you've reviewed the troubleshooting guide:
https://github.com/brimdata/brim/wiki/Troubleshooting

**Describe the bug**
No warning message displayed on json load exception
Version 0.26 normally indicates a parsing error
**To Reproduce**
Loads a multi-line json file that has lines that cannot be parsed properly.
Brim only parses the data before the exception line, and there is no error message.

**Expected behavior**
Display a warning message about an engine parsing error.

**Screenshots**

**Desktop Info**
 - OS: Windows
 - Brim Version: 0.31.0

philrz:
Here's a repro at Brim commit 6eaaa54 which uses Zed commit `28f95a3`.

The attached test data [threelines.ndjson.gz](https://github.com/brimdata/brim/files/9658986/threelines.ndjson.gz) used in the repro has a syntax error on the second line. Indeed, standalone Zed CLI tooling recognizes this.

```
$ gzcat threelines.ndjson.gz 
{"one": 1}
{"two : 2}
{"three": 3}

$ zq -version
Version: v1.2.0-64-g28f95a3c

$ zq -z threelines.ndjson.gz 
threelines.ndjson.gz: parse error: string literal: unescaped line break

$ echo $?
1
```

However, as the community user identifies, the app currently silently imports everything up to the "bad" line and drops everything after it.

https://user-images.githubusercontent.com/5934157/192616043-22b34117-063f-4675-a74c-178320bb0950.mp4

Sniffing the traffic to the Zed service confirms the warning is being returned, but it's just not being surfaced in the app.

![image](https://user-images.githubusercontent.com/5934157/192616564-cee0fc99-947a-482e-8ec5-6aa5345f9184.png)

I also confirmed the user's statement that a warning was provided back in GA Brim tagged `v0.26.0` as shown in the attached video.

https://user-images.githubusercontent.com/5934157/192616187-e26c0a51-3a39-4179-a933-487a820b6cd4.mp4

I separately confirmed that the last time the warning was present was in GA Brim tagged `v0.28.0` and then the warning was gone in `v0.29.0`. I attempted to do a binary search between these to find the precise commit where it regressed, but I had difficulty compiling at old commits.

philrz:
Verified in Zui commit a372438.

[...]

PR #2625, opened 2022-12-19
Add load/export support for Arrow IPC Stream

philrz:
https://github.com/brimdata/zed/pull/4252 added support for loading/querying in Arrow IPC Stream in Zed, so here we add support for it in the app as well.

### DIFF
--- a/src/js/flows/exportResults.ts
+++ b/src/js/flows/exportResults.ts
@@ -25,7 +25,7 @@ function cutColumns(program, columns) {
 
 export function prepareProgram(format, program, columns) {
   let p = cutColumns(program, columns)
-  if (format === "csv") p += " | fuse"
+  if (format === "csv" || format === "arrows") p += " | fuse"
   return p
 }
 
--- a/packages/e2e-tests/tests/export.spec.ts
+++ b/packages/e2e-tests/tests/export.spec.ts
@@ -7,13 +7,14 @@ import fsExtra from "fs-extra"
 
 const tempDir = os.tmpdir()
 const formats = [
-  {label: "ZNG", expectedSize: 3744},
-  {label: "ZSON", expectedSize: 15137},
-  {label: "ZJSON", expectedSize: 18007},
-  {label: "Zeek", expectedSize: 9772},
+  {label: "Arrow IPC Stream", expectedSize: 46512},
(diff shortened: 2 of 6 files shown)


src/js/flows/exportResults.ts
```diff
@@ -25,7 +25,7 @@ function cutColumns(program, columns) {
 
 export function prepareProgram(format, program, columns) {
   let p = cutColumns(program, columns)
-  if (format === "csv") p += " | fuse"
+  if (format === "csv" || format === "arrows") p += " | fuse"
```
nwt, line 28: Maybe add Parquet while you're here?
```suggestion
  if (format === "arrows" || format === "csv" || format === "parquet") p += " | fuse"
```

nwt (approved):

src/js/flows/exportResults.ts
philrz: @nwt: As much as I'd love to, I've been holding off on adding support for Parquet export in the app. As described in #2591 and https://github.com/brimdata/zed/issues/4215, failures mid-export currently aren't being shown to the user and so there's a hazard of the user having incomplete output and not knowing about it. Granted, this seems to be a general problem with API interactions for both query and load and hence affects other formats as well (#2606, #2554). But in testing with various data I found it was way too easy to trigger the "union type unsupported" with Parquet export in particular, so that's been my rationale for holding it back. I know you've been doing more work with Parquet due to the new Arrow library so I've been waiting for the dust to settle on that before revisiting the topic, since I wondered if maybe the union support might magically appear. 😄  If you feel strongly about getting Parquet export in sooner I could surely change my tune since there's no perfect answers here. On the whole I see this topic of absent error messages during API interactions as a Zui v1.0 release blocker, so it will definitely not be forgotten as a priority.

nwt: > As much as I'd love to, I've been holding off on adding support for Parquet export in the app.

That's fine. I don't feel strongly about this.

> I know you've been doing more work with Parquet due to the new Arrow library so I've been waiting for the dust to settle on that before revisiting the topic, since I wondered if maybe the union support might magically appear.

I'm afraid it won't since the Parquet format itself doesn't support unions.

PR #2633, opened 2023-01-12
Start using macos-12 for builds

philrz:
I noticed our Zui Insiders builds have been failing for the past few days. The top-level symptom of the breakage looked like this:

```
Run maxim-lobanov/setup-xcode@v1
Switching Xcode to version '13.0'...
Available versions:
┌─────────┬──────────┬─────────────┬─────────────┬────────┬──────────────────────────────────┐
│ (index) │ version  │ buildNumber │ releaseType │ stable │               path               │
├─────────┼──────────┼─────────────┼─────────────┼────────┼──────────────────────────────────┤
│    0    │ '14.2.0' │   '14C18'   │    'GM'     │  true  │  '/Applications/Xcode_14.2.app'  │
│    1    │ '14.1.0' │  '14B47b'   │    'GM'     │  true  │  '/Applications/Xcode_14.1.app'  │
│    2    │ '14.0.1' │  '14A400'   │    'GM'     │  true  │ '/Applications/Xcode_14.0.1.app' │
│    3    │ '14.0.0' │  '14A309'   │    'GM'     │  true  │  '/Applications/Xcode_14.0.app'  │
│    4    │ '13.4.1' │  '13F100'   │    'GM'     │  true  │ '/Applications/Xcode_13.4.1.app' │
│    5    │ '13.4.0' │  '13F17a'   │    'GM'     │  true  │  '/Applications/Xcode_13.4.app'  │
│    6    │ '13.3.1' │  '13E500a'  │    'GM'     │  true  │ '/Applications/Xcode_13.3.1.app' │
│    7    │ '13.2.1' │  '13C100'   │    'GM'     │  true  │ '/Applications/Xcode_13.2.1.app' │
│    8    │ '13.1.0' │ '13A10[30](https://github.com/brimdata/zui-insiders/actions/runs/3903511327/jobs/6667997602#step:9:32)d'  │    'GM'     │  true  │  '/Applications/Xcode_13.1.app'  │
└─────────┴──────────┴─────────────┴─────────────┴────────┴──────────────────────────────────┘
Error: Could not find Xcode version that satisfied version spec: '13.0'
```

The Actions workflows in the Zui Insiders repo happened to be pointing at `macos-latest` whereas the ones in the Brim repo point at `macos-11`. After some digging I found [this blog article](https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/) that explains how they've been gradually transitioning from 11 to 12, so it looks like we became subject to this just a few days ago such that our builds were now failing on the macOS 12 runners that lacked the Xcode 13.0 our automation requested. Per https://github.com/actions/runner-images/issues/6384 and their [docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), it looks like they finalized the transition yesterday such that `macos-latest` is now a hard pointer to just macOS 12.

Since I was already taking the hit of fixing the builds and doing some smoke testing, rather than hard-pointing Zui Insiders at `macos-11` and making things work again the way they used to, I decided to look at making the transition to `macos-12` and that seemed to go ok. I found I was able to drop the hard pointer we had to `xcode-version: "13.0"` since the newer runners have Xcode 14.0.1 by default, so it's plenty new to have the `notarytool`.

Once I had Brim's build automation pointing at `macos-12` and no longer had the Xcode problem, that revealed another problem waiting behind that, which was the version of Electron Builder we'd been pointing at was still dependent on Python2, which disappeared in macOS 12. [This comment](https://github.com/electron-userland/electron-builder/issues/6726#issuecomment-1070632761) indicated this could be fixed by moving to Electron Builder 23, so I've now pointed us at the latest GA dot release and it does indeed seem to do the trick. When I ran `yarn` after that change a bunch of other dependencies also got updated as well, but I assume that's expected.

In this PR I also updated our support statement to reflect where our macOS automation now runs. In terms of smoke testing on an older macOS, I only went as far as to reinstall and test on Big Sur 11, so I've dropped the statement we had before about Mojave 10.14. It might still work on Mojave for all I know, but I'd like to drop the overhead.

In addition to passing all the regular/e2e tests, I also did some quick/manual smoke tests on Windows/Linux/macOS just to confirm there were no unexpected side effects from the new Electron Builder. Everything seemed to check out fine.

Once this PR merges, I'll update the Zui Insiders workflows to point to `macos-12` rather than `macos-latest`, which should change nothing in the moment but better prepares us for the future. I know from prior runner transitions that when we're pointing at an explicit version they give us several months of lead time before they remove support for an old version, so that would allow us to do the testing and plan the transition when we have cycles rather than have stuff spontaneously start breaking, not be sure why, and have to react in the moment.

### DIFF
--- a/.github/actions/build-zui/action.yml
+++ b/.github/actions/build-zui/action.yml
@@ -33,12 +33,6 @@ runs:
       run: sudo apt-get install -y rpm
       shell: bash
 
-    - name: Setup Xcode 13 # This lets us use the new notarytool cli
-      if: runner.os == 'macOS'
-      uses: maxim-lobanov/setup-xcode@v1
-      with:
-        xcode-version: "13.0"
-
     - name: Set developer ID certificate in keychain
       if: runner.os == 'macOS'
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
   release:
     strategy:
       matrix:
-        platform: [macos-11, ubuntu-20.04, windows-2019]
+        platform: [macos-12, ubuntu-20.04, windows-2019]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout Zui
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11, ubuntu-20.04, windows-2019]
+        os: [macos-12, ubuntu-20.04, windows-2019]
     steps:
       - run: git config --global core.autocrlf false
       - uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
   release:
     strategy:
       matrix:
-        platform: [macos-11, ubuntu-20.04, windows-2019]
+        platform: [macos-12, ubuntu-20.04, windows-2019]
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Checkout Zui
--- a/docs/Supported-Platforms.md
+++ b/docs/Supported-Platforms.md
@@ -23,7 +23,7 @@ recommendations on which to run Brim:
   - Windows 10 or newer
   - Windows Server 2019 or newer
 - macOS
-  - macOS Big Sur 11.6.8 or newer (see [below](#hardware) for hardware considerations)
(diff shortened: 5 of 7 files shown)


jameskerr (approved):
This makes sense to me Phil. Thank you for the research and the testing!

I saw this blog post a while back. It seems like sometime in the next year it'd be worth it to starting using the official electron build and distribution tools. 
https://www.electronjs.org/blog/forge-v6-release

philrz:
Thanks @jameskerr. I've opened #2634 as a reminder for us to look at moving to those official tools.

PR #2648, opened 2023-02-02
Quit the App in a Sane Way

### LINKED ISSUE #1817: Tabs & Session History disappear after closing the last Brim window
Repro is with Brim commit af1f358 via macOS beta artifact https://storage.googleapis.com/brimsec-releases/brim/v0.25.0-prerelease-af1f3581.0/macos/Brim-0.25.0-prerelease-af1f3581.0.dmg. This issue was originally reported by a community user.

While the **File > New Window** option exists in Brim, it seems like many users may only ever open Brim by clicking the icon in their OS and then close/quit from that one Brim window when they're done. In this regard, the subtle differences between "quitting the app" and "closing the last window" might elude these users.

The setup for both of the following two videos is the same. I've imported [ifconfig.pcapng.gz](https://github.com/brimdata/brim/files/7131110/ifconfig.pcapng.gz) (after uncompressing) and opened the same pool in multiple tabs, executing Zed query `curl` in one tab and `count() by _path` in the other.

As shown in the attached video, If I then select **File > Quit** to to close the app, then when I re-open the app, all my tabs are right where I left them.

https://user-images.githubusercontent.com/5934157/132566032-82aeb153-3323-467d-a143-a0df089e5897.mp4

OTOH, if I click the "red dot" on macOS, when I click the icon in the dock to "re-open Brim", my tabs are gone.

https://user-images.githubusercontent.com/5934157/132566111-3b93a0da-69f5-45a8-8d92-088a056fd0a5.mp4

Now, as a Brim insider, I understand what's going on here: The first video shows "quitting the app" and maintaining all my state (which might have involved multiple windows and all their tabs, if I had more open), while the second is "closing the last window" and clearing all state related to it (effectively bringing me back to a clean slate). As a user, if I'd had multiple windows open with different tabs on each, the idea of clicking the "red dot" to close _one_  window and having its tabs disappear is probably not shocking. OTOH, if I only ever use Brim with a single window, the subtle difference between "quit" and "close" (and the tiny black dot under the Brim icon in my dock) might be lost on me. Indeed, the community user who reported this bug effectively fell into that trap and thought they were seeing a bug. Once I explained all of the above, they now understood what was going on and may change their workflow to be more careful about using "quit". However, the fact they had to hear a lengthy explanation to be convinced it's not a bug implies we could do better.

My first thought is that maybe if the user is closing the _last_ window, we could pop up some kind of warning along the lines of:

> You're closing the last window in Brim, which will close your <N> tabs. Are you sure you want to do this?
> (Select File > Quit to close the application while maintaining all window/tab settings)

If we wanted to get fancy, we could also:

1. Not pop this up if the only tabs that are open are "blank" (no pool selected and no query entered)
2. Have one of those "Don't show me this warning again" checkboxes

Personally I'm not advocating doing fancy things, but figured I'd write them down for completeness.

Also, the community user said they observed a similar effect on Windows, so when we decide to pursue this topic, let's make sure the fix works on all supported platforms.

jameskerr:
The logic for closing the last window of zui is now as follows:

1. If mac, hide the last window, don't close it. Upon activation, just shows that window again.

2. If not mac, abort the normal close window event and initiate an app quit event.

The app quit event saves the session, then closes all the windows and shuts down the main process.

This will fix much of the confusion people have had with their tab state being removed.

Closes #1817

### DIFF
--- a/src/js/electron/windows/window-manager.ts
+++ b/src/js/electron/windows/window-manager.ts
@@ -7,11 +7,14 @@ import {State} from "src/js/state/types"
 import {ZuiWindow} from "../windows/zui-window"
 import {SerializedWindow, WindowName, WindowsState} from "../windows/types"
 import {createWindow, deserializeWindow} from "./create"
+import {EventEmitter} from "events"
 
-export class WindowManager {
+export class WindowManager extends EventEmitter {
   private windows: WindowsState = {}
 
-  constructor(private session?: SessionState | null | undefined) {}
+  constructor(private session?: SessionState | null | undefined) {
+    super()
+  }
 
   async init() {
     if (this.session) {
@@ -32,6 +35,10 @@ export class WindowManager {
 
   get count(): number {
     return this.all.length
+  }
+
+  get visible() {
+    return this.all.filter((w) => w.name !== "hidden")
   }
 
   create(name: WindowName) {
@@ -69,6 +76,7 @@ export class WindowManager {
 
   private async register(win: ZuiWindow) {
     this.windows[win.id] = win
+    win.ref.on("close", (e) => this.emit("window-will-close", e))
     win.ref.on("closed", () => this.unregister(win))
     await win.load()
     log.debug(`window registered:`, {id: win.id, name: win.name})
@@ -78,15 +86,6 @@ export class WindowManager {
   private unregister(win: ZuiWindow) {
     delete this.windows[win.id]
     log.debug(`window unregistered:`, {id: win.id, name: win.name})
-    // whenever a window is closed in Linux or Windows check if 'hidden' window is last
-    // open, and if so tell it to close so the rest of the app will shutdown
-    if (
-      process.platform !== "darwin" &&
-      this.count === 1 &&
-      this.all[0].name === "hidden"
-    ) {
-      this.all[0].ref.close()
-    }
   }
 
   private getNextDimensFor(name: WindowName) {
--- a/src/js/electron/windows/window-manager.ts
+++ b/src/js/electron/windows/window-manager.ts
@@ -74,6 +74,14 @@ export class WindowManager extends EventEmitter {
     )
   }
 
+  get singleHidden() {
+    return this.visible.length === 1 && !this.visible[0].ref.isVisible()
+  }
+
+  unhideAll() {
+    this.visible.forEach((win) => win.ref.show())
+  }
+
   private async register(win: ZuiWindow) {
     this.windows[win.id] = win
     win.ref.on("close", (e) => this.emit("window-will-close", e))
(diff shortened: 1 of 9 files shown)


src/js/electron/windows/window-manager.ts
```diff
@@ -78,15 +86,6 @@ export class WindowManager {
...
     log.debug(`window unregistered:`, {id: win.id, name: win.name})
-    // whenever a window is closed in Linux or Windows check if 'hidden' window is last
-    // open, and if so tell it to close so the rest of the app will shutdown
-    if (
-      process.platform !== "darwin" &&
-      this.count === 1 &&
-      this.all[0].name === "hidden"
-    ) {
-      this.all[0].ref.close()
-    }
```
jameskerr, line 89: This is now handled in the window-events file.

philrz:
Based on my tests with these changes in a Dev build, it seems like things are mostly right on macOS, but still not working as expected on Linux and Windows. Details below.

First, early in testing with Dev builds I kept getting foiled by the fact that the app kept auto-updating itself to old Brim and hence screwing up my tests. Therefore I made a new branch [app-quit-higher-version](https://github.com/brimdata/brim/tree/app-quit-higher-version) based on the branch from this PR with the only additional change being upping the version string in `package.json` to `0.32.0` to make it stop doing that. 😛

The video below shows what I see on macOS. After I close the last window, the dot remains under the app as evidence that it's still actually running it. If I re-click it, the last window comes back with the tabs and history intact. The same is true if I right-click and select **Quit** and then relaunch. The only finding that might be of worthy of review is if I click **File > New Window** after closing the last window and it's still in the "alive & showing dot" state. In _that_ case the tabs & history in the last window do not come back. The only reason I thought to look for this was because in https://github.com/brimdata/brim/issues/1817#issuecomment-931805100 @jameskerr expressed an interest in restoring stuff when **New Window** is selected.

https://user-images.githubusercontent.com/5934157/216851482-8cf30f36-059c-46a1-ae1b-595a15ecd0bb.mp4

On Linux and Windows, the change doesn't seem to be having the intended effect. Closing the last window causes the app to completely exit as usual. However, when I relaunch the app, the tabs and history that were there before are gone.

https://user-images.githubusercontent.com/5934157/216852284-0644aa31-79f0-4602-a271-e0ed0a3dd1ac.mp4

My test on Windows showed the same result as on Linux, but I'll spare you another video. 😄

philrz:
The plot thickens! After that finding on Linux I happened to try the test with the most recent Zui Insiders and could not repro the problem. This led me to a binary search by which I confirmed that the changes in #2584 seem to have fixed the problem on Linux (and presumably Windows as well... I'll test on that next.) Perhaps this should be unsurprising since a lot changed in that PR and at least one of the commit notes speaks of "Rework app quit".

In conclusion, it seems possible that we only need fixes on macOS. 🤔

philrz (approved):
After the recent additional changes, I've re-tested this branch with a Dev build based on commit 2769f5d and everything checks out fine on macOS, Linux, and Windows. I'll put some official verification videos in the linked issue when it merges, but this PR is a functional 👍 for me.

PR #2649, opened 2023-02-03
E2E Fix

jameskerr:


philrz (approved):
I wanted to see for myself if this improved things, so I made a personal fork and rigged up the e2e tests to run in Actions similar to how they do in the "advance zed" workflow. Alas, I'm not sure if it helped as it still failed most of the time. But it still seems to run fine with these changes, so it probably doesn't hurt either. Perhaps one day I can set aside the time to hack at this and figure it out. 🙁

jameskerr:
Ugg. I don't want to merge this then. It adds more complexity and, if it doesn't help anything, it's not worth it.

PR #2651, opened 2023-02-03
fix ingest warnings

### LINKED ISSUE #2554: No error message displayed on json load exception (regression after GA Brim v0.28.0)

### LINKED ISSUE #2606: No error message displayed on CSV load exception

nwt:
Ingest warnings have been broken for a while.  Fix them.

Included are fixes to some global dispatcher bugs that were causing the Pools.appendWarning reducer to run three times per dispatch.

The video below shows what ingest warnings look like now.

https://user-images.githubusercontent.com/2574448/217313236-44ebf753-2e5c-4545-abba-5af486c24bf0.mov

Closes #2554
Closes #2606

### DIFF
--- a/src/js/components/IngestWarningsModal.tsx
+++ b/src/js/components/IngestWarningsModal.tsx
@@ -33,7 +33,7 @@ export default function IngestWarningsModal({onClose}) {
         <>
           <p>
             {
-              "The data you've attempt to import was not recognized as any supported packet capture or log format. The errors returned by each log parser:"
+              "The data you've attempted to import was not recognized as any supported packet capture or log format. The errors returned by each log parser:"
             }
           </p>
           <Scrollable>
--- a/src/js/components/IngestWarningsModal.tsx
+++ b/src/js/components/IngestWarningsModal.tsx
@@ -2,7 +2,7 @@ import React from "react"
 import {useDispatch, useSelector} from "react-redux"
 import ToolbarButton from "src/app/query-home/toolbar/actions/button"
 import Current from "../state/Current"
-import Loads from "../state/Loads"
+import Pools from "../state/Pools"
 import useEnterKey from "./hooks/useEnterKey"
 import {
   ButtonGroup,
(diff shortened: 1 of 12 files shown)


src/js/components/IngestWarningsModal.tsx
```diff
@@ -31,7 +31,11 @@ export default function IngestWarningsModal({onClose}) {
       <Title>Ingest Warnings</Title>
       {warnings.length ? (
         <>
-          <p>{"Data ingest failed with these warnings."}</p>
+          <p>
+            {
+              "The data you've attempt to import was not recognized as any supported packet capture or log format. The errors returned by each log parser:"
```
jameskerr, line 36:
```suggestion
              "The data you've attempted to import was not recognized as any supported packet capture or log format. The errors returned by each log parser:"
```

jameskerr (approved):
One spelling error, then good to merge.

philrz:
It's definitely a huge improvement that the warnings are now visible to the user, so I'd absolutely take this fix as is. That said, given that "load" and "ingest" are roughly synonymous, it feels strange that the messages "Ingest failed with warnings" and "Load successful" appear on the screen at the same time. If it were possible in this case to change the "Load successful" to something that reflects the load only partially succeeded, that would help a lot, I think.

nwt:
@philrz: I agree, but that "Load successful" toast comes from code I'm not touching here, and the way to make the change you suggest isn't obvious to me, so I think that's fodder for another PR.

philrz:
Thanks @nwt. I've opened #2660 so we can revisit that in the future.

PR #2653, opened 2023-02-04
Change registered URL prefix and redirect URL from "brim" to "zui"

philrz:
I've been testing Auth0 configs to verify the changes in:

* https://github.com/brimdata/zed/pull/4354
* https://github.com/brimdata/zed/pull/4355
* https://github.com/brimdata/zed/pull/4356

...with intent to draft a doc that walks through an example/working Auth0 config that a user could follow to add user auth to their own `zed serve`. I noticed that the registered URL prefix of the app is still `brim://` and this is a bit of config that we'll need to advise users to add on the Auth0 side, so I figured it's as good a time as any to update these to their Zui equivalents.

I've made a Dev build based on this branch and confirmed a successful auth flow with with it using my own scratch Auth0 trial account.

### DIFF
--- a/src/js/auth0/index.ts
+++ b/src/js/auth0/index.ts
@@ -7,7 +7,7 @@ interface Auth0Response {
 
 export default class Auth0Client {
   private audience = "https://lake.brimdata.io"
-  private redirectUri = "brim://auth/auth0/callback"
+  private redirectUri = "zui://auth/auth0/callback"
   // 'offline_access' ensures we receive a refresh_token
   private scope = "openid offline_access"
 
--- a/src/js/electron/initializers/custom-protocol.ts
+++ b/src/js/electron/initializers/custom-protocol.ts
@@ -2,7 +2,7 @@ import {app} from "electron"
 import {BrimMain} from "../brim"
 
 export function initialize(main: BrimMain) {
-  const brimCustomProtocol = "brim"
+  const brimCustomProtocol = "zui"
   app.setAsDefaultProtocolClient(brimCustomProtocol)
   app.on("second-instance", (e, argv) => {
     for (let arg of argv) {


nwt (commented):
I think we might need to update the "protocols" field in electron-builder.json.

philrz:
@nwt: Thanks for catching. I've pushed a commit that covers `electron-builder.json` as well and everything re-tested fine. I scratched my head for a bit because my tests had worked fine previously, and the online resources I could find about these settings via web search didn't really explain how their effects overlap or are complementary, so I became mildly obsessed with figuring out if they were wholly redundant or what. This brought me down a rabbit hole of the `lsregister` command on macOS that shows the prefixes registered by various applications.

The one observable difference I found is that when I had _only_ the changes in my first commit that _didn't_ touch `electron-builder.json`, the URL prefix shows up in that list only _after_ the first time I launch the app. That means if I just installed the `.dmg` and then immediately tried to do `open zui://foobar` it'd fail with the message `No application knows how to open URL zui://foobar...` Once I launched the app the first time, that registry would get populated and `open` would work fine from there on out. By comparison, now with the second commit that also makes the change in `electron-builder.json`, the entry shows up in `lsregister` as soon as I install the `.dmg` and so I can `open zui://foobar` immediately after install and the app launches.

If anyone reading this should want to fiddle with this stuff as well, I'll also share that it seems this registry is _not_ always cleaned of entries when apps are uninstalled. Part of what made it difficult for me to spot the difference from adding the `electron-builder.json` config was because after uninstalling old versions of Brim/Zui/Zui Insiders the registry still had an old entry for `zui` in it, but that wasn't obvious to me at first. Therefore when I started from this not-clean-but-I-thought-it-was-clean state and then installed a test build that had only the changes from the first commit, `open zui://foobar` would work right after install anyway. It was only when I started making builds with never-before-seen fake test prefixes that the difference became apparent.

I'll wait for @jameskerr to look this PR over too. Once it merges I'll update the Auth0 config behind our public cloud beta and make sure the team knows to be running a new Zui Insiders that'll have the updated prefixes if they want to connect to the cloud service.

jameskerr (approved):
Great Phil! Thank you for the research. I learned something new from that.

PR #2654, opened 2023-02-05
Use server-supplied audience in Auth0Client

nwt:
Auth0Client uses a fixed audience value (https://lake.brimdata.io) in the authorization request it sends to Auth0, limiting Auth0Client to use with Auth0 APIs created with that value as their API identifier.  Change Auth0Client to use the audience supplied by the lake service in the GET /auth/methods response instead.  (The lake service's audience is configured with the -auth.audience flag to "zed serve", added in brimdata/zed#4356).

### DIFF
--- a/src/js/auth0/index.ts
+++ b/src/js/auth0/index.ts
@@ -6,12 +6,15 @@ interface Auth0Response {
 }
 
 export default class Auth0Client {
-  private audience = "https://lake.brimdata.io"
   private redirectUri = "brim://auth/auth0/callback"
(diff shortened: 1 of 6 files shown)


philrz (approved):
I tested this using a Dev build based on this branch and and Audience setting of `https://mydemoapi.example.com` in my Auth0 config and `zed serve -auth.audience=https://mydemoapi.example.com ...` and was able to authenticate successfully from Zui to my remote lake and execute queries. 👍

jameskerr (approved):

PR #2661, opened 2023-02-07
Support for additional Zed float types

### LINKED ISSUE #2659: Support for additional Zed float types
Repro is with Zui commit 1d5c473.

Support for additional float types such as `float16` and `float32` were recently added in Zed. However, as shown in the video below, when this data is imported into the app the views don't render the values and instead the user is shown an "Unknown primitive" error.

https://user-images.githubusercontent.com/5934157/217318483-34429a17-2db7-4cc7-8ac8-9b0aaeacaa1c.mp4

jameskerr:
Fixes #2659


philrz (approved):
I tested with this branch and these float types were now rendered in the viewers as expected. A functional 👍 from me.

PR #2664, opened 2023-02-09
remove e2e from .github/workflows/advance-zed.yml

nwt:
This workflow fails routinely because "yarn e2e" is flaky in CI, so there's no point in continuing to run those tests.


nwt:
As discussed, we're going to keep this around.

PR #2665, opened 2023-02-09
use concurrency group in .github/workflows/advance-zed.yml

nwt:
Unlike the softprops/turnstyle action, which requires a runner, the concurrency group just queues the workflow.


philrz (approved):
Learned something new. 👍

PR #2669, opened 2023-02-13
Escape new lines in the inspector and table

### LINKED ISSUE #2668: Bad rendering of string value containing newlines
Repro is with Brim commit 1adeeafd.

First repro is using thee attached the standalone test data [greeting.zson.gz](https://github.com/brimdata/brim/files/10714665/greeting.zson.gz):

```
$ zq -Z greeting.zson.gz 
{
    greeting: "hello\nworld\this\nis\nsome\ntext"
}
```

After loading this in and looking in the Table view, the app shows just the "is".

image

When this is mixed in as part of a larger record in the attached [record.zson.gz](https://github.com/brimdata/brim/files/10714668/record.zson.gz), we see it does try to render it with the newlines, but the text steps on other text.

image

jameskerr:
Fixes #2668

### DIFF
--- a/src/app/features/inspector/views/string-view.ts
+++ b/src/app/features/inspector/views/string-view.ts
@@ -2,6 +2,6 @@ import {View} from "./view"
 
 export class StringView extends View {
   render() {
-    return `"${this.args.value.toString().replaceAll("\n", "\\n")}"`
+    return `"${this.value.toString().replaceAll("\n", "\\n")}"` // Escape all new lines
   }
 }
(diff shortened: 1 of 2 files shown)


philrz (approved):
@jameskerr: Is there a reason why one is wrapped in:

```
`"${}"`
```

...and the other is not? Seeing it as a teachable moment I tried removing the wrapper and it still seemed to work, but maybe there's more to it.

In any case, my tests show it providing the intended fix, so it's a functional 👍 for me.

jameskerr:
Yeah! The StringView class adds quotes to the beginning and end of strings. The BareStringView does not. The BareStringView is used in the table while the StringView is used in the inspector.

This bug appeared in the BareStringView Class where I override the StringView, but forgot to include the escaping new line feature.

Issue #2638, opened 2023-01-27
Add preference for specifying non-comma CSV delimiter


philrz:
At the time this issue is being filed, Brim is at commit 6f4f939.

https://github.com/brimdata/zed/issues/4238 tracked the ability to invoke a delimiter other than the default comma (`,`) in Zed's CSV reader. Since this alternate delimiter can now be requested over the Zed API, we should add a setting in **Preferences** so a user can have the benefit of this setting when they drag CSV data into the app.

PR #2670, opened 2023-02-14
Fix spelling of "delimiter"

philrz:
I was taking a shot at implementing something for #2638 but I noticed that there's already many references in the code to the word "delimeter" which is a misspelling of "delimiter". Therefore I figured I could try my hand at addressing the prior spelling as a prerequisite. Since one of the places it's misspelled is in the `appState.json` it seemed like a state migration was in order, so I've tried my hand at doing that for the first time.


jameskerr (approved):
Looks great Phil

PR #2679, opened 2023-02-16
Remove old state, add migration

jameskerr:
Here is a summary of state changes since the last release:

Places where State Changed

 /Appearance
- [x] Added poolsOpenState
- [x] Added queriesOpenState

 /Configs
- [x] This whole state slice can be deleted. Only the prop values need to be stored. This is state config.

 /DraftQueries
This was deleted

 /Editor
Brand new state, no migration needed

 /Handlers
This was deleted

 /Histogram
Not persisted

 /Inspector
Not Persisted

 /Lakes
Added audience to auth zero data, no migration needed

 /Layout
Need to add migration for these defaults
- [x]     currentPaneName: "history" as PaneName,
- [x]    isEditingTitle: false,
- [x]    titleFormAction: "create" as "create" | "update",
- [x]     showHistogram: true,

 /Loads
Not persisted

 /LogDetails
Not persisted

 /Modal
Not persisted

 /Pools
Not persisted

 /Queries
No migration needed, changed selectors and actions but not data

 /QueryVersions
Brand new slice, no migration needed

 /RemoteQueries
No migration needed

 /Results
Not persisted

/SessionHistories
Brand new slice, no migration needed

 /SessionQueries
Brand new slice, no migration needed

 /SystemTest
Removed

 /TabHistories
Changed descriptor object to strings, no migration needed

 /Table
Not persisted

 /Tabs
No migration needed

 /Viewer
Not persisted

- [x] Looks like I don’t need to persist window state that is also in the global state…


jameskerr:
I also removed state slices we don't rely on anymore. What a relief. These include Configs, SearchBar, Search, and Viewer

philrz (approved):
To test on the branch I started at the prior commit on `main`, populated a couple small pools, saved a query, quit the app, and saved off my `appState.json`. I then restarted in the branch at commit `7345619` so the migration could happen, did some simple smoke test operations that went fine, and saved off that `appState.json`. I turned them both into ZSON for easier viewing. I can see the new state file is indeed much smaller.

```
$ ls -l appState*.zson
-rw-r--r--  1 phil  staff  12318 Feb 17 09:47 appState-after.zson
-rw-r--r--  1 phil  staff  54167 Feb 17 09:46 appState-before.zson
```

I've also attached them for possible `diff`'ing purposes.
* [appState-before.zson.gz](https://github.com/brimdata/brim/files/10769558/appState-before.zson.gz)
* [appState-after.zson.gz](https://github.com/brimdata/brim/files/10769559/appState-after.zson.gz)

Alas, beyond that, I don't think I'd be able to tell from looking at the code changes if there's some gotcha lurking in there, so I've gotta kinda take a leap of faith and Approve. I figure once it merges I can repeat one of my "make a personal repo fork, do super realistic Brim/Zui upgrade tests" just to double check.

philrz:
Update: I did do another thorough round of upgrade tests in a personal fork at commit b80ce29 in prep for the Zui v1.0.0 release and saw no problems.

PR #2680, opened 2023-02-23
Don't migrate lake data in Brim/Zui transition

philrz:
The Zed lake storage format has recently gone through a change that's not backward compatible. Meanwhile, changes back in #2428 included a recursive copy of the lake storage as part of the Brim/Zui upgrade.

The attached video was taken with [this Dev Build](https://github.com/brimdata/brim/actions/runs/4247780001) that points at the newest Zed and shows the problem created: Upon first launch in Zui, `zed serve`'s inability to read the older storage format prevents the backend from starting up at all, so the user is confronted with a failure message.

https://user-images.githubusercontent.com/5934157/221018037-2de3ef7c-731d-4997-84f6-b08e7dfe0e79.mp4

If users want to migrate their data, the current plan of record is to point them at [migration tools](https://github.com/brimdata/zed-lake-migration) that they can run manually. Therefore what this PR does is revert the parts of #2428 that auto-migrate the lake storage. Instead the user's `zed serve` will come up ok with an empty lake. Like we did back in [Brim v0.29.0](https://github.com/brimdata/brim/releases/tag/v0.29.0), we can put up a prominent "Where are my pools?" Release Note that'll show up in the app and point them to the migration tools.

### DIFF
--- a/src/js/electron/migrateBrimToZui.ts
+++ b/src/js/electron/migrateBrimToZui.ts
@@ -7,19 +7,16 @@ export default () => {
   if (app.name !== "Zui") return
   const zuiPath = app.getPath("userData")
   const zuiAppStatePath = path.join(zuiPath, "appState.json")
-  const zuiAppLakePath = path.join(zuiPath, "lake")
   const zuiAppDataPath = path.join(zuiPath, "data")
 
   const brimPath = zuiPath.replace("Zui", "Brim")
   if (zuiPath === brimPath) return
 
   const brimAppStatePath = path.join(brimPath, "appState.json")
-  const brimAppLakePath = path.join(brimPath, "lake")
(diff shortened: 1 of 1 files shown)


philrz:
The video below shows a [Dev build](https://github.com/brimdata/brim/actions/runs/4256466800) based on this branch. Now when upgrading to Zui from Brim v0.31.0, the Zed service comes up ok so the user no longer sees an error. Since we did have a pool view open pre-upgrade they do get the "This pool has been deleted or never existed" message on that tab, but as long as the Release Notes with the "Where are my pools?" guidance show up on a tab we should be ok.

https://user-images.githubusercontent.com/5934157/221036141-152cfe0d-228b-4468-9ac2-84b008d70965.mp4

jameskerr (approved):

Issue #2685, opened 2023-02-28
Per Machine install?


Jhacker9:
Is there a way to install per-machine?  In our environment per-user installs are discouraged.  We need a silent per machine install in MECM/SCCM.  I note that in your ARP in registry the uninstall is /currentuser /S.  So wondering if there is something like /machine /S?  Alternatively, if I create an MSI with all the installed files and install that to say C:\Program Files (x86)\Brim with matching shortcut, would the product work?

philrz:
@Jhacker9: Interesting question! I must admit, we've not considered that one before. But based on the wording of your question, it sounds like your Windows admin skills exceed ours. Therefore I'll first share what I know and perhaps add your knowledge onto that. 😉

Per the [contributing](https://github.com/brimdata/brim/blob/main/CONTRIBUTING.md) doc, Brim is an [Electron](https://www.electronjs.org/)-based app, and it's packaged with [electron-builder](https://www.electron.build/). Per [these docs](https://www.electron.build/configuration/win), it looks like we're using the default package type of NSIS, so I guess [these NSIS docs](https://nsis.sourceforge.io/Main_Page) become relevant. Knowing I'm not going to be an instant expert in NSIS, I instead just read through the options [here](https://www.electron.build/configuration/nsis) and it confirmed that we've been relying on its default of a one-click, per-user install. And this all results in things landing in the paths described in https://github.com/brimdata/brim/wiki/Filesystem-Paths, which has the app binaries in one location and the user data in another.

[...]

PR #2686, opened 2023-03-01
Enable assisted installer so per-machine can be selected

philrz:
https://www.electron.build/configuration/nsis describes the options being used here. Up until now we've always gone with the [electron-builder](https://www.electron.build/) default of a one-click, per-user install. With the combination I'm using in this PR, the user will be prompted as shown in the following video (made with a Dev build from [this Actions run](https://github.com/brimdata/brim/actions/runs/4298560382)), such that they can choose the per-machine install. After selecting that install option I logged out and logged in as a different user and confirmed I saw the Zui icon on its desktop and could launch the app ok. The app binaries in that case end up in `%ProgramFiles%\Zui` (i.e., `C:\Program Files\Zui`) rather than the `%USERPROFILE%\AppData\Local\Programs\Zui` that we're accustomed to seeing with the per-user install.

https://user-images.githubusercontent.com/5934157/222031766-a5646a5d-5aec-4025-9cdb-9d0eb54c541c.mp4

Since Zui Insiders derives its electron-builder config from Zui, I also made a Dev build in [this Actions run](https://github.com/brimdata/zui-insiders/actions/runs/4299006773) and confirmed that also worked fine with per-machine install.

If this merges, I'll update the docs in #2531 to mention this as an alternate install option and filesystem path.

For #2685

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -12,6 +12,7 @@
   "linux": {"target": ["deb", "rpm"]},
   "rpm": {"depends": ["openssl"]},
   "deb": {"depends": ["openssl"]},
+  "nsis": {"oneClick": false, "perMachine": false},
   "forceCodeSigning": true,
   "afterSign": "electron-builder-notarize",
   "publish": {


philrz:
This morning I also used my personal fork to confirm that this plays nicely with auto-update, and the tests were successful. A summary:

* Installed a test `v1.0.1` machine-wide as an Admin user with a `v1.0.2` available on GitHub as the latest. Still logged in as the Admin user, I was eventually prompted to install the newer release. After I clicked **Restart** I was once again shown the assisted installer and once again picked the machine-wide install. Logging out & logging in as my non-Admin user, the Zui shortcut on the desktop launched the app and it went to the `v1.0.2` app as one would expect.

* Installed a test `v1.0.1` machine-wide as an Admin user with a `v1.0.2` available on GitHub as the latest. Once the `v1.0.1` install completed, I immediately logged out and logged in as my non-Admin user, clicked the desktop shortcut to launch the `v1.0.1` app, and was eventually prompted to install the newer release.  After I clicked **Restart** I was once again shown the assisted installer and once again picked the machine-wide install, but was immediately prompted to enter Admin credentials which I canceled out of since I was a non-Admin user. At that point it brought me back to the assisted installer and I was able to pick the single-user install, and that went through fine and that user could now run `v1.0.2`. This user now had duplicate Zui shortcuts on the desktop which is indeed potentially confusing, but I see that more as an IT issue. At that point I logged out and switched over to my Admin user whose desktop shortcut was confirmed to still launch `v1.0.1`. As that Admin user I uninstalled Zui and then switched back to my non-Admin user who now had the single desktop shortcut to their personal install of `v1.0.2`, which still ran fine.

jameskerr (approved):
Very cool Phil!

PR #2692, opened 2023-03-03
Add link to new Zui docs site in Help pull-down menu

philrz:
I just noticed we weren't yet linking to the new site in the Help menu, so adding that here. The URL is not yet rigged up with tags since we're still on our first Zui docs release, but I can enhance this when it's needed.

I also moved the "Show Welcome Page" entry down a hair so the most helpful docs would be up top.

image

### DIFF
--- a/src/app/core/links.ts
+++ b/src/app/core/links.ts
@@ -4,6 +4,7 @@ const currentZedTag = pkg.dependencies.zed.split("#")[1] || "main"
 const zedDocsTag = currentZedTag.startsWith("v") ? currentZedTag : "next"
 
 export default {
+  ZUI_DOCS_ROOT: `https://zui.brimdata.io/docs`,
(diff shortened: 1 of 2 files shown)


src/app/core/links.ts
```diff
@@ -4,6 +4,7 @@ const currentZedTag = pkg.dependencies.zed.split("#")[1] || "main"
 const zedDocsTag = currentZedTag.startsWith("v") ? currentZedTag : "next"
 
 export default {
+  ZUI_DOCS_ROOT: `https://zui.brimdata.io/docs`,
```
nwt, line 7: Maybe put this down by ZUI_DOWNLOAD?

nwt (approved):

src/app/core/links.ts
philrz: @nwt: Ok, moved.

PR #2703, opened 2023-03-06
More Brim/Zui renames

philrz:
Now that the repo has been renamed and the [old Brim wiki](https://github.com/brimdata/zui/wiki) has been replaced with a pointer to the new Zui docs site, here's more Brim-to-Zui renames throughout the repo.

@jameskerr: I know you already did a pass through the actual code with lots of renames. I found a few lingering references to "Brim" in variables names and such that I did _not_ change since I assume you left them where they were intentionally. But if you spot more and want to change them in this PR, please feel free to push to the branch. Thanks!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Create a report to help us fix a problem in Brim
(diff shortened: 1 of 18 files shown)


jameskerr (approved):
Right on. Yes I left all the brim references that I found in URLs for after we renamed the repo.

PR #2718, opened 2023-03-09
Officially add reselect to our dependencies

### LINKED ISSUE #2717: Error: Cannot find module 'reselect'
This issue was reported in [Zui insiders v1.0.1-2](https://github.com/brimdata/zui-insiders/releases/tag/v1.0.1-2) by a community user in a [Slack thread](https://brimdata.slack.com/archives/C03MW6XT7HC/p1678394433968049). It has been reproduced on both Windows and macOS. I've also confirmed that the problem doesn't seem to show up in [Zui Insiders v1.0.1-1](https://github.com/brimdata/zui-insiders/releases/tag/v1.0.1-1), nor does it show up when I run Zui in Dev mode at the current tip of main (243c348).

Just install and try to launch, and the error pops up:

image

jameskerr:
It's not a good practice to rely on a node module that you don't have listed as dependency. This should fix #2717


package.json
```diff
@@ -161,13 +161,14 @@
...
     "react-markdown": "^6.0.2",
-    "react-redux": "^8.0.1",
+    "react-redux": "^8.0.5",
     "react-router": "5.3.1",
     "react-spring": "^8.0.27",
     "react-tooltip": "^4.2.7",
     "react-transition-group": "^2.9.0",
     "react-window": "^1.8.7",
     "regenerator-runtime": "^0.13.6",
+    "reselect": "4.0.0",
```
nwt, line 171: Should we make this `^4.0.0`?

nwt (approved):

philrz:
FWIW, I made a build based on this branch in https://github.com/brimdata/zui/actions/runs/4379068778 and confirmed it installed without the issue. I'll wait for @jameskerr to weigh in on @nwt's `^` question before we merge and make a new Zui Insiders release.

package.json
jameskerr: I want this pinned explicitly so that we don't have two versions of reselect. Right now it's pinned to the exact version of reselect that @reduxjs/toolkit depends on.

PR #2720, opened 2023-03-10
Link to Troubleshooting doc when Zui shows a connectivity error

### LINKED ISSUE #1497: Hyperlink from "Connection Error" screen to Troubleshooting article
Recently we experienced the second incident of a community user bumping into the "Connection Error: The service at localhost:9867" symptom detailed at #1490. Unsurprisingly, for both users we know of that have hit this issue, they ultimately end up coming to public Slack for help. Now that this has happened a couple times, we've got an article on the Brim wiki that would hopefully help a user start troubleshooting the problem immediately rather than having to take the detour to Slack and wait for a response:

[https://github.com/brimsec/brim/wiki/Troubleshooting#brim-shows-connection-error-the-service-at-localhost9867-could-not-be-reached](https://github.com/brimsec/brim/wiki/Troubleshooting#brim-shows-connection-error-the-service-at-localhost9867-could-not-be-reached)

In a group discussion on March 2, 2021, there was consensus that we could add a hyperlink from the "Connection Error" screen directly to the troubleshooting article. Proposed text attached to the hyperlink (with the link attached to the text that's in bold):

> Please see the **Brim troubleshooting wiki** for guidance to debug/resolve this issue

philrz:
As described in #1497, we've had a few incidents of users coming to Slack with tricky problems losing connectivity to the local Zed service behind their app. In some cases we've been able to help them figure it out, and some problems have remained a mystery. As a result, we've got a [section in the Troubleshooting doc](https://zui.brimdata.io/docs/support/Troubleshooting#zui-shows-an-error-the-service-could-not-be-reached) that shares what we know and provides tips, but that still leaves the challenge of making the user aware that the doc exists. This PR attempts to plug that gap by pointing them toward the doc in two places in the app where such a connectivity error occurs.

https://user-images.githubusercontent.com/5934157/224385178-cac116f1-9190-4bd4-b82b-64f9588d8816.mp4

Re-reading that section of the Troubleshooting doc in current times, I recognize conditions have changed slightly since it was first written. When first written, the concept of the app connecting to remote lakes was something we knew was supported by the architecture but 100% of our users were using the app exclusively within their local desktops. Therefore the article was written as if the only reasons these errors could surface would be things like a local Zed service crash, a local firewall or anti-virus tool causing problems, etc. Now that "remote lakes" are something more easily configured in the app and described in [its own article](https://zui.brimdata.io/docs/advanced/Remote-Zed-Lakes) (albeit framed as an "advanced" guide) it's easy to see how these connectivity errors could come up in very ordinary circumstances, e.g., during a routine/temporary network outage. That made me slightly nervous that the text as it was written might sent a remote lake user down a rabbit hole of advanced debug when in fact a simple ping would suffice. So what I've included is a small adjustment to the top of that Troubleshooting section to nod toward the article about remote lakes before diving into the local debug details. It's probably not too big of a concern because a user that set up a remote Zed service already has a decent handle on the pieces, and a user connecting to such a remote service is likely to funnel their problems first through the person who's managing the remote Zed service. 🤞 

Closes #1497


nwt (commented):
Looks good to me but I'll leave approval to @jameskerr.

jameskerr (approved):

PR #2728, opened 2023-03-15
Zealot Streamline

jameskerr:
These changes make zealot work in browsers and node.

It also streamlines the tests for both environments and the publish process.


jameskerr:
This work has been moved to the zealot repo.

PR #2733, opened 2023-03-21
disable Playwright logging in .github/workflows/advance-zed.yml

nwt:
The pw:api logs from Playwright are voluminous, and I've never found them to be useful in CI.

### DIFF
--- a/.github/workflows/advance-zed.yml
+++ b/.github/workflows/advance-zed.yml
@@ -83,8 +83,6 @@ jobs:
         with:
           options: -screen 0 1280x1024x24
           run: yarn e2e
-        env:
-          DEBUG: pw:api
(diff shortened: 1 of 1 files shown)


jameskerr (approved):

philrz:
Thanks @nwt. FWIW, my take is that I've always found myself skipping over the debug output because the summary at the end of which test failed & why usually told me what I needed to know. That said, this PR also made me realize that I'd never looked super close at that debug output in the first place, so I started doing that for the first time yesterday while trying to debug #2734. Ultimately it wasn't too useful, mostly because I found it super difficult to pair up the debug output with the test being run. It does look like it could have the potential to be useful if that shortcoming could be addressed. In the meantime, I agree that it seems fine to get it out of the way by default.

Issue #2605, opened 2022-12-01
Allow dismissal of pop-up notifications


philrz:
Repro is with Brim commit 0a07e2a.

As shown in the attached video, rapid creation of lots of pools causes quick stacking of "Load Successful" pop-ups.

https://user-images.githubusercontent.com/5934157/205153618-717f9dc3-8528-4c2b-9563-aa0e8cd0737b.mp4

While human users may not typically work this fast, this is currently causing a challenge when trying to improve the e2e test in #2602, since it's not possible for the automation to know with certainty which "Load Successful" or "Load error" pop-up corresponds to which pool creation step.

One way to deal with this is to just wait for each new pop-up to clear on its own before proceeding to the next test step, but it takes several seconds for each one to disappear, so that slows down the test. An improvement I proposed would be to allow quicker dismissal of the pop-ups, such as via clicking them. This is something that even users may appreciate, so @jameskerr agreed this is something worth pursuing.

philrz:
Note to self: Once this change is in place, my hope is to next revisit the changes from #2602. What I think would be ideal is to add a last step to [`createPool()`](https://github.com/brimdata/brim/blob/1a2c460910ebcb310f385e70acf242ab8b6513aa/packages/e2e-tests/helpers/test-app.ts#L57-L70) to click-to-dismiss a load-related pop up right after the test acknowledges that it's visible.

PR #2735, opened 2023-03-21
Make the Export Completed check more strict

### LINKED ISSUE #2734: Intermittent export.spec.ts failures in CI
We see occasional failures of https://github.com/brimdata/zui/blob/main/packages/e2e-tests/tests/export.spec.ts in CI that look like this:

```
  1) export.spec.ts:38:5 › Export tests › Exporting in ZJSON format succeeds =======================

    Error: expect(received).toBe(expected) // Object.is equality

    Expected: 18007
    Received: 16384

      54 |       ).toBeVisible()
      55 |
    > 56 |       expect(fsExtra.statSync(file).size).toBe(expectedSize)
         |                                           ^
      57 |     })
      58 |   })
      59 | })

        at /home/runner/work/zui/zui/packages/e2e-tests/tests/export.spec.ts:56:43
```

I have an idea of what's wrong and will put up a fix.

philrz:
The intermittent failure in #2734 is a pattern that's been seen before, such as described in #2605. The stacking of the "Export Completed" pop-ups make it such that every export after the first is prone to seeing the pop-up from a prior successful export and thinking the one it just initiated has already finished. I believe this can explain why the check for the expected file size can be attempted on a file that's only been partially been written and hence fail the test.

image

Unlike in #2605, in this case the pop-up has enough specific text (the file extension) that we can uniquely wait on the pop-up that references each format.

Fixes #2734

### DIFF
--- a/packages/e2e-tests/tests/export.spec.ts
+++ b/packages/e2e-tests/tests/export.spec.ts
@@ -49,7 +49,7 @@ test.describe("Export tests", () => {
       await dialog.getByRole("radio", {name: `${label}`}).click()
       await dialog.getByRole("button").filter({hasText: "Export"}).click()
       await app.mainWin
-        .getByText(new RegExp("Export Completed: .*results." + label))
+        .getByText(new RegExp("Export Completed: .*results\\." + label))
         .waitFor()
 
       expect(fsExtra.statSync(file).size).toBe(expectedSize)
--- a/packages/e2e-tests/tests/export.spec.ts
+++ b/packages/e2e-tests/tests/export.spec.ts
@@ -48,10 +48,9 @@ test.describe("Export tests", () => {
       const dialog = app.mainWin.getByRole("dialog")
       await dialog.getByRole("radio", {name: `${label}`}).click()
       await dialog.getByRole("button").filter({hasText: "Export"}).click()
-
-      await expect(
-        await app.mainWin.locator("text=Export Complete").first()
-      ).toBeVisible()
+      await app.mainWin
+        .getByText(new RegExp("Export Completed: .*results." + label))
+        .waitFor()
 
       expect(fsExtra.statSync(file).size).toBe(expectedSize)
     })


packages/e2e-tests/tests/export.spec.ts
```diff
@@ -48,10 +48,9 @@ test.describe("Export tests", () => {
       const dialog = app.mainWin.getByRole("dialog")
       await dialog.getByRole("radio", {name: `${label}`}).click()
       await dialog.getByRole("button").filter({hasText: "Export"}).click()
-
-      await expect(
-        await app.mainWin.locator("text=Export Complete").first()
-      ).toBeVisible()
+      await app.mainWin
+        .getByText(new RegExp("Export Completed: .*results." + label))
```
philrz, line 52: I'll point out that the final `.` here should ideally be a `\.` since we're matching against the separator before the file extension. However, when I had `\.` in there, the linter grumped about it being an "Unnecessary escape character". Per [this post](https://stackoverflow.com/questions/47277133/disable-unnecessary-escape-character-no-useless-escape), I'm not the only person to have bumped into this. I had it working with the `//eslint-disable-line` to shut it up, but since the regexp is still specific enough to get the job done even with the `.` matching "any character" I left it like this. Happy to switch it over if folks feel otherwise.

nwt: I think the linter's complaint is legitimate. You want `Regexp()` to see a string that ends with `\.`, and because `\` is the string escape character, you need to escape the `\` itself.
```suggestion
        .getByText(new RegExp("Export Completed: .*results\\." + label))
```

philrz: Thanks @nwt. Fixed!

nwt (commented):
Looks good to me but I'll let @jameskerr approve.

jameskerr (approved):

Issue #2719, opened 2023-03-09
Dependency scrub


philrz:
As pointed out by @jameskerr in #2718:

> It's not a good practice to rely on a node module that you don't have listed as dependency.

Since we bumped into #2717 by accident, it got me to wondering if there was an automated tool that could help us sniff out other instances of this problem waiting to bite us. I did a quick search and the first thing I found was the popular [`depcheck`](https://github.com/depcheck/depcheck) tool. Running it at Zui commit 243c348 that's before the fix in #2718 gets merged, indeed we see that `reselect` was flagged in here, so it probably does the trick.

```
$ npx depcheck
Unused dependencies
* @babel/core
* @codemirror/commands
* @codemirror/comment
* @js-joda/core
* acorn
* ajv
* electron-dl
* event-source-polyfill
* prismjs
* react-input-autosize
* zed
Unused devDependencies
* @swc/jest
* @testing-library/dom
* @types/mocha
* @types/prismjs
* @types/sprintf-js
* abort-controller
* babel-core
* babel-eslint
* babel-plugin-module-resolver
* brimcap
* cpx
* electron-builder-notarize
* electron-notarize
* lint-staged
* livereload
* prettier-eslint
* rimraf
* sprintf-js
* whatwg-fetch
* win-7zip
Missing dependencies
* src: ./src/zui-kit/core/value-view/types.ts
* @tanstack/table-core: ./src/zui-kit/core/table-view/table-view-api.ts
* packages: ./src/plugins/log-loader/index.ts
* redux: ./src/js/state/types.ts
* redux-thunk: ./src/js/state/types.ts
* reselect: ./src/js/state/utils.ts    <------------------------------------------- Flagged!
* history: ./src/js/state/migrations/202103051447_convertHistoryToUrls.ts
* immer: ./src/js/state/Tabs/reducer.ts
* settings: ./src/css/main.scss
* shared: ./src/css/main.scss
* forms: ./src/css/main.scss
* global: ./src/css/main.scss
* buttons: ./src/css/main.scss
* icons: ./src/css/main.scss
* tags: ./src/css/main.scss
* text-content: ./src/css/main.scss
* prism: ./src/css/main.scss
* layout: ./src/css/main.scss
* modal: ./src/css/main.scss
* conn-versation: ./src/css/main.scss
* chart: ./src/css/main.scss
* filter-node: ./src/css/main.scss
* filter-tree: ./src/css/main.scss
* arrows: ./src/css/main.scss
[...]

PR #2737, opened 2023-03-24
Remove unneeded dev dependencies

nwt:
For #2719.


jameskerr (approved):
I added a commit that finished removing prism. Looks all good to me. The only one im uncertain of is prettier-eslint, but that's just a dev tool. If we need to add it back we can later.

PR #2743, opened 2023-03-30
Ensure a search window appears

### LINKED ISSUE #2730: On Linux, can't get back main window after quitting with just Details open
Repro is with Zui v1.0.0 on Ubuntu Linux. This issue was originally reported by a community user in a [public Slack message](https://brimdata.slack.com/archives/C010DR0HHMF/p1679101663032929).

As shown in the attached video, if the contents of the "Details" view are popped out into their own separate window, and then the app's main window gets closed somehow, when the app is ultimately quit and relaunched, it comes back with just the blank Detail window and no main window. I don't see a way to force it to re-open a main window when it's in this state.

https://user-images.githubusercontent.com/5934157/226077037-afbd62d3-0f4f-453e-b531-54a3d3790cc5.mp4

jameskerr:
Only make the search window persistable.

I found the bug. I was setting the "type" of the persistable property in the Detail Window class when I needed to set the "value" of it. The difference was a ":" needed to be an "=". They were both valid syntax so typescript didn't complain.

I've made it so that windows are not persistable by default. You have to opt in. 

The only window that opts in is the Search Window. 

And for people with old state, I make a check on startup that ensures at least on search window always opens when the app is first launched.

That should fix the issue.

Fixes #2730


src/js/electron/windows/about-window.ts
```diff
@@ -2,7 +2,6 @@ import {WindowName} from "../windows/types"
 import {ZuiWindow} from "./zui-window"
 
 export class AboutWindow extends ZuiWindow {
-  persistable: false
```
jameskerr, line 5: There's the bug! The ":" should have been a "="

src/js/electron/windows/window-manager.ts
```diff
@@ -23,7 +23,7 @@ export class WindowManager extends EventEmitter {
         .map(deserializeWindow)
       await Promise.all(windows.map((w) => this.register(w)))
     }
-    if (this.count === 0) await this.create("search")
+    if (this.byName("search").length === 0) await this.create("search")
```
jameskerr, line 26: Always ensure a search window gets created.

src/js/electron/windows/zui-window.ts
```diff
@@ -17,7 +17,7 @@ export abstract class ZuiWindow {
   lastFocused: number
   state: State | undefined
   dimens: Dimens | null = null
-  persistable = true
```
jameskerr, line 20: It now defaults to false.

philrz (approved):
@jameskerr: I tested this out on the branch and it's definitely hitting the goal of making sure there's always a search window when a user opens the app. However, when testing I noticed what's captured in the attached video, which I suspect may be an unrelated thing that was hiding behind this primary issue, but I figured I'd point it out. If I do the same sequence where I close the main search window first and then the Details window, the History for that search window is not present when I relaunch. I suspect this is a repeat of something we've discussed in the past where closing windows is assumed to express willingness to lose the tabs/history that were attached to that window, though I seem to remember we'd made some special exceptions for the "last window" just in case users saw no difference between "closing the last window" and "exiting the app". If I have all that correct, then maybe what I'm witnessing here is the Detail window being treated as "last window" in a way it maybe shouldn't? In any case, I'm happy to merge what's here since it's a significant improvement and I'd be happy to open a fresh issue if you think it's something worth looking into more.

https://user-images.githubusercontent.com/5934157/229176731-23b13c5d-15d1-43e3-9258-237f33062a51.mp4

PR #2758, opened 2023-05-02
upgrade Zed to f733ef3722f78d988cb94071e5ba8d2539a39da9

nwt:


jameskerr (approved):
Nice. The code got cleaned up. It looks like there is no more Sequential Proc.

PR #2759, opened 2023-05-02
Separate Renderer from Main Code

jameskerr:
To Do List:

- [x] Publish relevant Zealot Packages
- [x] Update this PR to point to updated zealot packages
- [x] Test about window
- [x] Test detail window
- [x] Write migration script for the packet indexes
- [x] Write migration script for suricata data
- [x] Spawn suricata updater
- [x] Test with a packaged app
- [x] Bring in changes from old zealot that got made since the new zealot repo was created

## Things to Test

1. Import some pcaps into the app before this change
2. Upgrade to this change, make sure you can still extract those pcaps
3. All the windows work
4. The packets button works
5. Ingesting pcaps works
6. Ingesting non-pcaps works

## Summary

An electron app is composed of a main node process, and one or more chrome renderer processes. Electrons makes it possible to use node APIs within the renderer processes. This is done by providing an option to the BrowserWindow constructor called nodeIntegration: true. 5 years ago, this was the default. It was what made electron unique. However, this created big security risks. Since a web browser is built to load remote files. When remote files are loaded, they also have full access to the node apis, and thereby the computer they are running on. (file system, processes, etc...).

It became good practice to turn off nodeIntegration and use interprocess communication to access the powerful node apis. That is now the default in latest electron versions.

Historically, our app has been like a node application that runs in the browser. With this PR, I have turned off nodeIntegration and migrated to ipc for everything that needs node APIs. 

## Plugins

The biggest place we used the node apis was in the plugins for brimcap, zeek, and suricata. These have been re-arranged to run in the node process. The API they use to interact with the app has been much improved as well in this PR. 

## Zealot

This PR now uses the built versions of @brimdata/zed-js and @brimdata/zed-node published on NPM and maintained in the brimdata/zealot repo.

## Packaged Files

I changed the electron-builder configuration to only include files from `dist/`, `out/`, and `zdeps/`. It was previously including everything. I also notice that we bundle two versions of the `zed` and `zq`. We also include some other rather large binaries from the zed repo we don't need. We could reduce the size of the packaged app by 200Mb if we fix this and only include what we need.

## Installers

Installers now get put in `installers/`. They were previously in `dist/installers/`. But since dist was and is included in the package other versions of the installers could be included in the app package. That's why some builds were many gigabytes large. 🤦

## New Tools & Commands

**Renderer**

We now use nextjs to build the renderer javascript bundle.

```
yarn start:renderer
```
This starts a dev server on port 3000 and serves the html the browser window needs. It will perform hot module reloading whenever a file changes. The main process, when in dev mode, will load this url. Changes to files will immediately be reflected in the browser window.

```
yarn build:renderer
```
> outputs to `out/`

This will create an folder (`/out`) of optimized, static files to serve in production mode. This is located in the `out/` folder. When the app is in production mode, it will load these files directly from the file system.

**Main**

We use esbuild to build the main process bundle. 

Esbuild bundles all the typescript files used in the main process into one javascript file. This is not required in node, but is convenient as it allows us to keep all our source typescript files (main and renderer) in the same directory. Esbuild works by only including that are imported, not everything in a directory.

```
yarn start:main
```
This runs the below command and triggers a rebuild whenever files change.

```
yarn build:main
```
> outputs to `dist/main.js`

```
yarn start:electron
```

This wraps the electron start script in a tool called nodemon. This will watch a directory and restart the node script on changes. It watches the /dist folder where the main process bundle lives. This is perfect for development.


philrz:
I'm testing on macOS with the build from https://github.com/brimdata/zui/actions/runs/4920574674 and the **About** window comes up blank for me.

philrz:
Repro is with the build from https://github.com/brimdata/zui/actions/runs/4920574674.

As shown in the attached video, when I try to load the [Zeek format zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default) into Zui, it fails with the error:

> Error invoking remote method 'loadFilesOp': invalid operation: no records in request

https://user-images.githubusercontent.com/5934157/236970589-59cf975e-445f-4b87-8456-4e3a1945b46c.mp4

philrz:
Now that GA Zui v1.1.0 is out and this branch has been caught up with `main`, I'm getting back to testing it again, this time with the new Dev build at https://github.com/brimdata/zui/actions/runs/5072267182.

First, I've verified that the three issues I reported in comments above have all been fixed. 👍 

Here's a new issue I found, though. While the **About** window comes up ok now, I can click on the links shown for **Website** and **Source** and they come up in my browser as expected. However, clicking on either **License** or **Acknowledgments** is a no-op. Doing the same in a GA Zui on my Mac, the relevant files are opened up for view in a text editor.

image

philrz:
I've made a note-to-self for when preparing the next Zui release I'll need to update these docs to show the new paths in this PR that take the place of the former `brimcap-root` path:

* https://zui.brimdata.io/docs/support/Filesystem-Paths
* https://zui.brimdata.io/docs/advanced/Remote-Zed-Lakes
* https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config

philrz (approved):
Using a new Dev build in https://github.com/brimdata/zui/actions/runs/5083087906 I've verified the fixes to the last couple issues reported in comments above. I feel like I've done enough testing on this one for it to merge and get some extended soak time in Zui Insiders before the next GA release, so, 👍.

PR #2761, opened 2023-05-04
Improve .github/workflows/advance-zed.yml

nwt:
* Add on.workflow_dispatch for testing.
* Use the setup-zui action.
* Set user.name and user.email in the "git commit" step.
* Change user.name to the value used in brimdata/brimcap.


philrz (approved):
LGTM. 👍  Given https://github.com/brimdata/zed/pull/4690 I expect you intend to next update `setup-zui` to use the Go version in `go.mod` rather than the explicit ref to 1.19, eh?

nwt:
> Given [brimdata/zed#4690](https://github.com/brimdata/zed/pull/4690) I expect you intend to next update `setup-zui` to use the Go version in `go.mod` rather than the explicit ref to 1.19, eh?

I do.

PR #2764, opened 2023-05-05
Have Shift+Enter run queries when "Run Query on Enter" is unchecked

### LINKED ISSUE #2763: Have "Shift-Enter" run queries when "Run Query on Enter" is unchecked
A community zync user made this suggestion, at which point @jameskerr observed:

> I keep doing that personally wishing it would run.

So it sounds like we should go ahead and do it. 😉

philrz:
Fixes #2763

I tested this out on macOS, Linux, and Windows using the dev artifacts from https://github.com/brimdata/zui/actions/runs/4889204739.


jameskerr (approved):
Dig it!!

PR #2790, opened 2023-06-28
Ensure Initializers Contain a Default Export

### LINKED ISSUE #2778: Auto-update not working
Repro is with Zui Insiders [1.1.1-2](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-2) and newer.

I recently happened to open up my Zui Insiders for the first time in a while. Based on how it's updated daily, I'm accustomed to usually expecting an auto-update notification within the first couple minutes of it being open, but to my surprise it never arrived. I happened to be running release [1.1.1-2](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-2). I did some further testing and found that auto-update was working in older release [1.1.1-1](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-1) (in which case it auto-updated to the latest available as normally expected, which is currently [1.1.1-12](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-12)) but [1.1.1-2](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-2) and newer releases such as [1.1.1-11](https://github.com/brimdata/zui-insiders/releases/tag/v1.1.1-11) just silently show no mention of auto-update attempts/failure in the DevTools console nor in the `main.log`.

Further digging shows that the Actions run https://github.com/brimdata/zui-insiders/actions/runs/5092300194 was based on Zui commit b7dbd16 which was associated with #2759, a recent big refactor that did indeed cause several other regressions in the app that we found & fixed. Therefore it seems likely this is yet another casualty and hence one we'll need to fix before we tag our next GA Zui release.

jameskerr:
The auto update initializer was not running because it did not alias itself when exporting its symbols in the src/electron/initializers/index.ts file. When initializers are run, they check if the name has a function called "initialize". Since the auto update export had no name, its name was "initialize". Then when the code checked if "initialize" had an "initialize" function, it was false.

The changes here are:
1. Now if that check returns false, we will throw an error.
2. The auto-update.ts initializer is now exported with an alias so that it works when called by src/electron/run-main/run-initializers.ts.
3. I added documentation of all this in CODE.md for my future self.

Fixes #2778


jameskerr:
I'm not sure what happened to this branch. I tried to force push, which would clean up the commits, but GitHub might be trying to protect me by not rewriting the history. I'm not sure, so I'm opening a new PR #2793

PR #2797, opened 2023-07-07
Zui with No Tabs Open

jameskerr:
In order to achieve this, we had to make these changes.

1. Remove the lake from the URL

The "url" for each tab was previously prefixed with `/lake/:id`. This had to be removed, and a migration created to migrate all the urls saved in the state. The routes have been updated to reflect this as well.

2. Store the current lake id in the state for each window.

A window will only be able to connect to one lake at a time. This is like how slack and vscode have designed their apps. Their "workspaces" are the equivalent of our "lakes". You can have multiple windows connected to different lakes though. Wrote migration to set the lake id to whatever the lake id in the last opened tab.

3. Remove the constraint of having 1 tab always open.

4. Create a view for when there are no tabs open. Right now it's just the big zed logo background image.

5. Assign the tabs to a lake. When switching lakes, the tabs for the previous lake will be saved and hidden. When switching back, they will be restored.

Extras:
* I made the width and presence of the detail pane the same for every tab in a window. This keeps it from jumping around as you move through the tabs. Feels much better.
* I styled the tabs slightly different.

Questions Found To Follow Up?

Do we need the errors reducer anymore?
Do we need the URL reducer??
Do we need the toolbars reducer anymore??


philrz:
I'm testing this PR at commit d8bcfb1.

In the attached video I'm running Zui Insiders to create a second lake I can easily connect to from my main test Zui. The video shows that after I **Logout** of the "remote" Insiders lake the app says some correct things like "The lake previously on this tab has been removed" and a tab says "Choose a lake", but the controls I'm accustomed to seeing that would normally allow me to fall back to just the default lake behind Zui or connect to a different remote lake don't seem to be present.

https://github.com/brimdata/zui/assets/5934157/3de25ca8-3584-4751-8e52-3c91f81d49b8

philrz:
I'm prepping to do some testing of what happens when upgrading from the last GA Zui release [v1.1.0](https://github.com/brimdata/zui/releases/tag/v1.1.0) to Dev builds based on this branch, as I want to fully understand what happens to details like editor contents, History, etc. after upgrade. As the old design made it somewhat difficult to even get to a state with connections to different lakes in different tabs, I've settled on a a consistent recipe that gets me into that as my "before upgrade" state. Therefore I'm describing & showing that in this comment and will refer back to it in other comments where I cite problems I might find after upgrading to a Dev build from this "before" state.

The steps described in text:

1. Have GA Zui [v1.1.0](https://github.com/brimdata/zui/releases/tag/v1.1.0) and the latest Zui Insiders installed
1. Start with Zui and Zui Insiders not launched and all lake/state cleared out
1. Open Zui Insiders and import [results.zng.gz](https://github.com/brimdata/zui/files/12045163/results.zng.gz), then minimize Zui Insiders (we only need it for its "remote" lake)
1. Open GA Zui and import [sample.zng.gz](https://github.com/brimdata/zui/files/12045164/sample.zng.gz)
1. Add the connection to the Zui Insiders lake on http://localhost:9988
1. Click on the `results.zng.gz` remote pool and then click **Query Pool** to start a Query Session
1. Click back to the **Welcome to Zui** tab
1. Change the lake connection to the local one behind GA Zui
1. Click on "**+**" to open a new Query Session
1. Select the `sample.zng.gz` pool from the drop-down
1. Close the other tabs leaving just the query sessions for `results.zng.gz` and `sample.zng.gz` (note how the selection in the lake pull-down changes as I click between tabs)

https://github.com/brimdata/zui/assets/5934157/bb7c5fd9-ac2f-4057-8ca3-5153cd6e3098

philrz:
I'm starting from the "before" state described in https://github.com/brimdata/zui/pull/2797#issuecomment-1635068520 and am testing with a [Dev build](https://github.com/brimdata/zui/actions/runs/5548691294) based on this branch at commit 4aae2fb.

The attached video picks up after I've closed the GA Zui v1.1.0, installed the Dev build, and am launching it for the first time. The first odd things I see involve the right pane. Specifically:

1. It starts out collapsed even though it was expanded when I closed the GA Zui v1.1.0 before upgrading
2. Clicking the headers in the right pane like **History** and **Detail** cause its width to change

https://github.com/brimdata/zui/assets/5934157/c57eb170-71d4-4888-9aa4-75483fc52dde

By comparison, if I launch the Dev build "fresh" (i.e., not an upgrade) I don't see these problems, so I suspect something wrong with the migration.

https://github.com/brimdata/zui/assets/5934157/f3dc0dae-9532-4d73-90e6-1f9beb9b2484

philrz:
I'm once again starting from the "before" state described in https://github.com/brimdata/zui/pull/2797#issuecomment-1635068520, though this time before I close GA Zui v1.1.0 I have a `count() by _path` query  in my editor/History for the local pool and a `count() by commit` query in my editor/History of the remote pool. I happen to have the _remote_ lake connection selected when I close GA Zui v1.1.0. After closing the app I upgrade to a [Dev build](https://github.com/brimdata/zui/actions/runs/5548953611) based on this branch at commit 07af8fb.

As shown in the attached video, when the updated app launches, both tabs are effectively bound to the remote lake, i.e., the one that had my `count() by _path` of the local pool now correctly cites that the pool doesn't exist on the remote lake and the pool selector only offers pools on the remote lake. This is all an accurate reflection of the post-upgrade state, but as a user, if my editor/History contents were actually something I was hoping to continue using, I'd now have to manually recreate them in a blank tab for the local lake connection, since that's effectively now a blank slate the first time I select it from the lake drop-down.

Given that the new design allows for different sets of tabs to be bound to different lake connections, is it feasible for the migration to automatically bind the tabs to their respective lake connections so the editor/History contents can continue on as they had been? If the reason we're not already doing this is because it's too difficult, perhaps that's an ok sacrifice to make as a price for getting into the new design. But I figured I'd check first in case what I'm showing here is unexpected.

https://github.com/brimdata/zui/assets/5934157/3c3c47ca-2d8c-4ab7-8443-dac4ffb3d323

philrz:
I'm testing this branch using a [Dev build](https://github.com/brimdata/zui/actions/runs/5577804306) at commit 6f113cd.

This may be something already planned to be addressed in a follow-on PR to deal with this point mentioned in the PR intro text:

> Create a view for when there are no tabs open. Right now it's just the big zed logo background image.

...but here's a couple things in this attached video that somewhat concern me. In this case I'm starting out with a clean launch of the app with no prior data/state.

1. When I drag in my first piece of data, the main window remains blank. At the tip of `main` the area shown blank here changes to the pool summary page with the **Query Pool** button. The way things are right now in this branch I'm not sure a new user would know to click the pool name in the left nav or click the "+" to start a Query Session tab.

4. When I drag in my _second_ piece of data, I _am_ now automatically shown the pool summary page with the **Query Pool** button. However, it seems like that pool summary tab has effectively "written over" my prior Query Session tab, as the `count() by _path` I had active in there & in its History now seems to be gone.

https://github.com/brimdata/zui/assets/5934157/0761a077-0c6a-46b0-8fad-e2d05255ae2c

philrz (approved):

PR #2519, opened 2022-09-06
Comments at the end of lines will work now

jameskerr:
Fixes #2452 

Comments at the end of lines don't break queries anymore. Thanks for the rec @nwt 

image

Issue #2452, opened 2022-07-20
Comments in Zed code not handled cleanly


philrz:
Repro is with Brim commit 55d1fdc0.

As shown in the following video (test data is [sample.zng](https://github.com/brimdata/brim/blob/main/src/test/shared/data/sample.zng)), adding a comment to a working Zed script caused it to stop returning results.

https://user-images.githubusercontent.com/5934157/180063284-688449e7-942c-403e-bc5d-f1bdfa215983.mp4

On closer inspection, i think I see the problem. Sniffing in Wireshark shows how the app tacks on more pipeline elements to help with pagination of results.

```
{"query":"from 'sample.zng' | _path==\"dns\" // foo | { i: count(), v: this} \n      | i > 0 \n      | head 500\n      | yield v"}{"query":"from 'sample.zng' | _path==\"dns\" // foo | { i: count(), v: this} \n      | i > 0 \n      | head 500\n      | yield v"}{"query":"from 'sample.zng' | _path==\"dns\" // foo | { i: count(), v: this} \n      | i > 0 \n      | head 500\n      | yield v"}
```

Indeed, if we hit the API directly, we see it fail similarly.

```
$ cat app-query.json 
{"query":"from 'sample.zng' | _path==\"dns\" // foo | { i: count(), v: this} \n      | i > 0 \n      | head 500\n      | yield v"}

$ curl http://localhost:9867/version
{"version":"v1.2.0-8-gd10050fe"}

$ curl -X POST -H 'Accept: application/x-zson' -H 'Content-Type: application/json' http://localhost:9867/query -d @app-query.json
[no response]
```

However, if we remove the `\n` newlines from the query, it starts working.

```
$ cat no-newlines.json 
{"query":"from 'sample.zng' | _path==\"dns\" // foo | { i: count(), v: this} | i > 0 | head 500 | yield v"}

$ curl -X POST -H 'Accept: application/x-zson' -H 'Content-Type: application/json' http://localhost:9867/query -d @no-newlines.json
[...]

PR #2822, opened 2023-07-26
Add newlines to histogram queries

### LINKED ISSUE #2819: Empty histogram when program closes with a comment
Repro is with Zui commit 1feb29c.

As shown in the attached video, when I end my Zed program with a comment, the histogram is drawn empty and instead the message "Field 'ts' did not return time values" is displayed.

https://github.com/brimdata/zui/assets/5934157/09be1087-6549-4907-ada5-2dd24e608176

I recall a fix we made for a similar issue in the past, so I may try my hand at fixing this one myself.

philrz:
The issue in #2819 looks to be a revisit of what we first saw in #2452. Like that issue, it seems the fix is as simple as adding a newline after the user's base query so that way the additional Zed that app tacks on will still be active even if the user's program ended with a comment.

Regarding coding style, I wasn't sure if indents were necessary to make it obvious that these string literals spanned multiple lines. However, the prior fix in #2519 didn't have indents and what I've got in this branch seems to pass the linter.

I've put down @nwt as an additional reviewer just to double check that there's no negative side effects we can think of from these newlines being added.

Here's the video evidence on this branch of the fix working for the three variations laid out in #2819

https://github.com/brimdata/zui/assets/5934157/0dced062-47d7-4f34-a244-1e43f716d085

Note that I've not tried to dig into the "TypeError: Cannot add property 2000, object is not extensible" error I showed in #2819 since that's a bit beyond my expertise. As I expected, my fixes here seem to prevent that error from being triggered anymore, but maybe there's still something to look closer at there.

Fixes #2819

### DIFF
--- a/src/panes/histogram-pane/run-query.ts
+++ b/src/panes/histogram-pane/run-query.ts
@@ -59,8 +59,8 @@ export async function runHistogramQuery(api: ZuiApi) {
   }
 
   async function getNullTimeCount() {
-    const query = `${baseQuery}
-    | ${timeField} == null | count()`
+    // Newline after baseQuery in case it ends with a comment.
+    const query = `${baseQuery}\n | ${timeField} == null | count()`
     const id = "null-time-count"
     const resp = await api.query(query, {id, tabId})
     const [count] = await resp.js()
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Interesting!

src/panes/histogram-pane/run-query.ts
```diff
@@ -59,15 +59,17 @@ export async function runHistogramQuery(api: ZuiApi) {
   }
 
   async function getNullTimeCount() {
-    const query = `${baseQuery} | ${timeField} == null | count()`
+    const query = `${baseQuery}
+    | ${timeField} == null | count()`
```
nwt, line 63: I'd add a comment so no one unwittingly undoes this. And I think an escaped newline improves readability.
```suggestion
    // Newline after baseQuery in case it ends with a comment.
    const query = `${baseQuery}\n | ${timeField} == null | count()`
```

nwt (approved):

PR #2825, opened 2023-07-28
Fix links and paths

philrz:
Recent changes to the organization of the Zed docs and the move of some filesystem paths in #2759 made a couple things go stale in the Zui docs as of when I put out the new releases yesterday. I'm fixing all that here. I've seen a build of this run clean in a broken link checker on a personal staging deploy based on this branch. I also ran through the updated Remote Lakes article on a scratch AWS VM to make sure it's all correct with the new GA releases.


jameskerr (approved):
Love it. Thank you.

PR #2828, opened 2023-07-29
Don't require Community ID on all pcap-eligible conn records

### LINKED ISSUE #2827: Community ID is a prerequisite to activating the Packets button
Repro is with Zui commit e1da723.

I recently went to update the [Custom Brimcap Config](https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config) article and found something had stopped working since I last ran through the steps. In that doc, logs for a pcap are loaded using Zeek & Suricata that are _not_ the ones bundled with Brimcap, since the doc wants to show users the a la carte steps for working with customized analyzers. What I found was that the Packets button (the shark fin icon) was not becoming active when I clicked on normally eligible Zeek records that I'd loaded.

@jameskerr helped me find the root cause. The problem relates to this chunk of code, which gets called when the user clicks on a record.

https://github.com/brimdata/zui/blob/e1da72361e90dffdf56ef695e3a915c42f273bb0/src/plugins/brimcap/zeek/queries.ts#L21-L35

The `community_id` field is present in all the logs generated by the Zeek that we bundle with Brimcap since it includes a non-default Community ID package. However, an out-of-the-box Zeek does _not_ have Community ID enabled by default, so that `is(community_id, <string>)` check fails 100% of the time on such Zeek logs and the fin never lights up.

I show the effect in the attached video where I have two `conn` logs generated from the same pcap using out-of-the-box Zeek v6.0.0. The first log is with Zeek in its default config such that the `community_id` field is not present, and in the second I've uncommented the necessary line in my `local.zeek` config so that `community_id` is present. As it shows, in the latter case the Packets fin does indeed become active. It shows "no packets found" when clicked because I didn't happen to have the Brimcap packet index populated, but that's a separate topic.

https://github.com/brimdata/zui/assets/5934157/fe9002bb-aca8-42ec-8e4d-9d61bfb754b0

philrz:
Fixes #2827, which has all the background.

The attached video shows the fix having the desired effect: Now when I have a `conn` log that lacks a `community_id` field that meets all the other requirements, the Packets button still lights up.

https://github.com/brimdata/zui/assets/5934157/163d9dfd-98b3-4a21-b323-dfc24253f62b

### DIFF
--- a/src/plugins/brimcap/zeek/queries.ts
+++ b/src/plugins/brimcap/zeek/queries.ts
@@ -26,7 +26,6 @@ export function findConnLog(pool: string, uid: string) {
     uidFilter(uid) +
     `)
   | is(ts, <time>) 
-  | is(community_id, <string>) 
   | is(duration, <duration>) 
   | is(uid, <string>)
   | head 1


jameskerr (approved):

philrz:
The file moves in #2818 have made for a merge conflict too complex for me to untangle, so since this is a one-line change I'm going to just close this PR and start over. 😄

PR #2831, opened 2023-08-01
Use github token in ci build

jameskerr:


jameskerr:
I snuck this in the Monaco PR.

PR #2833, opened 2023-08-04
Fix the Insiders Release Flow

jameskerr:
The steps we now take are:

1. Grab the latest release tag from the zui-insiders repo.
2. Download the build_sha.txt file from the latest release assets
3. Compare that with the GITHUB_SHA that triggered the workflow
4. If they are the same, nothing has been updated, fail the workflow
5. If they are different, continue to publish the next version of zui-insiders
6. Upload a build_sha.txt to the latest release containing the GITHUB_SHA

Here is an example of when the build_sha and the GitHub.sha are the same:

CleanShot 2023-08-04 at 10 13 47@2x


.github/workflows/release-insiders.yml
```diff
@@ -0,0 +1,100 @@
...
+          repo: zui-insiders
+
+      - name: Upload build_sha file to the latest release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          file: build_sha.txt
+          tag: ${{ steps.just_released.outputs.release }}
+          repo_name: brimdata/zui-insiders
+          repo_token: ${{ secrets.PAT_TOKEN }}
+          overwrite: true
```
nwt, line 100: I think you might be able to get rid of this job by using [electron-builder's afterAllArtifactBuild hook](https://www.electron.build/configuration/configuration#afterallartifactbuild) to add build_sha.txt to the release.

nwt (approved):

.github/workflows/release-insiders.yml
jameskerr: Since electron-builder is being deprecated in favor of electron's own build process, I don't want to get more tied to it. I'll just merge as is, but I'm glad I learned about the afterAllArtifactBuildHook.

PR #2834, opened 2023-08-04
Preview & Load Feature

jameskerr:
Preview and Load Steps

Dropzone

- [x] Save that data to the form
- [x] Display a cool drop zone
- [x] Drag a file onto the app

Form

- [x] Create the container for the form on the right sidebar
- [x] Data Input
- [x] Pool Input
    - [x] Dropdown options
    - [x] Mother when existing pool
    - [x] Pool Name field
    - [x] Pool Key Auto Complete
    - [x] Sort order dropdown
- [x] Commit message Text Area
- [x] Author field
- [x] Cancel Button
- [x] Load Button

Shaper Toolbar

- [ ] ~Choose a saved shaper dropdown~ (Postponed)
- [x] Run button

Results Pane

- [x] Container
- [x] Split the container in two

Original results

- [x] Display original data result in a results pane

Shaped Results

- [x] Run the shaper and output the results in that pane

Results Instance

- [x] View Chooser
- [ ] ~Number of unique fields~ (Postponed)
- [x] Number of types
- [x] Number of rows

Submit Form to Existing Pool

- [x] Create a pool first with all the settings
- [x] Or select the existing pool
- [x] Stream the data through zq into the zed load
    - [x] Add the shaper script
    - [x] Add the commit
    - [x] Add the author

What about PCAPs?

The loader should first run brim cap and output results to a file, that file could then be “preview and loaded”.

Phil's findings during testing (see links to comments below)

- [x] Status bar shows "0 Types" and "Rows" without a number when loading large files ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1736299877))
- [x] `zq` processes are left running after cancelling out of Preview & Load of a large file ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1736299877))
- [x] Prior shaper is cached for loads to different pools ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1736309961))
- [x] Vertical carets in column headers of Table view don't do anything ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1736318783))
- [x] Errors when mixing different files types in the same Preview & Load (or loading any Parquet) ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1736334793))
- [x] No indication is given that the user is being shown only the first 100 rows ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1738242346))
- [x] "ZQ Errors" may be appear, but it may not be clear to the user what `zq` is or why it's relevant here ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1738250837))
- [x] **Message** portion of the configurable **Commit Settings** don't seem to be making it through intact ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1738266364))
- [x] A "quick load" option to bypass the Preview step might be useful ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1738274259))
- [x] Settings/Preferences doesn't appear if clicked when Preview & Load is already open ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1743605274))
- [x] Maintain default behavior & preserve existing setting for "Run Query On" ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1749551398))
- [x] Maintain default setting for **Sort Order** ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1749587666))
- [x] Table view in Preview & Load seems broken ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1749599772))
- [x] **Load** can be clicked multiple times, initiating redundant load operations ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1753282400))\
- [x] Add Custom Views to the preview and load page
- [x] Don't delete pool if load fails
- [x] Display load errors in the UI either a modal or collapsed section.  
- [x] If you have a deleted pool, then you drag data into the app, the pool selection is empty
- [x] `zq` process left running after clicking "x" to abort a Load in progress ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1761933959))
- [ ] ~Inconsistent Expand/Collapse buttons in respective Table views in Preview & Load vs. main query results screen ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1762130367))~ Postponed
- [x] A la carte pool creation pop-up still defaults to "Ascending" ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1769244578))
- [x] Clicking **Load** for second new pool doesn't bring up that pool's summary page ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1769285982))
- [x] Hitting **Enter** with expanded **Pool Settings** causes load to fail ([link](https://github.com/brimdata/zui/pull/2834#issuecomment-1773277296))


philrz:
I'm currently testing with this branch at commit cb3c015.

The following video shows two findings related to attempted preview & load of very large files.

1. Once I've dragged the large file icon into the app and the Preview & Load screen has appeared and the `zq` processes have started running in the background to gather preview details about the data, the status bar at the bottom shows "0 Types" and "Rows" without any number. While the info is still being gathered I think it would be better to have some kind of spinner here instead.

2. When I click the **Cancel** button to back out of the Preview & Load operation, we can see the `zq` processes are still running. With large data sets these could potentially continue running for a long time and consume lots of resources, so we should make sure these get killed.

https://github.com/brimdata/zui/assets/5934157/6b8cdb28-a2df-4111-b153-c786a7b84fbc

philrz:
I'm currently testing with this branch at commit cb3c015.

In the following video, I go through the Preview & Load workflow once to apply a shaper while loading data to a new pool. When I then drag a separate file with intent to Preview & Load and create a separate new pool, the shaper I'd used for my previous pool is already filled in. We've talked about maybe doing some creative things with re-usable shapers (e.g., offering a "recent shapers" picker), but in terms of how it's working right now, I'd suggest the we _not_ always persist the prior shaper. Different pools are often intended to hold wholly different kinds of data, so it doesn't seem safe to assume the same shaper will be relevant to the next newly-created pool.

That said, when the user is doing incremental data loads to an existing pool, if we wanted to persist the last-used shaper that was used for a successful load to _that specific pool_, I think that would make sense.

https://github.com/brimdata/zui/assets/5934157/b12297c7-514d-43da-ae56-aa9c54b32dc8

philrz:
I'm currently testing with this branch at commit cb3c015.

As shown in the attached video, when I click to the Table view in Preview & Load, there's vertical carets next to each column name that would normally be used to sort their values ascending/descending, but they don't do anything here and should therefore be hidden/removed. No errors appear in DevTools as I click.

https://github.com/brimdata/zui/assets/5934157/c659f541-4243-427d-85fd-fd302e745a6c

philrz:
I'm currently testing with this branch at commit cb3c015.

In the attached video I'm attempting Preview & Load with different combinations of the "sample.tsv" and "sample.zng" (from the Zui repo) and "prs.json" (from the Zed repo). What I see is that each of these loads fine individually, but if I attempt to load the combination of sample.tsv+prs.json I get an error "too few values" and with sample.zng+prs.json I get an error "unknown compression format 0x22".

I'm guessing the app is concatenating together the file inputs rather than naming them one at a time, which would present a problem because it would cause Zed's auto-detector to lock in on the input format based on the first bytes it sees. I can repro the same failure messages at the CLI:

```
$ cat sample.tsv prs.json | zq -z 'count()' -
stdio:stdin: line 77: too few values

$ cat sample.zng prs.json | zq -z 'count()' -
stdio:stdin: zngio: unknown compression format 0x22
```

But they work if I enumerate the filenames separately so the auto-detector handles each separately.

```
$ zq -z 'count()' sample.tsv prs.json
31(uint64)

$ zq -z 'count()' sample.zng prs.json
32(uint64)
```

https://github.com/brimdata/zui/assets/5934157/df946c49-b565-4127-adcc-db9ae83f142e

philrz:
I'm currently testing with this branch at commit 7ccf151.

In the attached video I'm initiating the Preview & Load workflow for the [schools.zson](https://github.com/brimdata/zed/blob/main/testdata/edu/schools.zson) test data from the Zed repo. After selecting the file, the bottom status bar shows the correct count of 17686 records in the original data, but scrolling to the bottom shows we're still setting just the first portion of the input data. @jameskerr has confirmed that indeed only the first 100 rows are shown.

```
$ zq -z 'head 100 | tail 1 | cut School' schools.zson 
{School:"Academy of Personalized Learning"}
```

Indeed, this sampling is surely adequate for most preview/shaping purposes. However, it would be helpful to provide some explicit detail about this, e.g., have the status bar say something like "17686 Rows (first 100 shown)".

If we wanted to be extra fancy, making the number user-configurable to something other than 100 might be handy for some use cases. One of the limitations of the current approach is that if multiple files are selected, and the first file contains more than 100 rows, none of the contents of the additional files will be observable in the Preview & Load workflow. This could present a challenge if, for example, the additional files had different data types such that the user would want to be able to work with them as part of the interactive shaping exercise.

https://github.com/brimdata/zui/assets/5934157/e5e97ac5-6ae1-4a6e-888c-6414642406bc

philrz:
I'm currently testing with this branch at commit 7ccf151.

In this test I was attempting to see what the user experience would be if there's a syntax error in the Zed of my Shaper Script. As you can see, I'm shown a "ZQ Error".

image

Thinking about this as a Zui user, the message may or may not make sense. While we're hopeful that most Zui users will have taken the time to look at Zed docs (and hence be aware of what `zq` is), it's not a given that all users will be familiar. Even if they are, it may not be immediately obvious to users how the Preview & Load workflow is invoking `zq` behind the scenes to preview the files.

In thinking about how we might improve, the best idea I've got is to draft something for the Zui docs site that explains the feature in brief and describes how `zq` is invoked behind the scenes. We could put some kind of "What's this?" or "Help" hyperlink to the doc on this screen in a prominent place. That article, in turn, could link off to the [zq Tutorial](https://zed.brimdata.io/docs/tutorials/zq) and [Shaping docs](https://zed.brimdata.io/docs/language/shaping) so users could learn as much as they want about the CLI tooling and Zed language. Another advantage of this approach is that it would help us educate users to how they could construct similar `zq` command lines on their own outside the app for shaping and other purposes. I'm happy to write the article once the feature is mostly finished.

philrz:
I'm currently testing with this branch at commit 7ccf151.

The attached video shows that the **Message** portion of the configurable **Commit Settings** don't seem to be making it through intact.

https://github.com/brimdata/zui/assets/5934157/59a65aa0-2807-4ada-acde-f42b9f49f267

philrz:
I'm currently testing with this branch at commit 7ccf151.

While I expect this new Preview & Load workflow will be loved for almost all use cases, I've wondered if existing users might be nostalgic for Zui's current quick drag & drop that then jumps straight to seeing fully-loaded data & being able to perform searches. For small amounts of data where things like shaping or having a proper pool key don't matter, this can indeed hit the spot.

@jameskerr and I have talked about a couple lightweight ways we might still make this available.

1. When the giant "Preview & Load" graphic is shown when the user is in the middle of a drag & drop into Zui for what would be data to load to a new pool, we could add a message saying something like "(hold down Shift for a quick load to a new pool with default settings)". This would educate users that might be interested in such an option.

image

2. Similar for when adding data to existing pools, @jameskerr noted that in the blank space shown in the red box in this screenshot we might eventually add areas as drag & drop targets for additional data. For instance, there might be a picker for recently-used shapers for the given pool (ideally defaulting to the last-used one), and hence a spot where the user could drag additional files to have them immediately loaded to the existing pool using the selected shaper.

image

philrz:
I'm currently testing with this branch at commit 67e02f5.

As shown in the attached video, I happened to try bringing up the Settings/Preferences while the Preview & Load screen was open, and it behaves like a no-op. Once I cancel out of the Preview & Load I can see that it looks like it _did_ actually open up when I clicked it, but I guess it was hidden behind the Preview & Load screen. But it felt at first like something was broken, and I expect new users may react the same.

I'm not sure what our options are here, e.g., if it's too complex to actually have multiple of these stacked atop each other? I'm also not sure if there's other screens in the app that are also subject to this same behavior. I stumbled onto this one kind of by accident.

https://github.com/brimdata/zui/assets/5934157/5c675fc1-f84b-4a6d-a5b6-4ed9ecae70a4

philrz:
I'm currently testing with this branch at commit 71e62ba.

It looks like re-working the **Settings** menu to fix one of the issues described above has also caused an unexpected toggle in the default behavior for "Run Query On..." In the attached video, we now see it now behaving with "Shift+Enter" as the default.

https://github.com/brimdata/zui/assets/5934157/c2d6a228-8968-4ff7-9437-7650ff5215dc

Meanwhile, on GA Zui v1.3.0, the default behavior is still that just hitting Enter should be enough to run the queries.

image

This all makes me also realize that, in addition to preserving the default, we'll also want to make sure we preserve any prior change from the default that a user might have made to the setting.

philrz:
I'm currently testing with this branch at commit 71e62ba.

After loading some common test data like `sample.zng` using this branch, I noticed that the data was presented in the opposite order of what I'm accustomed to seeing. I realize now that this is because the default **Sort Order** of the pool key got flipped. Here's how it currently looks on this branch:

image

Whereas this is how it currently looks with GA Zui v1.3.0.

image

Granted, this default is in many ways a reflection of Zui's log-centric/Zeek-centric past, i.e., an assumption that users might be loading recent time-sorted logs and hence want to look at the most "recent" activity. Given Zui's current applicability to other general data use cases where keys might be almost anything (e.g., incrementing numeric identifiers), "Descending" might make less sense than it once did (and for that matter, assuming a default pool key of a field named `ts` might not either). However, given that the majority of Zui's active user base come from that legacy past, I expect we'd want to maintain the historical default for now. If so, over time we might look at ways users could seamlessly diverge from that, e.g., maybe only use the descending/`ts` default if the data being Previewed actually has a field of that name, and if not, make "ascending" the default and require the user to pick a pool key.

philrz:
I'm currently testing with this branch at commit 71e62ba.

The Preview & Load screen currently defaults to Inspector view. This makes a lot of sense to me since it's well-suited to showing data of different shapes, type decorators are visible, etc.

By comparison, for the first time I just clicked on the Table view from within the Preview & Load and realized it looks preety busted. For instance, the attached video shows it trying to lay out the multi-shaped rows into columns, resulting in things like what look like the `id` address/port fields in a column called `peer`. Also, there's what appear to be some Expand/Collapse buttons in the upper-right that look similar to the ones visible in the Inspector view, but rotated 90 degrees. When clicked, they don't seem to do anything.

https://github.com/brimdata/zui/assets/5934157/c83b2c9a-defc-4e01-9b9b-8846ad8b05f4

philrz:
I'm currently testing with this branch at commit 71e62ba.

I was doing some Preview & Load testing using the JSON file inside [apt29_evals_day1_manual.zip](https://github.com/OTRF/Security-Datasets/raw/master/datasets/compound/apt29/day1/apt29_evals_day1_manual.zip) as my sample data. As the attached video showed, the large size of the data combined with the current UX caused me to make a mistake: When I clicked **Load** there was no indication that the load had actually begun, so after several seconds, I clicked it again wondering if I'd "missed" or something, then even clicked it a third time. As it turns out, a separate load operation was initiated with each click, which is probably not what we want. I'd have instead expected either the **Load** button to become inactive or for the Preview screen to immediately disappear once I click **Load** and put me in some view where I'm watching the load progress.

https://github.com/brimdata/zui/assets/5934157/025e3867-edd9-4273-bb4c-17b64a3a4c53

philrz:
I'm currently testing this branch at commit c09b11b.

Earlier in the life of this branch there was a problem where `zq` processes were left running after clicking **Cancel** out of the Preview & Load screen. That problem has been squashed, but as shown in the attached video, there's a similar feeling issue now such that if **Load** is clicked for a large file and then I click "x" out to cancel that load, a `zq` process continues running. I've reproduced this several times and it seems like the life of the process can vary, i.e., I'm not convinced it always sticks around long enough to have read the entire input file, but it can definitely be there long enough that it's probably worth seeing if it can be cleaned up quicker (e.g., in this case it's there for almost 3 minutes). Also, while I didn't capture it on yet another video, I also saw that the `zq` process will stick around like this if I hit Cmd-q to quit the app and confirm I'm ok with aborting the load.

https://github.com/brimdata/zui/assets/5934157/4adb79b3-cf16-4821-8a79-cbbfc2bb4789

philrz:
I'm currently testing this branch at commit c09b11b.

In the attached video, I start to Preview & Load some multi-shape data and follow the guidance to apply `fuse`. I had become unfamiliar with the functionality of the "Expand all" and "Collapse all" buttons in the Table view so I scratched my head for a bit when I clicked them and it seemed to just toggle the height of the blank space at the top. Then I thought to scroll right and was reminded how it allows for columnar presentation of some embedded records such as `geo` in this Zeek data. Problem solved!

However, then I found that when I loaded my fused data, the equivalent Expand button in the regular Table view for query results is grayed out. I can't see why it would be working in one place and not the other. I then checked the tip of `main` and it looks like it's grayed out there too. This doesn't seem like a crisis and we could surely postpone dealing with it. But since it seems to be working in Preview & Load in this branch I figured I might as well point it out in case it'd be easy to fix in the main results view as well so it'll be consistent.

As long as I'm talking consistency, I'll also note that the tooltips for these buttons in Preview & Load just say "Expand All" and "Collapse All" whereas in the main results view they're "Expand all container values" and "Collapse all container values". 🤷‍♂️ 

https://github.com/brimdata/zui/assets/5934157/f35c36b2-0d20-451d-857a-2c66d54f235a

philrz:
I'm currently testing the branch at commit b67c794.

I spotted a little thing noted in the attached screenshot. While the prior pool key defaults of `ts`/Descending were restored in the Preview & Load page, I see that in this "a la carte" pool creation pop-up it's still defaulting to Ascending. Would be good to switch that around for consistency.

image

philrz:
I'm currently testing this branch at commit b67c794.

I was just doing some back-to-back loads to different pools and noticed something curious shown in the attached video. When my first load is of big data to a new pool, once I click **Load** I'm brought to the new pool's summary page where I can watch the progress thermometer fill up, click "x" to abort, etc., which is all great. If I now initiate a load to yet _another_ new pool, I'm _not_ brought to that new pool's summary page. Instead the pool name appears in the left sidebar and I can click it to bring up the pool summary page if I want to watch/abort.

I suspect this might have been an intentional design decision since surfacing lots of pool summary pages has felt cluttery in the past and were of questionable value since they didn't communicate much info offer much functionality. However, now that there's more to see & do on them, I figured I'd point this out in case there was an oversight.

https://github.com/brimdata/zui/assets/5934157/cfb37374-8a4a-46ad-86a8-260aef4e6964

philrz:
I'm currently testing this branch at commit b67c794.

I bumped into the bug shown in the attached video quite by accident. Repro steps:

1. Drag a piece of data in for Preview & Load
2. Expand the **Pool Settings**
3. Enter a value for **Pool key**
4. Hit **Enter**. Notice how the outputs for Original/Preview go blank, which is unexpected.
5. Click **Load** and find that the data load fails

If I skip step 4 and just click **Load** of course the load goes fine. I'm not sure what hitting **Enter** at that step actually triggered, but since **Load** is the prominent button on the screen I could imagine a user expecting that hitting **Enter** would be the same as clicking the button, so something is definitely buggy here.

https://github.com/brimdata/zui/assets/5934157/abc34f3f-e5c5-4d96-b36b-7437e9c7ac37

philrz (approved):

PR #2835, opened 2023-08-09
If insiders release workflow finds same SHA, exit early with success

philrz:
Currently when the Zui Insiders release workflow finds the SHA has not changed since the last release, it exits as a red failure. Since PRs merge to `main` on Zui infrequently, this is likely to happen often, so I'm concerned this would get us out of the habit of checking the outputs when they "fail" and hence be at risk of missing legit failures. The changes in this PR make it so that the workflow will instead exit early as successful when the SHA hasn't changed.

FWIW, I first tested out the logic in a personal repo's workflow https://github.com/philrz/build-sha-test/blob/main/.github/workflows/foo.yml.

### DIFF
--- a/.github/workflows/release-insiders.yml
+++ b/.github/workflows/release-insiders.yml
@@ -17,17 +17,19 @@ jobs:
           owner: brimdata
           repo: zui-insiders
       - name: Download last build_sha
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Very nice.

PR #2849, opened 2023-09-28
Update zed to v1.10.0, brimcap to v1.5.2, and prep v1.3.0

philrz:
https://github.com/brimdata/zui/compare/v1.2.0...5597a25

### DIFF
--- a/apps/zui/CHANGELOG.md
+++ b/apps/zui/CHANGELOG.md
@@ -1,3 +1,12 @@
+## v1.3.0
+- Update Zed to [v1.10.0](https://github.com/brimdata/zed/releases/tag/v1.10.0)
+- Update Brimcap to [v1.5.2](https://github.com/brimdata/brimcap/releases/tag/v1.5.2)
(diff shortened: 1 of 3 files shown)


apps/zui/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## v1.3.0
+- Update Zed to [v1.10.0](https://github.com/brimdata/zed/releases/tag/v1.10.0)
+- Update Brimcap to [v1.5.2](https://github.com/brimdata/brimcap/releases/tag/v1.5.2)
+- The [Zui GitHub repo](https://github.com/brimdata/zui) is now structured as a monorepo (#2818)
```
nwt, line 4: This strikes me as an odd thing to include in the change log.

nwt (approved):

apps/zui/CHANGELOG.md
philrz: @nwt: I figured if there's users that have hacked with the code (we do show 125 forks, FWIW) they might notice that everything got moved around and hence might appreciate the pointer to where it's all explained.

PR #2863, opened 2023-10-24
Fix Zui Insiders dev builds

philrz:
I tried to use this workflow for the first time in a while and it failed (https://github.com/brimdata/zui/actions/runs/6630692331). I dug into the history a bit and it looks like some changes to the regular Insiders Release workflow in #2833 and #2818 didn't make their way into this workflow. I've made what I think are the necessary changes to get it running again and have successfully built/installed/used an artifact based on this branch (https://github.com/brimdata/zui/actions/runs/6632098539).

### DIFF
--- a/.github/workflows/build-insiders.yml
+++ b/.github/workflows/build-insiders.yml
@@ -26,6 +26,7 @@ jobs:
 
   release:
     name: Build
+    needs: check_latest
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Makes sense to me.

PR #2876, opened 2023-11-02
Fix More Nulls

jameskerr:
- Records can have null fields.
- Fix More Nulls

That emties file now looks like this: 

CleanShot 2023-11-02 at 16 38 47@2x

### DIFF
--- a/apps/zui/src/app/features/inspector/views/null-view.ts
+++ b/apps/zui/src/app/features/inspector/views/null-view.ts
@@ -2,6 +2,14 @@ import {View} from "./view"
 import * as zed from "@brimdata/zed-js"
 
 export class NullView extends View<zed.Any> {
+  get showDecorator() {
+    return !zed.isPrimitiveType(this.type)
+  }
+
+  get decorator(): string {
+    return this.type.toString()
+  }
+
   render() {
     return "null"
   }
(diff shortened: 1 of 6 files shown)


philrz (approved):

philrz:
I clicked Approve because this indeed fixed all the examples shown above. However, there was a yet another variation I floated yesterday to @nwt that I now see is still divergent here from how ZSON shows it.

>```
>$ echo '{"r": [null,{},1]}' | zq -Z -
>{
>    r: [
>        null,
>        {},
>        1
>    ]
>}
>```

This branch at commit 3726e3d still shows an explicit decorator here.

image

I discussed this with @nwt and he pointed out that having an explicit decorator in this case is not _wrong_ necessarily, since the [Type Decorators part of the ZSON spec](https://zed.brimdata.io/docs/next/formats/zson#22-type-decorators) points out where they "might be required" with unions to remove ambiguity, but that does not preclude an implementation from choosing to be explicit and include them even if there's no ambiguity.

That said, I'm still not sure if there's a problem here. It's not like I can cut & paste the output as-is and treat it as ZSON since there's additional decoration unique to the app (e.g., the array indexes like `0:`). But if I drop those and keep the rest, and put the ZSON-ish value from app next to the original ZSON value, I get different `typeof()` resuls.

```
$ cat funky_r.zson 
{r: [null ((int64,{})), {}, 1 ]}
{r: [null,{},1]}

$ zq -z 'typeof(r)' funky_r.zson 
<[(int64,{},(int64,{}))]>
<[(int64,{})]>
```

It looks like it's showing a union inside of a union. @nwt: Is that something that should be addressed?

PR #2881, opened 2023-11-06
Zui Polish

jameskerr:
- Limit autosize columns to 700 width
- Change font size in results and editor
- Use recursive font in editor
- Add grid lines to the table view
- Make rows slightly larger
- Fix whitespace overflow bug in column pane


philrz (approved):
The changes all look solid, though CI is showing red, and it looks maybe related to the `node-pipe` problems tracked in #2882. Feel free to merge when CI is green.

Issue #2882, opened 2023-11-06
Failure to launch Zed service


philrz:
This issue seems to repro starting with Zui and Zui Insiders builds created some time starting around November 3, 2023.

I first reported the symptom in detail at https://github.com/brimdata/zui/pull/2866#issuecomment-1793244921 while I was testing a Dev build based on that branch. However, I've since confirmed that recent Zui Insiders releases starting with [1.3.2-9](https://github.com/brimdata/zui-insiders/releases/tag/v1.3.2-9) all exhibit the same symptom. I also went back and dug up the [Dev Build from this Actions run](https://github.com/brimdata/zui/actions/runs/6726301082) based on commit acb5267 and confirmed it works fine as I remembered it, then just now made [another Dev Build in another Actions run](https://github.com/brimdata/zui/actions/runs/6775178417) also based on commit acb5267 and found _that_ one fails with the symptom. So the problem seems to follow "when the build was made" and not "our code that went into the build".

**Conclusion**: Based on all this, I have to suspect that maybe there's some dependency for which we're pointing at "the latest" and that dependency has advanced to something that breaks us. I have no idea what that might be though!

philrz:
I've studied this a bit more and it seems the symptom varies based on the OS platform.

### macOS

Since I happen to use macOS on the desktop, everything described in the opening text of this issue was first reproduced with macOS. Given this reference:

```
Cannot find module '/Applications/Zui.app/Contents/Resources/app.asar/node_modules/node-pipe/build/Release/pipe'
```

in https://github.com/brimdata/zui/pull/2866#issuecomment-1793244921, then comparing the Actions run log from the working build with the failed one, the smoking gun does seem to reveal itself in the failed build's "Build Zui" step:

```
• cannot build optional native dependency  name=node-pipe
                                             version=0.1.1
                                             cause=exit status 1
                                             errorOut=gyp info it worked if it ends with ok
    gyp info using node-gyp@8.4.1
    gyp info using node@16.10.0 | darwin | x64
    gyp info find Python using Python version 3.12.0 found at "/usr/local/bin/python3"
    gyp http GET https://electronjs.org/headers/v22.3.25/node-v22.3.25-headers.tar.gz
    gyp http 200 https://artifacts.electronjs.org/headers/v22.3.25/node-v22.3.25-headers.tar.gz?force_headers_dist=1
    gyp http GET https://electronjs.org/headers/v22.3.25/SHASUMS256.txt
[...]

PR #2885, opened 2023-11-07
Expect slightly larger VNG export

philrz:
The Zed pointer hasn't advanced successfully for a while due to the `node-pipe` build breakages tracked in #2882. But now that's fixed, we've got legit failures like the one adapted to here which I've traced to the changes in https://github.com/brimdata/zed/pull/4862. Please approve if the VNG file size change is expected.


nwt (approved):
Might make sense to disable this test until VNG is no longer a moving target.

Issue #2598, opened 2022-11-22
Single click to apply each of several saved queries to a pool


philrz:
One of the workflows I became accustomed to in the classic Brim app was the ability to select a single pool, then single-click each of several Saved Queries and see what results came up. This was particularly handy in security use cases when quickly seeing which queries returned some/zero results gave a quick sense of the essence of some pcap/logs (any alerts? any Windows SMB activity? etc.) I've shown an example of this in the attached video.

https://user-images.githubusercontent.com/5934157/203430159-5e27905a-8316-4010-9de0-885210e488ba.mp4

Thus far Zui doesn't yet have quite the same smooth workflow for this. `from` pins and their referenced pools are currently preserved as part of a Saved Query. This means that applying several Saved Queries against a new pool requires setting/replacing the contents of the `from` pin each time in order to see the result. This means one or more additional clicks compared to what we had in Brim, such as needing to click the pool name in the `from` pin pulldown or clicking to the **Pools** list and doing alt-click. These are shown in the attached video.

https://user-images.githubusercontent.com/5934157/203430529-2ff22507-fefd-4c47-b6e2-42253de9a7bc.mp4

Before opening this issue, I searched and found existing issue #2478, and the solution proposed there might do the trick. However, with the alt-click functionality in the **Pools** list fresh in my mind, I thought of yet another approach that might work instead/also: When on the **Queries** list, alt-clicking on a Saved Query entry could copy _just_ the query text from the Saved Query but leave any existing pins (e.g., `from`) in the Query Session in place. This might have some slight advantages over the #2478 approach because it could allow saving the query with a hard-wired `from` pin but effectively override it when needed. I could see this being handy for use cases like if there's a "production" pool for which a query was originally written and you frequently want to apply it against that pool (so, click without alt so it's executed with its pre-populated pool name reference) but you also want to be able to easily apply it to many staging/test pools (now you alt-click).

philrz:
A community Zui Insiders user also asked about this [on public Slack](https://brimdata.slack.com/archives/C03MW6XT7HC/p1676533413300119). In their own words:

> is there a way to use the same "Pin" for the pool for all saved querys? it is annoying to select multiple querys and have to select the pool for each one

I suggested they review this issue and #2478 and provide their input on how they'd like to see this work.

PR #2895, opened 2023-11-15
Zui Polish 2

jameskerr:
Dark Mode
Translucent Mac Window
New Pin Design
New Toolbar Design
New Detail Pane Design
New History Pane Design

New Save Query Workflow

Remaining Work:
- [x] New Icons
- [x] Expand/Collapse Headers in toolbar
- [x] Columns Pane
- [x] Set Max Height on Editor
- [x] Set Max Height on Chart Area
- [x] Move and Delete Code
- [x] Fix Any e2e Tests
- [x] Add background color for windows and linux


philrz:
I'm currently testing this branch at commit 517cea9.

Something seems to be broken with image loading. For instance, this screenshot shows how the Zui logo in the **About** window is currently rendered in black rather than its usual orange.

image

I also happened to create a Dev Build via https://github.com/brimdata/zui/actions/runs/7189168064 and when I run that the logo is not visible at all. I also happened to notice the **+** that's usually rendered next to **phil's Zed Lake** inside the blue circle is missing.

image

I assume this is some general problem so I've not gone hunting for every similar oddity that might be lurking in other screens.

philrz:
I'm currently testing this branch at commit 517cea9 using the Dev Builds from https://github.com/brimdata/zui/actions/runs/7189168064.

Dark Mode seems ok on macOS, but looks like it needs help on Windows and Linux.

Here's how things look on macOS after loading the `sample.zng` test data.

image

Looking at the same on Windows, the sidebars seem effectively washed out.

image

Seems similar on Linux (Ubuntu 20.04 at the moment).

image

philrz:
I'm currently testing this branch at commit 517cea9.

Repro steps for what looks like a bug that's shown in the video below:

1. Load the `sample.zng` test data into the app and click **Query Pool**
2. Open the **Detail** sidebar
3. Follow the app's advice link to`fuse` the results so I can see them in the default Table view
4. Click the top row of results so I can start to see contents in the **Detail** panel

https://github.com/brimdata/zui/assets/5934157/416dc97e-d9be-47d5-85a5-bfc3325b8ae7

As the video shows, when the row is clicked, a column header `this` is placed above all the data and everything gets crammed into a single truncated column. The output of the **Detail** panel also shows a top level "this" next to every value which seems redundant.

By comparison, this video shows the problem not happening at the tip of `main` (commit 4518517 at the moment).

https://github.com/brimdata/zui/assets/5934157/74b9f878-9b1b-4a04-8279-b09dafc15ac0

philrz:
I'm currently testing this branch at 0fd73e7.

As shown in the attached video, the open-in-Wireshark functionality appears to be broken, popping up:

```
Unhandled Runtime Error
Error: Error invoking remote method 'commands.run': RangeError: Maximum call stack size exceeded
```

If it matters at all, here I'm using the test data [ifconfig.pcapng.gz](https://github.com/brimdata/zui/files/13746813/ifconfig.pcapng.gz) (after un-GZIP'ing).

As a possible second order consideration, I ran this same test using a Dev build based on this same commit. I didn't bother capturing it in yet another video, but in that case the error was _not_ popped up and instead appeared only in DevTools. I'm not sure if this is expected behavior, but as a general rule when errors are popped up prominently it makes it more likely that we'll be able to repro/address them when they're initially reported to us, so it'd be great if we could do the same error presentation when in a "real" build.

https://github.com/brimdata/zui/assets/5934157/bba3808d-c158-469e-b13d-a47286b3d442

philrz:
I'm currently testing this branch at 0fd73e7 with the test data [ifconfig.pcapng.gz](https://github.com/brimdata/zui/files/13746813/ifconfig.pcapng.gz) (after un-GZIP'ing).

As the video below shows, we have tooltips for the icons at the top but currently lack tooltips for the icons at the bottom. As long as we're using icons as buttons for key functionality, I find the tooltips very helpful since it's not always obvious from the icon's graphic alone what the button is intended to do, and some buttons are no-ops under some circumstances, making it even more difficult to figure out through trial and error.

A second problem captured in the same video is that after I clicked the button to expand the headers in the table view, the top half of the top results row is cut off by the table headers.

https://github.com/brimdata/zui/assets/5934157/a6742324-0add-4c34-a63f-2597498ace8c

philrz:
I'm currently testing this branch at 0fd73e7 with the `sample.zng` test data.

As the attached video shows, it seems the right-click menu is not working when results are presented in a table. I've turned on an effect of "yellow rings" with each right click to make this obvious. I don't see any errors in DevTools that correspond with the clicks.

https://github.com/brimdata/zui/assets/5934157/742fbef1-e52b-4fbc-a6b8-0a56a568ed74

philrz:
I'm currently testing this branch at commit 0fd73e7.

As shown in the attached video, the **+** button to open a new tab doesn't seem to work at all. I don't see any messages in DevTools that correspond with my left-clicks.

I don't know if this is intentional, but I also noticed a subtle left/right "wiggle" of a pixel or two as I clicked. I don't have a problem with that necessarily, but thought I'd point it out in case it was unexpected.

https://github.com/brimdata/zui/assets/5934157/ab6ed789-c36e-4468-ac7a-4f4105535370

philrz:
I'm currently testing this branch at commit 0fd73e7.

Repro steps shown in the attached video:

1. Load the [conn.zng.gz](https://github.com/brimdata/zed-sample-data/blob/main/zng/conn.zng.gz) sample data
2. Click to get into Inspector mode (since right-click menu isn't working currently on this branch)
3. Select "Count by this field"
4. Click the Back button

The crash:

image

https://github.com/brimdata/zui/assets/5934157/f0f29160-f2b9-4bf8-b31b-b4305d4bc0bb

philrz:
I'm currently testing this branch at commit 0fd73e7.

I noticed that on a "fresh" install, the right-side panel is now open by default, such as shown in this first launch after a fresh install of a [Dev build](https://github.com/brimdata/zui/actions/runs/7293877709) at that commmit.

image

Meanwhile here it is starting out collapsed on a fresh install of GA Zui v1.5.0.

image

I'm not saying this is a problem necessarily, but I seem to remember we've toggled this default once or twice in the past so before I go digging up the history and seeing how we rationalized the change in the past vs. now I'd first check if this is expected.

philrz:
I'm currently testing this branch at commit 0fd73e7 with the `sample.zng` test data.

As shown in the video below, it seems the right-click options to **Sort Asc** and **Sort Desc** no longer include the names of the field that was right-clicked. That's still working at current tip of `main`.

https://github.com/brimdata/zui/assets/5934157/a9ee5764-0c2e-4c18-8d0b-b0c825d87ba7

philrz:
I'm currently testing this branch at commit 0fd73e7 with the `sample.zng` test data.

As shown in the video below, it seems the **Virus Total** right-click option is no longer working. I'm not sure if the feature gets exercised a whole lot at this point, but FWIW It's still working on the tip of `main`. I don't see any messages in DevTools that correspond with my attempted clicks. I've got a "red rings" effect on left-clicks and a "yellow rings" effect on right-clicks to make it clear where I'm attempting to make it work.

https://github.com/brimdata/zui/assets/5934157/148c9496-6f82-4856-a598-76090927f02c

philrz:
I'm currently testing this branch at commit 0fd73e7 with the `sample.zng` test data.

As shown in the attached video, when I attempt to import the Brimcap [queries.json](https://github.com/brimdata/zui/files/13747279/queries.json), an error pops up:

```
Unhandled Runtime Error
TypeError: Cannot read properties of undefined (reading 'select')

Source
src/app/features/sidebar/queries-section/queries-tree.tsx (72:47) @ select

  70 | const [{isOver}, drop] = useQueryImportOnDrop()
  71 | const initialOpenState = useSelector(Appearance.getQueriesOpenState)
> 72 | selectQuery.useListener((id) => tree.current.select(id))
     |                                             ^
  73 | 
  74 | return (
  75 |   <>
```

It seems this doesn't happen if I've already clicked to the **QUERIES** tab in the left sidebar.

https://github.com/brimdata/zui/assets/5934157/65f2aeb9-5112-4215-abe9-a1d789257901

philrz:
I'm currently testing this branch at commit 0fd73e7 with the `sample.zng` test data.

## tl;dr

Maybe it's just muscle memory, but I find myself missing the ghostly outlined "from pin" that used to be there.

## Details

This effect is particularly noticeable when working with saved queries. We're already aware that some users that know us from the "Brim app" security days still miss the way it used to be where a single click on a saved query would render results against a "current pool" (see #2598). As a baseline of how things are currently at the tip of main, this first video shows that in recent Zui after a user has clicked their saved query, they need to perform two additional clicks: One on the ghostly "from pin" and one more to select the pool.

https://github.com/brimdata/zui/assets/5934157/5a679d24-b2ba-43dd-a8be-158e45d0a872

Then this next video shows that it requires five clicks on this branch, which seems heavy by comparison.

https://github.com/brimdata/zui/assets/5934157/835dbbad-0571-4889-a382-df221385f858

philrz:
I'm currently testing this branch at commit 679502b.

This seems like just a cosmetic thing so pointing it out just in case. I'm fine with how the active pool in the left side panel is now highlighted with a white background instead of the blue one that was there before. However the load progress bar to the right of the pool name also happens to fill in white so now during a load it kind of has the effect of "filling in the rest of the white background" instead of the more explicit "filling in a small progress bar" look that was there before. 🤷‍♂️

https://github.com/brimdata/zui/assets/5934157/f4917580-6211-436b-92bd-ff51503d40bd

philrz:
I'm currently testing this branch at commit 679502b.

As shown in the attached video, it seems that selecting the right-click menu option **Show in detail pane** is now just toggling the right sidebar but not actually showing the Details of the clicked item. The current tip of `main` still shows the Details though.

https://github.com/brimdata/zui/assets/5934157/46debe48-ca5a-441b-9331-f157629cbf7a

philrz (approved):

PR #2916, opened 2023-12-01
Add load/export support for TSV

philrz:
Now that TSV read/write support exists in Zed (https://github.com/brimdata/zed/issues/4746) this enables it in Zui as well.

In addition to the tests that will be triggered in CI, I've confirmed locally that all the e2e tests run successfully with this branch.


packages/zed-js/src/client/types.ts
```diff
@@ -70,6 +71,7 @@ export type LoadFormat =
   | 'line'
   | 'parquet'
   | 'vng'
+  | 'tsv'
```
nwt, line 74:
```suggestion
  | 'parquet'
  | 'tsv'
  | 'vng'
```

nwt (approved):

Issue #2915, opened 2023-11-30
Monaco editor doesn't load when airgapped


philrz:
## tl;dr

When running airgapped (i.e., no Internet connection) the editor fails to load. There's Monaco-related errors in DevTools and other errors that don't show up if I repeat the same steps while Internet-connected.

## Details

Repro is with GA Zui [v1.4.1](https://github.com/brimdata/zui/releases/tag/v1.4.1). I bumped into this while trying to reproduce a problem reported in a [community Slack thread](https://brimdata.slack.com/archives/CTT24051C/p1701324006460239).

In the attached video I'm launching Zui fresh after having just installed it and while having no working Internet connectivity.

While the mere launch of the app seems healthy, I do see some errors in DevTools right out of the gate, and I don't see those same errors if I launch when Internet-connected. They may be unrelated to the editor problem, but I figured I'd point them out regardless if we want to better understand their origin and maybe catch/silence these in some way.

The more concerning problem is after I load some data and the editor area is populated with text that just says "Loading..." The DevTools console shows errors related to Monaco, among others.

https://github.com/brimdata/zui/assets/5934157/a5cc39c1-4ab8-40e8-8722-d3beafae2422

If there's some reason why we simply need Internet connectivity for the app to work, we'd certainly be free to set that expectation with users, but ideally we'd want to fail gracefully in a situation like this so the user is informed what's wrong.

SoftTools59654:
I had this problem in a Windows 10 virtual machine

The virtual machine did not have access to the Internet and the data is not loaded

So the problem was not with the virtual machine, the problem was with the lack of internet in the virtual machine

SoftTools59654:
I still have this problem in a system without internet, the latest version of zui insiders has not solved this problem?

philrz:
@SoftTools59654: That is correct. The problem is still being worked on.

philrz:
@SoftTools59654: Just a heads-up that we're about to put out a GA Zui release v1.5.0 but the fix for this issue is unfortunately is not yet complete. We hope to have a fix in a Zui Insiders release soon after. Thanks for your patience.

SoftTools59654:
Is there a solution to this problem?
I installed the latest versions and the problem is still there

philrz:
@SoftTools59654: Correct, the problem has not been fixed yet. We expect to have a fix for it soon. I'll update here when the fix is verified.

philrz:
[...]

PR #2922, opened 2023-12-06
Monaco bugs

### LINKED ISSUE #2921: Cursor position becomes inaccurate while typing in editor
## tl;dr

When typing text interactively into the editor in Zui, the cursor placement can gradually become out-of-sync such that it creeps leftward of the characters being entered. The changes in #2881 likely caused the bug.

## Details

The problem seems to repro starting with Zui Insiders [1.3.2-14](https://github.com/brimdata/zui-insiders/releases/tag/v1.3.2-14), though I've also reproduced in the latest which is currently [1.4.2-7](https://github.com/brimdata/zui-insiders/releases/tag/v1.4.2-7). Also, the community zync user that initially reported the problem was using Windows 11, but I've since reproduced the problem on Windows 11, Windows 10, and Windows 2019 Server.

The repro is shown in the attached video. I simply hit "+" to start a new tab and begin to type a long pool name `from 012345678901234567890123456789`. As I'm typing I can see the cursor start to move leftward of the new characters that I'm entering. After I'm done typing I can hit **Home** and **End** keys to move the cursor to the beginning/end of the line and it's clear that there's multiple characters now rendered to the right of where the cursor lands.

I'll note that I've sometimes managed to get the app into a state where the problem _won't_ repro. The most reliable way I've found to _always_ repro is to uninstall Zui Insiders completely, then clear out the app directories left behind and reinstall fresh. Then after first launch, _immediately_ hit "+" and start typing.

https://github.com/brimdata/zui/assets/5934157/f89b875f-67ab-42a4-8626-e43da8494e0b

In terms of narrowing down a likely root cause, the user that first reported the issue said:

> The ZUI version 1.3.2-13  work fine, after this version, i have this problem.

I found this was indeed the case. I first used the same repro steps as in the video with Zui Insiders [1.3.2-13](https://github.com/brimdata/zui-insiders/releases/tag/v1.3.2-13) and it did _not_ repro. Then I used [1.3.2-14](https://github.com/brimdata/zui-insiders/releases/tag/v1.3.2-14) as shown in the video and it _did_ repro. The Actions run for building `1.3.2-13` was at https://github.com/brimdata/zui/actions/runs/6789083410 and shows it to be based on Zui commit d29e674, and `1.3.2-14` was built at https://github.com/brimdata/zui/actions/runs/6793346596 and shows it to be be based on Zui commit 2ffbdd8. The report https://github.com/brimdata/zui/compare/d29e674...2ffbdd8 shows what changed between those Zui commits, so based on that, it seems almost certain that the changes in #2881 caused the bug.

jameskerr:
This PR fixes two things.

1. ~~The monaco files are now served from the public folder of the app.~~
2. The font size is remeasured whenever a new font loads on the page.

Notes, I tried to use the npm package as the source for the editor to load, but that did not work with Nextjs trying to server render things. Some dom apis were not available and it broke.

Fixes #2921


philrz (approved):
I ran some tests with a Dev build based on commit a7b609e from this branch ([Actions run](https://github.com/brimdata/zui/actions/runs/7119464051)) and it seems to fix both #2921 and #2915 so I'm a functional 👍 on this merging. A couple things I'll point out though:

1. When I ran the test of the fix for #2915, the editor now loads fine, but I did still see the attached failed `GET` in DevTools. If it's benign I'm ok letting it happen, but figured I'd point it out.
 
image

2. It's not clear to me why we went from local port `3000` to `4567` as part of these changes. Perhaps that's explained by your opening PR notes and I just don't see how it relates. This actually seems like a win for me regardless because Grafana happens to run locally by default on `3000` so we'll no longer clash with that port. That said, I've often seen it advocated that rather than putting magic numbers like `4567` directly into the code they should maybe be defined as constants somewhere and referenced by name so that way if we ever make a change like this again we'll only have to make the change in one place. If there's some reason why that's not easy in this case, no problem, but figured I'd point it out. 😄

philrz:
@jameskerr: I just tested again now using the Dev build artifact from https://github.com/brimdata/zui/actions/runs/7146094385 that's at commit 5aeb461 of this branch. RE: #2915, what I observe looks much the same as what I reported in my prior test [above](https://github.com/brimdata/zui/pull/2922#pullrequestreview-1768607163): There's no errors in DevTools when I first launch the app, but that same `GET` error as before appears once I drag a file into the app and Preview & Load comes up, then it appears again when I click Load and am able to start querying. And as before, the editor otherwise gives the feeling of being healthy, so if we think these message are benign I'm fine living with them for now since the bugs being fixed here seem to be having more impact than noise in the DevTools console. But I don't know what other implications there might be so I'll let you make the call on if you wanna go ahead and merge with what's here. Just based on the tests, my Approve stands. 😉

image

philrz:
Due to the #2915 fix being more involved than anticipated, @jameskerr and I discussed offline backing out that part of the changes and using this PR to deliver just what's needed to fix #2921. To that end, I just tested with a Dev build from https://github.com/brimdata/zui/actions/runs/7146603945 that's based on commit e98ff5d from this branch. It does indeed still address #2921. However, I used the opportunity to confirm that the symptom of #2915 is the same as it was before, and it looks like it's actually gotten worse: Whereas before the editor simply became unusable (hung at "Loading...") when running airgapped, now if launched when airgapped the entire app window is a blank white. Granted, the app was not particularly usable airgapped before either so this is arguably not much different (and, if anything, the fact that the breakage is less subtle is almost a blessing since it doesn't give the user any false impression that the app is going to be usable airgapped). But if this was unexpected it might be worthy of additional study before we merge.

image

philrz:
I just did one more test, now with the Dev build from https://github.com/brimdata/zui/actions/runs/7147092706 which is based on commit 591e415. It showed that #2921 is still fixed, and the #2915 symptom is back to where it was with GA Zui v1.4.1, i.e., the app seems to load ok, but the editor hangs with "Loading..." which will be fixed separately.

Looks to me like it's good to merge. 👍

philrz (approved):

PR #2928, opened 2023-12-11
Copy Paste Data

### LINKED ISSUE #2909: Data import directly from paste buffer
## tl;dr

Particularly with the coming support for TSV format, Zui users would benefit from being able to import data directly from their paste buffer.

## Details

At the time of the filing of this issue, Zui is at commit 3a66931.

https://github.com/brimdata/zed/pull/4891 will soon bring support to Zed for easily reading TSV data. Once that lands I intend to add TSV to the list of import/export formats in Zui (#2908).

Zui users that have asked for TSV support have primarily been users of Excel or Google Sheets that have pointed out when doing cut & paste from these apps the data in the paste buffer is of TSV format. The attached format shows where things are headed at the Zed layer using the branch from https://github.com/brimdata/zed/pull/4891.

https://github.com/brimdata/zui/assets/5934157/4361e34a-aa84-4ebb-b63b-ecc3f71312b3

As this shows, it'll now be easy to read this TSV data, but the user needs to take the extra step if pasting it into a file. Such a file could be easily dragged into Zui, but even better would be if they could skip the file altogether and go directly from the paste buffer into Zui. I'm hopeful that this will flow nicely with the new Preview & Load experience, i.e., instead of having the `zq` take its input from files on a command line, just give it a stdin stream of what's in the paste buffer.

In terms of how to invoke this, I'm not yet certain. My first thought was to add another pull-down menu item similar to the **Open Data File...** (e.g., "Open From Paste Buffer"?) @jameskerr also wondered about taking over the **Paste** option that's in the **Edit** menu, which does seem like it would be possible in contexts where it's not already claimed for something else (e.g., when the Editor is active).

jameskerr:
Closes #2909

- [x] Integration Test
- [x] Unique File Name
- [x] Delete File after Loading
- [x] Allow Pasting More and More data into Preview And Load
- [x] Read the directory for the files names, don't store the this.files array
- [x] Change DropFilesTarget to use Dialog and ShowModal so that it will go to the top of the screen.
- [x] Destroy the folder when the app quits

Bugs:

- [x] Creating the temp dir needs to work
- [x] The end to end test ends too fast before the ingest finishes.


philrz (commented):
This is great! Some suggestions before we merge.

1. Could we have a simple e2e test for this? I'd like to avoid accumulating more features that we'll only be able to catch manually if they regress.

2. It looks like the temporary file created to hold the contents of the paste buffer is left behind after the operation. Can we make sure that gets deleted? I know we've rationalized things like leaving behind the temp files we create for pcap extraction because users might keep working on them in Wireshark, but for this case I don't see a reason to let them accumulate (and we've had users come to us concerned about droppings we've left behind in temp). My bigger concern is that there might still be environments (particularly Linux) where multiple desktop users live off the same host. Right now when running with defaults on Linux, Zui's temporary files land in `/tmp`, so it seems like two different users both using this feature have the potential to step on each other's `/tmp/clipboard.txt` (whether that be one user's temp file overwriting someone else's or one user's temp file not able to be created because they lack the permissions to stomp the other user's temp file).
 
3. Instead of the paste buffer always going to a static file called `clipboard.txt`, could we use one of those functions that's guaranteed to make a temp file with a unique name? Even if we do what I suggested in my prior bullet I can imagine scenarios (e.g., crashes) where temp files would still get left behind and hence leave the potential for things like I described in the prior bullet.

4. The way it's working now I noticed that if I start Preview & Load based on the contents of the paste buffer, I can click "Add" and lump on more files from disk to the single load operation, which is pretty cool (see video below). Unfortunately it seems I can't do the reverse, i.e., initiate Preview & Load with one or more regular files and then lump on the paste buffer contents in addition to that, since if I'm at the Preview & Load screen and invoke "Paste Data..." it _replaces_ all the files currently staged for load just like if I dragged a regular file into an already-open Preview & Load screen. I expect this feature will primarily be used for one-off pastes of data like spreadsheets similar to how we've been testing it so it probably doesn't make sense to do anything fancy like add a separate **+Add Paste Buffer** button on the Preview & Load screen. But I thought I'd point it out just in case.

https://github.com/brimdata/zui/assets/5934157/1f08f1ca-2b45-480c-9732-6189d89986c8

packages/zui-player/tests/copy-paste.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    await app.evalMain(({ clipboard }) => clipboard.writeText('{b: 2}'));
+    await app.invoke('loads.paste');
+    await app.evalMain(({ clipboard }) => clipboard.writeText('{c: 3}'));
+    await app.invoke('loads.paste');
+
+    await app.attached('listitem', 'paste');
+    await app.attached('listitem', 'paste_1');
+    await app.attached('listitem', 'paste_2');
+    await app.click('button', 'Load');
+    await app.attached(/Successfully loaded into jkerr_pastes/);
```
philrz, line 36: The e2e test failed for me because I'm not `jkerr`. 😉

apps/zui/src/domain/loads/temp-file-holder.ts
```diff
@@ -0,0 +1,41 @@
...
+import path from "path"
+import * as fs from "fs-extra"
+import {getUniqName} from "src/util/get-uniq-name"
+
+export class TempFileHolder {
+  dir: string
+
+  constructor(namespace: string) {
+    this.dir = path.join(os.tmpdir(), namespace)
+    fs.ensureDirSync(this.dir)
```
philrz, line 11: I can see that this is now creating the multiple paste files in a temporary directory whose name is derived from the username, e.g., `phil_pastes` for me. Revisiting my PR comment about concern that in a Linux style environment there might be collision among the temp files of multiple users, it seems that (at least in theory) that collision exposure now exists among these temp directories. Since it looks like we're leveraging the user's OS username I guess a collision is nearly impossible since OS usernames are assumed to be unique. If it's not difficult to do, I'd feel less paranoid if we were using `fs.mkdtemp()` in some way since that adds additional uniqueness, e.g. like in this sample program:

```
const fs = require('fs'); 
  
fs.mkdtemp("phil_pastes-", (err, folder) => { 
  if (err) 
    console.log(err); 
  else { 
    console.log("The temporary folder path is:", folder); 
  } 
});
```

apps/zui/src/util/fixed-positioner.ts
```diff
@@ -0,0 +1,119 @@
...
+  const targetRect = target.getBoundingClientRect()
+  const leftMin = 0
+  const leftMax = doc.clientWidth - leftMin
+  const topMin = 0
+  const topMax = doc.clientHeight - topMin
+  const [anchorX, anchorY] = parsePoint(anchorPoint)
+  const [targetX, targetY] = parsePoint(targetPoint)
+  const margin = parseMargin(targetMargin)
+
+  console.log({a: anchorRect, t: targetRect})
```
philrz, line 29: Was this left here by mistake?

philrz (approved):

PR #2934, opened 2023-12-20
Update Electron

jameskerr:
- Update Electron To Latest
- Update playwright to @next until 1.41 is released
- Simplify playwright invocations


apps/zui/src/electron/start.test.ts
```diff
@@ -14,14 +14,6 @@ jest.mock("@brimdata/zed-node")
...
 afterEach(teardown)
 
-test("start is called in zed lake", async () => {
-  const appMain = (await main({
-    devtools: false,
-    autoUpdater: false,
-  })) as MainObject
-  expect(appMain.lake.start).toHaveBeenCalledTimes(1)
-})
-
```
jameskerr, line 24: This test was causing problems and is not necessary. If the lake doesn't start, many many other tests will fail.

apps/zui/src/js/flows/lake/buildAndAuthenticateLake.ts
```diff
@@ -49,14 +49,13 @@ export const buildAndAuthenticateLake =
         return [false, null]
       }
 
-      const dialogOpts = {
+      const dialogChoice = await showMessageBox({
```
jameskerr, line 52: TypeScript liked this better.

apps/zui/src/electron/run-main/before-boot.ts
```diff
@@ -6,16 +6,18 @@ import {MainArgs} from "./args"
...
   args: Partial<MainArgs>
 ): Promise<string | null> {
   // Disable security warnings
-  process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = "true"
+  // process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = "true"
   // Setup app paths, this must be first
   appPathSetup()
   setLogLevel()
   // Disable for certain platforms
-  app.disableHardwareAcceleration()
```
philrz, line 18: This rang a bell when I saw it and so I did a search and just dug up #1248 where we disabled it to fix some specific issues. If it's absolutely necessary we revert this we could invest some time to try and reproduce the kinds of issues we'd fixed with this in the past and see if they've been chased away for some reason. But given how many other unsolved problems we've got on our list I'd rather avoid bringing back old ones if we can.

philrz (approved):

PR #2943, opened 2023-12-20
Update zed to v1.12.0, brimcap to v1.5.5, and prep v1.5.0

philrz:
https://github.com/brimdata/zui/compare/719d262...4caf78e


jameskerr (approved):
Love it

PR #2866, opened 2023-10-31
Update Modes

jameskerr:
Add different modes to auto update.

- [x] Check what "vibracy" looks like on windows
- [x] Write code to handle linux auto updates.
- [x] Make sure it works as expected.
- [x] Update the "download progress code"

PR #2950, opened 2023-12-30
Docs for new Update features

### LINKED ISSUE #2905: Docs: Updating Zui
Once the changes from #2866 have been exhaustively tested, we need to update the docs to describe how app updates work.

Things to make sure are covered (not a complete list):
- The default behavior of auto-update (as it has been in the past)
- How the different **Settings** options control/disable auto-update
- The manual **Check for Updates...** option in the pull-down menu
- How Linux differs from macOS/Windows

Once this exists, we should link to it from the [Installation](url) page.

philrz:
These docs updates cover the changes from #2866.

At first I'd thought about covering this in a wholly separate doc and maybe a video, but once I thought harder about the scope of the changes I realized that some minimal tweaks to the existing Installation doc with a couple more screenshots would probably do the trick.

Closes #2905


jameskerr (approved):
Dig it.

PR #2953, opened 2024-01-05
Fix Broken Title Bar Buttons

jameskerr:


apps/zui/src/components/icon-button.tsx
```diff
@@ -24,6 +24,7 @@ const BG = styled.button`
   justify-content: center;
   flex-shrink: 0;
   transition: background 100ms;
+  -webkit-app-region: no-drag;
```
jameskerr, line 27: This was the fix.

jameskerr: The app thought these buttons were draggable regions and intercepted all mouse events.

philrz (approved):
Looks good. Thanks for the test! 👍

Issue #2926, opened 2023-12-08
"Error: write EPIPE" or "Error: write EPIPE" on attempted pcap import


philrz:
## tl;dr

An attempt to import a pcap containing 802.11 traffic resulted in a JavaScript error pop-up. I suspect the root cause is a Zeek failure that bubbles up through Brimcap. It would seem ideal if we could catch and surface that error instead.

## Details

Repro is with Zui commit 29860a0.

This issue was reported in a recent [community Slack thread](https://brimdata.slack.com/archives/CTT24051C/p1701994378660399). In the user's own words:

> Hi everyone, new here. Im trying to use zui to view some a pcap file for a cyber security challenge. Installed zui and then tried to import the pcap file but i get this error. same error on my windows 10 and my linux vm. Can anyone help me solve this issue?
>  ![image](https://github.com/brimdata/zui/assets/5934157/c8f41cce-cf13-4350-8c3d-cd3b1e4caeb0)

The user shared their pcap [VanSpy.pcapng.zip](https://github.com/brimdata/zui/files/13607802/VanSpy.pcapng.zip) which I used to reproduce the problem with Zui 29860a0 on macOS as shown in the attached video. In my repro it happened to show `Error: write EPIPE` as the exception instead.

https://github.com/brimdata/zui/assets/5934157/a23d1039-929b-459c-b411-236359e5dc02

As it turns out, if I dismiss the pop-up I can see a handful of Zeek events were generated based on the pcap. However, since the user was clearly expecting to see more, the pop-up gives the impression that the limited data was possibly due to a bug.

After seeing this, I opened the pcap file in Wireshark and could see the likely problem: All the traffic is 802.11. We've confronted this before and a prior issue asking about 802.11 (#570) was closed a long time ago when the team concluded that there's really not much that can be done here since Zeek and Suricata seem to expect IP traffic. In the time since I can see some 802.11 references in the Zeek repo so maybe there's potential for something there, but we're currently stuck with the older Zeek/Suricata artifacts we've got.

After seeing this, I dropped to the shell and sniffed the `brimcap` command line out of `ps` and tried to run it by hand. Here's what I saw:

```
$ export BRIM_SURICATA_USER_DIR="$(pwd)"

$ cat VanSpy.pcapng | /Applications/Zui.app/Contents/Resources/app.asar.unpacked/zdeps/brimcap analyze -json -
[...]

PR #2955, opened 2024-01-07
Improve brimcap error handling

nwt:
Zui displays a Node exception if brimcap exits without reading its standard input to EOF.  Display an error toast with details instead.

Update, here's how the branch looks now.

https://github.com/brimdata/zui/assets/3460638/5d6de8d6-535b-4ecb-bc07-8276d9c6cc9b


philrz:
I tested out the branch with the pcap from https://github.com/brimdata/zui/issues/2926#issuecomment-1879898700. The error presentation is improved enough that I'd consider this a fix for #2926.

image

apps/zui/src/util/error-to-string.ts
```diff
@@ -15,6 +15,9 @@ export function errorToString(e: unknown) {
     if ("error" in e && typeof e.error == "string") {
       return e.error
     } else if ("message" in e && typeof e.message === "string") {
+      if ("details" in e && Array.isArray(e.details)) {
+        return [e.message, ...e.details].join("\n")
```
nwt, line 19: @jameskerr: react-hot-toast ignores these newlines (as well as any embedded in the strings in `e.details`) so these notifications aren't always as readable as they could be, as you can see here. Got any ideas for how we can improve on this?

image

philrz:
I'm currently testing this branch at commit 23af3d0.

The attached video shows we're now persisting the error that bubbled up from Brimcap and it's well-formatted, which is great. However, a JavaScript "Uncaught Exception" and stack dump still pop up first and that needs to be dismissed, which was unexpected. I made a Dev build just to double check that this wasn't one of those "it will show up only in DevTools in a real build" situations, but it appeared there too.

https://github.com/brimdata/zui/assets/5934157/b8f76d67-968d-4a93-98af-9d1b9a401100

philrz:
@jameskerr: The e2e test looks good. FYI, I made a temporary branch based on this branch to which I merged `main` so I could have the knobs available in my improved `e2e.yml`. With that I did this [Actions run](https://github.com/brimdata/zui/actions/runs/7617421001) that shows your e2e test running reliably on all OSes in CI.

At this point I think I'm good with this one merging, but since @nwt was already involved in this code and I know the Zui/Brimcap error channel has some complexities, I'd feel better if he had a chance to give it a 👍 first. I guess since he opened the PR he technically can't be the one to click "Approve" but I'll wait for him to give the word somewhere and then I'll do the clicking.

philrz (approved):
@nwt and I talked offline and he didn't feel he had more to add to the review, so I'm hereby giving it the 👍.

PR #2961, opened 2024-01-11
Add flexible inputs to Actions Workflow for running e2e tests

philrz:
I'm hoping to invest some time to better understand and hopefully fix some of the problems we have with e2e test reliability in GitHub Actions CI. As a step toward that, in this PR I've added some flexible Inputs to the "Run e2e tests" Workflow. After this change, if the Workflow is invoked with all inputs at their defaults, it'll still run the e2e tests the same way they are in the "Advance Zed" workflow. But those inputs can now be varied to run any command line (e.g., `yarn e2e` ...) on any set of OS platforms, which makes it easier to repro specific problems and confirm fixes.

For instance, in [this run](https://github.com/brimdata/zui/actions/runs/7483368374) I ran only the `copy-paste` tests and did it across all OSes.

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -2,14 +2,22 @@ name: Run e2e tests
 
 on:
   workflow_dispatch:
-
+    inputs:
+      run-target:
+        description: Command line for running e2e tests
+        default: 'yarn e2e:ci'
+        required: true
+      platforms:
+        description: OS platforms to test on (list of strings in JSON format)"
+        default: '["ubuntu-20.04"]'
+        required: true
 jobs:
   run-e2e-tests:
     name: Run e2e tests
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -2,14 +2,22 @@ name: Run e2e tests
...
+        default: '["ubuntu-20.04"]'
+        required: true
 jobs:
   run-e2e-tests:
     name: Run e2e tests
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04]
+        os: ${{ fromJSON(inputs.platforms) }}
```
philrz, line 20: I learned this trick from https://github.com/actions/toolkit/issues/184#issuecomment-1666870406

jameskerr (approved):
Cool!

PR #2964, opened 2024-01-11
Output the "Inputs" used for e2e run

philrz:
Since I plan to make ample use of Inputs in this "Run e2e tests" Workflow, I went looking for the best way to log the inputs that were used for a given run. Of the ideas batted around [here](https://stackoverflow.com/questions/71155641/github-actions-how-to-view-inputs-for-workflow-dispatch) I chose one for this PR. When in use it looks like:

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -21,6 +21,7 @@ jobs:
     steps:
       - name: Output inputs
         run: echo "${{ toJSON(inputs) }}"
+        shell: sh
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup-zui
       - run: yarn lint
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -19,6 +19,9 @@ jobs:
       matrix:
         os: ${{ fromJSON(inputs.platforms) }}
     steps:
+      - name: Output inputs
+        run: echo "${{ toJSON(inputs) }}"
+        shell: sh
```
philrz, line 24: I ask for `sh` specifically because the `echo` in the default Windows shell prints the output with all kinds of ugly escapes.

jameskerr (approved):
Very good.

PR #2968, opened 2024-01-14
Copy from /var/log into folder alongside Playwright artifacts

philrz:
When I started gathering system logs in #2962, this awakened an unfortunate behavior of the `upload-artifact` Action: Once I start asking it to include a root level path in the artifacts bundle, it uses root-based paths for _all_ the artifacts. So for instance on macOS the unpacked artifacts bundle was starting to look like:

image

This just creates lots of extra clicks to get to the good stuff. Therefore in this PR I make a directory alongside the other Playwright artifacts and put the system logs there so now I get the more sane hierarchy:

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -50,6 +50,11 @@ jobs:
           run: ${{ inputs.run-target }}
         env:
           VIDEO: ${{ inputs.video }}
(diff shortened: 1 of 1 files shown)


jameskerr (approved):
Makes sense

Issue #2576, opened 2022-10-28
Help user find the location in the editor for a syntax error


philrz:
Repro is with Brim commit ab5b75d.

Consider the same Zed program as in the #2574 repro. If I had reason to do some kind of edit in the middle of such a wide line, the syntax error such as shown in the following screenshot is not super helpful on its own. We have have handy row numbers in the editor, but there's no way way for a user to quickly land at column 2517 and see their mistake.

![image](https://user-images.githubusercontent.com/5934157/198709940-ab34c603-f23d-4d0d-bb19-c3e8aae9b107.png)

One way I've seen other tools help in this situation is to move the user's cursor to the location in the program indicated by that row/column in the syntax error.

Now, I'm aware that we quietly append some additional Zed on the end of a user's program to help with pagination, and I know we may do even more such changes when we address issues like #2533 and #2567, so this may be easier said than done.

philrz:
The changes in linked PR #3087 provide the "squiggly lines under an error in the editor" envisioned above by @jameskerr. If The red squiggly lines appear below whatever caused the syntax error. If I hover my mouse over that area the details are shown as a tooltip. These tips appear as I'm typing as a user before I've even tried to execute the program.

![image](https://github.com/brimdata/zui/assets/5934157/a9cfb66a-927d-473b-8bdb-96765b9e9d8c)

Another example is this [query from the Zed `join` tutorial](https://zed.brimdata.io/docs/tutorials/join#inputs-from-pools) that I've pasted into the editor. It correctly notifies me that the pools I've referenced don't exist.

![image](https://github.com/brimdata/zui/assets/5934157/007c618e-839f-4307-8cc0-c63580f93ff1)

Here's a few areas where we may continue to improve:

1. When the user _does_ submit a query to be run and it contains an error, the column numbers shown are incorrect, such as in the example below where the `;` is the 13th character from the left in line 1 of the editor but it's listed as 33. Thankfully the caret points at the precise location of the problem so we assume that for now that's more likely to draw the user's attention rather than them counting off characters one by one. 😄 

![image](https://github.com/brimdata/zui/assets/5934157/51a3c035-6de0-456b-8786-946f5f6081b1)

2. Regarding the text-based summary of errors that appears where the query output normally would be, @mattnibs made this suggestion:

   > I suppose you could also the list errors in the results area and maybe have a link that jumps the text edit cursor to the affected area and activates the error hover box?

[...]

Issue #2868, opened 2023-11-01
Choosing a saved shaper


philrz:
At the time this issue is being filed, Zui is at commit 28aede8.

In the first version of Preview & Load, the user is presented with an empty editor in which to draft their shaper. We're aware that this will not always be ideal, and indeed one of our own community members that tried out the feature in Zui Insiders gave this feedback in a [Slack thread](https://brimdata.slack.com/archives/C03MW6XT7HC/p1698246559548409?thread_ts=1698191919.351019&cid=C03MW6XT7HC). In their own words:

> A button to save/ load shaper scripts would be nice :grimacing:

Indeed, when @jameskerr and I exchanged thoughts as the feature was being built/tested, here's a few ideas we came up with.

1. For the case when a user is dragging more data to add to an existing pool, it might be handy if the editor could be pre-populated in this case with a cached copy of the last shaper used for load to that pool. At one point during development the editor contents always held the shaper contents from the last load even if it was for a different pool, and I recommended taking that out because different pools are likely to hold different data/types and hence are likely to need different shapers most of the time. But by comparison, I expect existing pools are often going to see loads of additional data of the same type, so caching the last-used shaper for that pool seems like it would be a handy convenience.

2. In @jameskerr's own words from the task list in #2834, it might be handy to have a dropdown near the editor that offers shapers to choose from. When one is selected, its code would then appear in the editor for possible further editing, in which case being able to Save/Save As might also be justified. Since Saved Queries are already a concept in the app and any Zed program could potentially act as a shaper, I feel like it would be reasonable for the selection mechanism to just be a way to select from all Saved Queries. If a user has a workflow where they have specific, purpose-built shapers, they'd be free to organize them into specific/named folders to suit their needs (i.e., I'd _not_ advocate having some kind of special, pre-allocated Saved Queries area just for "shapers").

[...]

PR #2972, opened 2024-01-17
Remove JS Zed Query Parser

### LINKED ISSUE #2843: Intermittent failure of pool-loads.spec.ts: ".zed-table__cell" resolved to 2 elements
At the time of the filing of this issue, Zui is at commit a231ce0.

While verifying that the fix in #2842 was effective CI, I bumped into a separate intermittent failure that seems worthy of examination. An example of the failure is visible in the Actions run https://github.com/brimdata/zui/actions/runs/5919160618/job/16049163945, and here's the relevant error output:

```
1) pool-loads.spec.ts:26:3 › Pool Loads › load more data into the pool ===========================

    locator.waitFor: Error: strict mode violation: ".zed-table__cell" resolved to 2 elements:
        1) <div id="col:0_row:0" data-column-id="col:0" class="z…>…</div> aka playwright.$("[id="col\:0_row\:0"]")
        2) <div id="col:0_row:1" data-column-id="col:0" class="z…>…</div> aka playwright.$("[id="col\:0_row\:1"]")

    =========================== logs ===========================
    waiting for selector ".zed-table__cell" to be visible
      strict mode violation: ".zed-table__cell" resolved to 2 elements:
        1) <div id="col:0_row:0" data-column-id="col:0" class="z…>…</div> aka playwright.$("[id="col\:0_row\:0"]")
        2) <div id="col:0_row:1" data-column-id="col:0" class="z…>…</div> aka playwright.$("[id="col\:0_row\:1"]")

    ============================================================

       at ../helpers/test-app.ts:98

       96 |   async getViewerResults(includeHeaders = true): Promise<string[]> {
       97 |     const fields = await this.mainWin.locator('.zed-table__cell');
    >  98 |     await fields.waitFor();
          |                  ^
       99 |     let results = await fields.evaluateAll<string[], HTMLElement>((nodes) =>
      100 |       nodes.map((n) => n.innerText.trim())
      101 |     );

        at TestApp.getViewerResults (/home/runner/work/zui/zui/packages/e2e-tests/helpers/test-app.ts:98:18)
        at /home/runner/work/zui/zui/packages/e2e-tests/tests/pool-loads.spec.ts:[35](https://github.com/brimdata/zui/actions/runs/5919160618/job/16049163945#step:8:36):21
```

I've seen this style of failure before (#2602) and I know we've made some other recent changes to tighten up this test (#2687). I reproduced this one in a Linux VM so I could see the contrast between the success & failure cases, and both of these are shown in the attached video.

https://github.com/brimdata/zed/assets/5934157/618077c4-17c5-47bc-a8e2-62e574453022

Looking at that alongside the test code:

https://github.com/brimdata/zui/blob/a231ce0310cf4f69ca56c45f896272d1e9916289/packages/e2e-tests/tests/pool-loads.spec.ts#L34-L36

The `query()` helper waits for the "Fetching" status message to become hidden before it quits.

https://github.com/brimdata/zui/blob/a231ce0310cf4f69ca56c45f896272d1e9916289/packages/e2e-tests/helpers/test-app.ts#L84-L88

If we jump to second 17 in the video and use the left/right arrow keys in QuickTIme, we can jump between the two frames that mark this precise moment of transition: The "Fetching" does indeed go away, and the two lines of query response are immediately replaced by the single line with the `count()` result of `2`. This animated GIF shows that toggle.

Success

Now here's the same toggle within second 41 where the failure happens. Now we see that in the frame that immediately follows the disappearing "Fetching" message, the `count()` result is not yet displayed.

Failure

Given that the failure message speaks of seeing two `.zed-table__cell` elements, my guess is that there's a timing problem where the query output is checked at that precise instance the "Fetching" goes away and it still sees what's left of the prior two lines that were there.

Is there a clean way we can make this reliable? If my theory is correct then I assume a brief sleep timer of just a couple seconds at the end of `query()` would give it all the time needed to fill in the result such that `getViewerResults()` will see the right thing. However, I know that's hacky and would slow down the tests a bit, so hopefully there's a better way. A few thoughts:

1. Is there a better check we can use than "Fetching" becoming hidden? If I trust the essence of the wording, "I am done fetching" is indeed not the same thing as "the result is now available to view", though I think the latter is what the test is hoping would happen.

2. Would it be possible to have some kind unique ID that "tag" issued queries and their results? Looking again at the test code, we've got the call to `await app.query("count()")` and then next the check `const results = await app.getViewerResults()`, but indeed, there's nothing that strictly ties the issued query to the retrieved results, though the use of `await` on both seems to express faith that these will happen serially. However, if `query()` returned an ID and then `getViewerResults()` could be wired up to wait specifically for a query result that's somehow tagged with that ID, it seems that might do the trick.
 
3. If no other ideas surface, perhaps it could be improved in the short term by just having the wait on the disappearing "Load Successful" happen earlier, since that takes a few seconds anyway, so that would give more time for the correct result to fill in before `getViewerResults()` is called. This has limited shelf life because I still have hopes we'll soon get to #2605 and make those go away faster, but even if it becomes possible to click those to make them go away quickly, that might still buy enough time to dodge the problem.

jameskerr:
## Overview

Removes our dependency on the JS version of the Zed Query Parser. 

Fixes #2843

**Behavioral Changes:**

* Infinite Scrolling works on non-summarized results
* Infinite Scrolling does not work on summarized results
* Summarized results only return the first 500 rows

**Technical Changes**

* Query Info is fetched from the current lake's compile endpoint asyncronously and saved in the store.
* Removed LIMIT status from results as well as aggregation and aggregationLimit values
* Created AsyncTasks domain model
* Refactored runHistogramQuery to use AsyncTasks

**Side Effects**

* Zui Tests are more robust by waiting for the results associated with the app's current location key which is now rendered in a data-attribute.
* Major re-organization of the zed-wasm repo and testing frameworks. Tests run on CI now
* Removed unused components/status-bar folder and containing files

**Tradeoffs**
~~* A query is always pushed into history, even if there is an error in the query text.~~

**Do Later**
* Make the default page size configurable
* Add squiggly lines under an error in the editor (#2576)
* Create better keyboard shortcut code that hooks into handler names


mattnibs:
@jameskerr I think we should back out the changes to models/Program then merge this.

packages/zed-wasm/test/browser/parse.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/test/pages/parse.html');
+  await page.getByText('Wasm Ready').waitFor({ state: 'attached' });
+});
+
+test('parse a simple zed string', async ({ page }) => {
+  await page.fill('input', 'zed := "world"');
+  await page.click('button');
+  await expect(page.locator('pre')).toContainText('"kind": "OpAssignment"');
```
mattnibs, line 11: On the zed side of things we avoid not having tests that introspect the contents of the ast since this is often subject to change. I think I'd just have a test here that looked for a parser error and verifies the error message- something like `put x :=`.

jameskerr:
I've updated the display of parsing errors.

CleanShot 2024-02-26 at 11 01 05@2x

philrz:
@jameskerr helpfully did a quick Zoom to walk me through some of the changes and answer my questions about what benefits a user might see from this PR vs. where it's laying foundations for other improvements we may add in the future.

To give some additional color around @jameskerr's last comment about the updated display of parsing errors, by comparison here's how the same error is currently presented in Zui Insiders `1.6.1-13`. So what's in this PR is indeed a vast improvement.

image

@jameskerr: While experiencing the improved error in this branch at commit 3be329a, I did spot a small glitch that's probably worth looking into: That "Fetching Total Rows..." informational message shows up and never goes away.

image

philrz:
Here's another example from testing this branch at commit 3be329a. Compare this screenshot to the one in the opening text of #2576. Once again, this is a vast improvement since it shows the user where the error occurred rather than giving a numeric reference to line & column and making the user go find that.

image

However, in the "foundations for the future" category, this doesn't bring us quite all the way to what's envisioned in #2576. If the syntax error had been thousands of characters "rightward" in the big line of text like is shown in #2576, it would still be quite tedious for the user to scroll right to find the `=== ^ ===` in the error message. But what's in this PR is definitely still solid incremental improvement before we make time to give #2576 the full treatment.

mattnibs:
@jameskerr re @philrz's last comment. Instead of using the error message field returned by compile endpoint, I was thinking we could use `info` field from the response error message e.g.:
```"info":{"parse_error_offset":6}```

It tells you where in the source the parse error occurred and we could pass that to Monaco and have it do the error message natively.

**Comment by philrz** [comment]
When going over this PR on a Zoom earlier with @jameskerr I drilled down a bit on how this branch currently leverages:

> ...the "/compile" endpoint from the default lake which runs locally in Zui.

An ongoing concern I've had for some time is that as more users start to use remote Zed lakes they may encounter difficult-to-debug problems when their local Zui and the remote Zed service have some kind of language/AST incompatibilities. I'm sure there's multiple ways we'll go about addressing this over time (e.g., I assume stricter API versioning could help here) but I've always got my eyes open to ways we can improve incrementally in the meantime.

As @jameskerr confirmed, up until this change Zui has always depended on the local Zed parser that shipped with the app, so relying on the `/compile` endpoint of the Zed that shipped with the app would not be making things worse. However, it seems like we have an opportunity now to use the `/compile` endpoint for a remote lake when one is in use, since this has the potential to catch some of the difficult-to-debug problems I cited above. In brainstorming about possible down sides to doing this, the increased latency of calling `/compile` over the network (which might be a WAN) was the one we could think of, so I set out to quantify this a bit.

I set up an EC2 instance in AWS `us-east-2`, started a Zed lake service on it, and connected to it from my home desktop in SF. I made a branch based on the one in this PR to which I added a patch @jameskerr provided to call out to `/compile` on whatever lake the user is currently connected to ([diffs](https://github.com/brimdata/zui/compare/wasm-parser...wasm-parser-remote-compile)) as well as a timer to log how long the parsing step takes. After loading some sample data to both the local and remote lakes, I set out to test the experience with needing to parse some simple Zed (just `count()`) and some sophisticated Zed (the attached 164-line shaper [sample.zed.gz](https://github.com/brimdata/zui/files/14412259/sample.zed.gz)).

The summary of the findings:

|**Lake location**|**`count()`**|**`sample.zed`**|
|-|-:|-:|
|Local|~4 ms|~75 ms|
|Remote|~150 ms|~300 ms|

The video shows the user experience in terms of visible delay in the app:

https://github.com/brimdata/zui/assets/5934157/22b6d2cd-388a-4241-abf4-1f8d0eca8a05

My conclusion: I think the change is worth making. Yes, speed is important, but I'd argue that correctness in catching/reporting errors is even more important, especially as we start adding more features over time that actually might _change_ the contents of a pool. In terms of how I imagine remote lakes will be used, I'm guessing what I'm doing here with accessing a lake on the other side of the country will be less common, and users that do it will likely become accustomed to other unavoidable delays. I suspect more remote lakes will be "local" in nature, e.g., like our community zync users that I think are all in the same geographic region. Finally, if the observed latency gets to be a problem I could picture fancy optimizations we could do, e.g., have Zui compare what's returned by the `/version` endpoint in both the local and remote lakes and if they match it could use the local `/compile` for parsing (though I'm not advocating doing any of that right now).

@mattnibs: Do you have any opinions to add here?

**Comment by jameskerr** [comment]
I am fine with using the current lake for the /compile endpoint. I can change the order of some of these operations to improve the speed if I know we are going to loose between 150 and 300 milliseconds on parsing.

Also, would it be possible (or does this already happen?) to return the parsing error message and location information from the /query endpoint when you submit a bad query?

I could also parse the query async to improve speed. The only reason for parsing first is to prevent adding to the history stack an invalid query entry. However, I could be hopeful that it will work, and if it does not, remove the entry from the history stack.

**Comment by philrz** [comment]
I did some thorough testing to quantify the benefits from this opening statement above:

> I've added a more robust mechanism for the zui-player tests to wait for the correct set of results after issuing a query. Many of our flaky integration tests were due to this problem.

## tl;dr

My results indicate that these benefits are indeed observable. This should allow us to bring some of our e2e tests back into the mix in CI. A detour during that work also led to a better understanding of separate issue #3021.

## Details

The changes in #2954 included no longer running the tests in [pool-loads.spec.ts](https://github.com/brimdata/zui/blob/main/packages/zui-player/tests/pool-loads.spec.ts) in CI because the tests often failed there but consistently succeed locally. One of the ways it was known to fail is described in #2843, and much of that issue text is speculation about the root cause being a timing issue related to not waiting long enough for results to appear after issuing a query, i.e., the exact improvement in this PR @jameskerr speaks to above. Therefore I set out to quantify if I could observe an improvement in the reliability of the `pool-loads.spec.ts` tests in CI.

Right out of the gate I became distracted because `pool-loads.spec.ts` includes the `bad data displays an error message` test that has nothing to do with checking query results but, as shown in #2954, it's been shown to fail in CI more than any other. I therefore took a detour to study that, which led to opening #3021. Now that's understood, I've excluded that particular test from all study of CI improvements from _this_ PR, i.e., hoping to just observe improvements in the symptoms shown in #2843.

First I did a baseline set of runs in https://github.com/brimdata/zui/actions/runs/8087230620 which was based on recent tip of `main`, removed the test from #3021, and ran `pool-loads.spec.ts` in a loop. On the Ubuntu runner this chalked up 7 failures out of 709 runs before the job got canceled (presumably for running too long) and all of them looked like the #2843 symptom.

Since #2843 speculates that waiting longer for results to appear might make the problem go away, I did another set of runs in https://github.com/brimdata/zui/actions/runs/8087202682 that's just like the prior set except that I added a 5-second sleep timer to each test before checking for the results output. Indeed this made it through 513 runs before the job got canceled and saw only one failure, which had a slightly different look than the #2843 symptom. Perhaps there was some intermittent Actions infrastructure problem that made the runner really, really unhealthy? Not sure.

```
> nx run zui-player:test -g pool-loads.spec.ts

Running 2 tests using 1 worker
TF
  1) pool-loads.spec.ts:17:7 › Pool Loads › load data into a pool ──────────────────────────────────
    Test timeout of 30000ms exceeded.
    Error: locator.waitFor: Application exited
    Call log:
      - waiting for locator('.zed-table__cell') to be visible
       at ../helpers/test-app.ts:120
      118 |   async getViewerResults(includeHeaders = true): Promise<string[]> {
      119 |     const fields = await this.mainWin.locator('.zed-table__cell');
    > 120 |     await fields.waitFor();
          |                  ^
      121 |     let results = await fields.evaluateAll<string[], HTMLElement>((nodes) =>
      122 |       nodes.map((n) => n.innerText.trim())
      123 |     );
        at TestApp.getViewerResults (/home/runner/work/zui/zui/packages/zui-player/helpers/test-app.ts:120:18)
        at /home/runner/work/zui/zui/packages/zui-player/tests/pool-loads.spec.ts:24:21
  2) pool-loads.spec.ts:28:7 › Pool Loads › load more data into the pool ───────────────────────────
    Error: expect(received).toEqual(expected) // deep equality
    - Expected  - 1
    + Received  + 1
      Array [
        "this",
    -   "2",
    +   "1",
      ]
      39 |     await app.sleep(5);
      40 |     const results = await app.getViewerResults();
    > 41 |     expect(results).toEqual(['this', '2']);
         |                     ^
      42 |   });
      43 | });
      44 |
        at /home/runner/work/zui/zui/packages/zui-player/tests/pool-loads.spec.ts:41:21
  Slow test file: pool-loads.spec.ts (39.2s)
  Consider splitting slow test files to speed up parallel execution
  2 failed
    pool-loads.spec.ts:17:7 › Pool Loads › load data into a pool ───────────────────────────────────
    pool-loads.spec.ts:28:7 › Pool Loads › load more data into the pool ────────────────────────────
```

Then for the big money, I did the set of runs in https://github.com/brimdata/zui/actions/runs/8087979145 which loops the test when starting from the branch in this PR, and there's no sleep timers. That made it through 655 runs on Ubuntu in CI and saw _zero_ failures.

## Conclusion

Since these failure rates are already small and might be prone to repro flare-ups based on how Actions is behaving on any given day, they're always a little tricky to study, but the tests do seem to show the desired improvement. I'll mark #2843 as fixed by this PR. If #3021 is still not close to a fix by the time this merges, I may look at breaking off that remaining test into its own file so we can start running the other now-reliable `pool-loads` tests in CI. It also looks like I could then look at revisiting issues #2026 and #2504 since they both depend on being able to reliably check query results.

File: apps/zui/src/app/core/models/zed-ast.ts [hunk 2]
```diff
@@ -34,6 +23,12 @@ export class ZedAst {
     return trunks.filter((t) => t.source.kind === "Pool").map((t) => t.source)
   }
 
+  get groupByKeys() {
+    const g = this.ops.find((op) => op.kind === "Summarize")
+    const keys = g ? g.keys : []
+    return keys.map((k) => fieldExprToName(k.lhs || k.rhs))
+  }
```
philrz, line 30: Just to capture something I think was discussed on Zoom, at some point in the future we'd hope that we could rely on the Zed backend to start providing this kind of summary hinting rather than the app having to pick apart the AST to make its own conclusions. @jameskerr is that accurate?

jameskerr: Yes that is correct.

jameskerr:
## Overview


**Behavioral Changes:**
Infinite Scrolling works on non-summarized results
Infinite Scrolling does not work on summarized results
Summarized results only return the first 500 rows

**Technical Changes**
Removed LIMIT status from results
Created AsyncTasks domain model
Refactored runHistogramQuery to use AsyncTasks

**Tradeoffs**
A query is always pushed into history, even if there is an error in the query text.

**Do Later**
Make the default page size configurable
Add squiggly lines under an error in the editor
Create better keyboard shortcut code that hooks into handler names

philrz:
In the tests I've performed on this branch thus far, I did discuss one finding offline with @jameskerr that I'll summarize here as a formal request before final PR approval. The following tradeoff was disclosed in the opening text:

> A query is always pushed into history, even if there is an error in the query text.

Once I had a chance to test drive this, it concerned me. I find syntax errors in Zed are pretty easy to make, especially as I'm building up more sophisticated pipelines, creating user-defined functions/ops, etc. Since Zed will be a new language for our user base, I expect this will be even more true for them as they're learning. Therefore if the History panels gets polluted with lots of error-filled iterations, I'm fearful that it would discourage users from leveraging the History panel.

@jameskerr indicated that he could commit a change here so that the History would not be populated with programs that failed to parse. He also spoke of the caveat that the bad-syntax programs would still be in the history walked by hitting the Back button, but we agreed this is probably ok: The Back button can be rationalized as "go back to the last thing I did", so that being something with a syntax error seems defensibly ok.

philrz:
I'm currently testing this branch at commit d331ab6.

While performing smoke tests, I happened to notice the border around this **Histogram Settings** pop-up are gone. I've confirmed that this problem actually showed up with the changes in #2985 and we just didn't happen to catch it then. But since we're already changing a lot in this branch, it might be worth including the fix for that.

image

philrz:
I'm currently testing this branch at commit d331ab6.

This is another that was actually introduced with the changes from #2985: The contents of the shaper editor are now persisted, whereas before they were cleared out. I was a fan of the "cleared out" behavior because my different data loads may not have much in common so having to clear out the shaper editor constantly would get annoying. For re-usable shapers I'm holding out for #2868.

https://github.com/brimdata/zui/assets/5934157/47ac964f-d1a2-4bb6-815f-cb96913f19d2

apps/zui/CODE.md
```diff
@@ -81,13 +81,29 @@ _Main Process Initializers_
 
 Code that needs to be run one time before the app starts up can be put in an initializer. An initializer is a file that lives in the folder `src/electron/initializers/`. It must export a function named _initialize(main)_ that takes the Main Object as its only argument. See the FAQ for an example of creating a new initializer.
 
+_Query Session_
+
+This is a type of page that a tab can hold it the app. The page contains an editor pane, a results pane, and an inspector with various tabs related to querying data. These are session history, global history, data details, columns, and more.
```
philrz, line 86:
```suggestion
This is a type of page that a tab can hold in the app. The page contains an editor pane, a results pane, and an inspector with various tabs related to querying data. These are session history, global history, data details, columns, and more.
```

philrz (approved):

PR #2975, opened 2024-01-17
Playwright screenshots for e2e tests

philrz:
Per https://github.com/microsoft/playwright/issues/8208, one of the general Playwright features that's not quite there yet for Playwright-on-electron is fancy screenshot options such as top-level config that would capture them automatically in the event of test failures. However, it's quite easy to take them during debug iterations using a helper function like the one I'm adding in this PR. https://github.com/brimdata/zui/commit/a71fbad9f1a1c46dbd9dfd139086b56c97f5b171 shows an example where I used this helper function in [this Actions run](https://github.com/brimdata/zui/actions/runs/7562784692) to take screenshots during the Exports test right after each export succeeds.

image

image

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -60,7 +60,5 @@ jobs:
         with:
           name: artifacts-${{ matrix.os }}
           path: |
-            packages/zui-player/run/playwright-itest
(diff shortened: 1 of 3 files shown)


jameskerr (approved):
Cool

PR #2976, opened 2024-01-18
Non-Action xvfb

philrz:
While working on the e2e tests I saw notification that the [`GabrielBB/xvfb-action`](https://github.com/marketplace/actions/gabrielbb-xvfb-action) we've been using to run Zui inside X11 in CI has been deprecated. While they point to a maintained replacement [`coactions/setup-xvfb`](https://github.com/coactions/setup-xvfb), someone points out in https://github.com/coactions/setup-xvfb/issues/21 that `xvfb` is now built right into the hosted GitHub Actions Ubuntu Runners. This PR switches us using the built-in `xvfb` instead.

One of the features of the Action was:

> If it detects you're not using linux then your tests still run, but without xvfb, which is very practical for multi-platform workflows.

This is why there's a conditional now in `e2e.yml`. [This](https://github.com/brimdata/zui/actions/runs/7563264061) is a test run with this branch on all three platforms showing that's working as expected, e..g., video recordings in the artifacts.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -48,12 +48,15 @@ jobs:
       - run: yarn build
       - name: End to end tests
         id: playwright
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          options: -screen 0 1280x1024x24
-          run: ${{ inputs.run-target }}
+        run: |
+          if [ $(uname) = Linux ]; then
+            /usr/bin/xvfb-run --auto-servernum -s "-screen 0 1280x1024x24" ${{ inputs.run-target }}
+          else
+            ${{ inputs.run-target }}
+          fi
         env:
           VIDEO: ${{ inputs.video }}
+        shell: sh
       - name: Put system logs alongside other artifacts
         run: |
           mkdir -p packages/zui-player/run/var_log
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -49,7 +49,7 @@ jobs:
       - name: End to end tests
         id: playwright
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -48,12 +48,15 @@ jobs:
...
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          options: -screen 0 1280x1024x24
-          run: ${{ inputs.run-target }}
+        run: |
+          if [ $(uname) = Linux ]; then
+            /usr/bin/xvfb-run --auto-servernum -s "-screen 0 1280x1024x24" ${{ inputs.run-target }}
+          else
+            ${{ inputs.run-target }}
+          fi
```
jameskerr, line 56: Maybe we can use the if syntax for github actions to test the os?


https://stackoverflow.com/a/57948488

jameskerr (approved):

.github/workflows/e2e.yml
philrz: Thanks @jameskerr. Of those alternatives discussed at https://stackoverflow.com/a/57948488, I just now switched over to the `if [ "$RUNNER_OS" = "Linux" ]` approach and tested it out successfully in [this run](https://github.com/brimdata/zui/actions/runs/7575104954/job/20630961009), so I'll go ahead and merge with that.

I like the cleanliness of `if: matrix.os == 'ubuntu-latest'` but as discussed [here](https://stackoverflow.com/questions/60916931/github-action-does-the-if-have-an-else) the Actions `if` doesn't have an `else` and instead requires a whole 'nother `if` step with the logic reversed, which feels ugly all over again, which is why I didn't go there. 😛

PR #2979, opened 2024-01-19
Shortcut for Export

jameskerr:
Adds the shortcut cmd+shift+e to pull up the export dialog.

### DIFF
--- a/apps/zui/src/electron/windows/search/app-menu.ts
+++ b/apps/zui/src/electron/windows/search/app-menu.ts
@@ -70,6 +70,7 @@ export function compileTemplate(
 
   const exportResults: MenuItemConstructorOptions = {
     label: "Export Results As...",
+    accelerator: "CmdOrCtrl+Shift+E",
     click: () => window.send("showExportResults"),
   }
 


philrz:
@jameskerr: Looks great. Alas, because you've now made it so easy to quickly call up Export via the keyboard, I found myself instinctively hitting the arrow keys to move between the formats and **Esc** to dismiss without exporting and found a couple new bugs unrelated to this specific branch. If you'd rather I open new issues we could get to them later. But if they're easy and you're keen to fix them in this PR while you're under the hood, I would not fight that, as I suspect users named Jason might bump into them. 😄 

First, I found that if I bring up Export and then immediately hit **Esc**, I can't seem to bring up Export again. I don't see any errors in DevTools.

https://github.com/brimdata/zui/assets/5934157/41dca892-adf7-490e-bb04-06a294cd59a0

Second, I found that if I bring up Export and then immediately click up/down arrows to switch formats, it jumps somewhere other than I'd expect relative to the default ZNG selection. Specifically, hitting up-arrow jumps me down to ZSON, and hitting down-arrow jumps me up to CSV. 

https://github.com/brimdata/zui/assets/5934157/7b9fce8d-e299-4571-a194-7a7e3926e2b3

jameskerr:
Closing this in favor of #2985 That PR will address these bugs.

PR #2985, opened 2024-01-22
Export Feature

jameskerr:
- Improve brimcap error handling
- WIP
- Fix Analyze
- Refactor Brimcap Loader
- Add unit test for quick-load bad pcap
- Increase timeoutfor this test on CI
- Keep errorToString simpler
- WIP
- Clipboard and Pool Export Comming Soon


philrz:
Note to self: When I test this I'll want to verify that the feature & fixes referenced in #2979 are all present here.

philrz:
I'm currently testing this branch at commit bc60704.

I noticed some odd formatting such as where the "traffic light" buttons are and how the Table/Inspector buttons bleed over to each other.

image

philrz:
I'm currently testing this branch at commit bc60704.

Overall I'm really liking the caching of previous Export settings. Combined with the Export shortcut, this should make it easy for users to repeat common workflows, e.g., repeatedly exporting TSV results to the paste buffer so they can quickly paste them into a spreadsheet.

One exception I might make for this is the settings for pool name, pool key, and sort order when **New Pool** is selected as the export destination. As shown in the video below, the current behavior guides the user toward what will be a repeated attempt to create the same pool and a somewhat ugly error message from the Zed service. This current behavior seems like it would be well-suited for users are repeatedly doing exports to pools that have slightly different names (e.g., exporting to MyPool, then MyPool2, then MyPool3, etc.) but my guess is this will be less likely than a user wanting to create a new pool with all fresh settings. So when they select New Pool or that was the last selected option, I'd recommend clearing those values out.

(That said, when the user's previous operation was an export to an existing pool, I'm still in favor of Export to Pool coming up with that pool pre-selected. I can imagine repeat exports to append values to the same pool as a common operation.)

https://github.com/brimdata/zui/assets/5934157/5f571768-7dc4-433d-bd47-c7a4331e3bba

philrz:
I'm currently testing this branch at commit bc60704.

In the **About Zui** pop up, I spotted these labels like "VERSION" showing up in all caps, which looked a little strange. I checked with @jameskerr and he said:

> Interesting, I did make all <label> tags look like that. But yeah, I don't think those need to be labels.

image

philrz:
I'm currently testing this branch at commit bc60704.

I had a fancy feature idea. Right now thanks to the keyboard shortcut and cached Export settings, it's possible to get data quickly into the paste buffer by invoking the keyboard shortcut, clicking **Copy to Clipboard**, then clicking **Close**. However, if a user gets into a steady rhythm of some repetitive operation (e.g., making small adjustments to their Zed query, then wanting to paste-export each result to a nearby spreadsheet) I could imagine them appreciating an even quicker workflow. Two approaches that come to mind:

1. A right-click on the **Export Results** button could offer a selectable option for "Export Results to paste buffer with last settings"
2. A keyboard shortcut could be added that does the same

We could perhaps do both. We could also avoid doing either and wait to see what users say. 😄

philrz:
I'm currently testing this branch at commit bc60704.

The attached video shows a couple problems with the Export to Pool.

1. Clicking the "X" to cancel the load in progress doesn't actually cancel it. If I wait long enough I see the load eventually succeeds.
2. Only the very beginning of it is shown here, but I noticed that during a long export-to-pool like this, the "thermometer" progress bar doesn't gradually fill up. Maybe the nature of the operation makes it impossible to measure incremental progress? I'm not sure if it's worthwhile, but if we wanted to be fancy, perhaps the UX could be a little different in this case (e.g., show a spinner instead of a progress bar?) to differentiate from what they're accustomed to seeing during Preview & Load where the bar fills up. I mention this because in other systems a progress bar that never fills is often a symptom that something is hanging/broken, so we might want to avoid giving that incorrect impression, especially if it might sometimes remain in this state for several minutes.

https://github.com/brimdata/zui/assets/5934157/14f58681-6a04-4a12-916a-911828c233ee

philrz:
I'm currently testing this branch at commit 54d632e.

As shown in the attached video, the cached name of the last-exported-to pool seems to create a problem when that pool has been deleted.

https://github.com/brimdata/zui/assets/5934157/d9616006-171c-4489-a7ec-1970ba4e96a8

jameskerr:
Fixed the issue when the pool is deleted, and added an indeterminate loader that looks like:
CleanShot 2024-02-20 at 14 37 56

philrz:
I'm currently testing this branch at commit 548487e.

As shown in the attached video, if I click "X" to potentially cancel an Export to pool, but opt to click "Cancel", it still stops the load immediately in a strange state that makes it look like it maybe succeeded, but then an error pops up several seconds later. I'd have expected the load to continue to completion.

https://github.com/brimdata/zui/assets/5934157/2dcd8060-6279-40b2-84dc-f7440d854d89

jameskerr:
Good find. I just fixed it. Here's a video:

https://github.com/brimdata/zui/assets/3460638/6880616b-7f05-4f97-ad15-286365539c4e

philrz:
Cool, thanks @jameskerr. Indeed, I just did my own validation at commit 1098379 and I saw much the same as captured in your video.

philrz (approved):

PR #2991, opened 2024-01-26
Fix Waiting for Load

### LINKED ISSUE #2989: Preview & Load not dismissed when Load is pressed
## tl;dr

A bug with the Preview & Load screen appeared at commit a5ee8af such that a load operation is initiated when the **Load** button is pressed but the screen does not close until the load completes.

## Details

I confirmed via binary search that this bug first appeared at commit a5ee8af (#2955).

As shown in the attached video, when I tee up a large set of files in Preview & Load and press the **Load** button, the load does begin (you can see the tab appear in the background). But since I'm accustomed to the screen being dismissed and watching the filling load thermometer, I ended up hitting the button again multiple times thinking I maybe mis-clicked and it ends up initiating repeated additional loads of the same data to even more pools.

https://github.com/brimdata/zui/assets/5934157/36d84e71-f2dd-4573-88f2-675848fd6a99

jameskerr:
I returned the promise to make a unit test that doesn't exist now. I've added a note to that file that says not to return that promise.

fixes #2989

### DIFF
--- a/apps/zui/src/domain/loads/operations/create.ts
+++ b/apps/zui/src/domain/loads/operations/create.ts
@@ -15,7 +15,8 @@ export const submit = createOperation(
     const pool = await createPool(data)
     const script = new ZedScript(data.shaper || "")
     // Async so that we can return this and subscribe to updates on the load.
-    const promise = zui.pools
+    // Do not wait for the load to finish in this operation.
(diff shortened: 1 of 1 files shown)


philrz (approved):
Looks great. Thanks!

PR #2993, opened 2024-01-29
Fix Long Pool Names

### LINKED ISSUE #2911: The problem of long pool name
The problem of long pool name

This is not a special problem, but if the pool name is long, the **Load Data -  Query Pool**  options are no longer displayed

jameskerr:
Fix the UI issues with long pool names.

CleanShot 2024-01-29 at 09 50 52@2x

Closes #2911


philrz (approved):
Tests out nicely! 👍

PR #2995, opened 2024-01-30
Pin Playwright dependencies to 1.41.1

philrz:
While prepping the Zui `v1.6.0` release I spotted the note from #2934 that said:

> Update playwright to @next until 1.41 is released

It looks like Playwright is now up to a release [v1.41.1](https://github.com/microsoft/playwright/releases/tag/v1.41.1) so here I've pinned the dependencies. While all tests pass at this new combo, please double check that I'm doing this the right way as I'm basically just following a pattern. Thanks!

### DIFF
--- a/packages/zui-player/package.json
+++ b/packages/zui-player/package.json
@@ -7,9 +7,9 @@
     "ci": "playwright test -c ci.config.js"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


jameskerr (approved):
Yup, happy with this.

PR #3017, opened 2024-02-26
Give the export modal a max width

jameskerr:
As I was making the screen cast to share on twitter, I noticed the export modal changes it's width when you change export destinations between file and pool. This PR gives the export modal a consistent width.

CleanShot 2024-02-26 at 12 46 11@2x

### DIFF
--- a/apps/zui/src/views/export-modal/index.tsx
+++ b/apps/zui/src/views/export-modal/index.tsx
@@ -29,7 +29,7 @@ export function ExportModal() {
   useSubmitKey(ref)
 
   return (
-    <PopoverModal ref={popover.ref} className="max-width:fit">
+    <PopoverModal ref={popover.ref} style={{maxWidth: "53ch"}}>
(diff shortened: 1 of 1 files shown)


philrz (approved):
I tested it out and it's certainly an improvement over what was there before, so I've clicked Approve.

While I'm commenting though, I figured I'd point out that based on what I can see the width is maybe not _quite_ 100% consistent. Here's the baseline at current tip of `main` (commit 04c470a) showing the stark contrast in modal width:

https://github.com/brimdata/zui/assets/5934157/40df30e0-9118-4cd2-b732-0f544a711f37

And on this branch at commit 692c863, the width is mostly the same as I click between the two modes, but I can see a small/subtle change in the width.

https://github.com/brimdata/zui/assets/5934157/9c80edf2-a23f-44a9-a6e0-31a675ac7509

Figured I'd point it out while you're still under the hood in case you wanna fiddle some more. 😉

PR #3029, opened 2024-03-21
README adjustments

philrz:
In the time since #2818 merged, the top-level README for the repo has been incomplete/inaccurate. In this PR I propose moving the Zui app README detail to the top-level README and including a section that describes it all being a monorepo and what other packages are available. I've included a few other clean-ups along the way, which is mostly just adding/fixing hyperlinks.


jameskerr (approved):
Thank you!

PR #3032, opened 2024-03-27
Load Monaco Files From Node Modules Folder

### LINKED ISSUE #2915: Monaco editor doesn't load when airgapped

jameskerr:
Fixes #2915

The editor now loads without an internet connection. Instead of reaching out to a CDN, electron intercepts the request and serves it from node modules.

Here is the editor loaded.

CleanShot 2024-03-27 at 15 57 25@2x

And here is the sources tab revealing no network requests.

CleanShot 2024-03-27 at 15 57 17@2x

In order to get this working properly in development, I had to disable web security in the browser windows to allow for cross-origin javascript requests. This is only needed in development because we serve the html from a dev server. In production we serve the files from the same origin that we serve the monaco files, so they are on the same origin already.

I would love to test this in a release to make sure I stitched all the paths together correctly in a release environment.

### DIFF
--- a/apps/zui/package.json
+++ b/apps/zui/package.json
@@ -33,7 +33,8 @@
     "release-insiders": "electron-builder --config electron-builder-insiders.json --publish always"
   },
   "dependencies": {
-    "keytar": "^7.7.0"
+    "keytar": "^7.7.0",
+    "monaco-editor": "^0.47.0"
   },
   "devDependencies": {
     "@babel/core": "^7.17.9",
--- a/apps/zui/src/components/zed-editor.tsx
+++ b/apps/zui/src/components/zed-editor.tsx
@@ -59,7 +59,7 @@ export function ZedEditor(props: {
         minimap: {enabled: false},
         renderLineHighlight: "none",
         renderControlCharacters: false,
-        fontSize: "14px",
+        fontSize: 14,
         fontFamily: "var(--mono-font)",
         fontVariations: "inherit",
         lineNumbersMinChars: 4,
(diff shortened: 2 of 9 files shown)


philrz (approved):
I tested this out on Windows using an artifact from [this Actions run](https://github.com/brimdata/zui/actions/runs/8460120406) and the editor came up fine with my Internet disconnected. LGTM! 👍

jameskerr:
Nice!

Issue #3021, opened 2024-02-28
Error message from zq sometimes not surfaced


philrz:
## tl;dr

When Zui calls on `zq` in a way that's expected to fail (such as in **Preview & Load** when trying to read a file that is not of a format that Zed can read via auto-detect) the expected error message sometimes is not surfaced.

## Details

Repro is with Zui commit 57afe04.

The repro in the video below is taken on a Linux VM run in VirtualBox. The file that's being opened in Zui is the same [soccer-ball.png](https://github.com/brimdata/zui/blob/main/packages/sample-data/data/soccer-ball.png) that's opened by this e2e test which has been described as "flaky":

https://github.com/brimdata/zui/blob/57afe0458ff84182d4e392aada10f1d4608911da/packages/zui-player/tests/pool-loads.spec.ts#L42-L46

In the video I begin the process to load the file in **Preview & Load** several times. While the auto-detect failure is shown in the app most times, you can see on both the second and final attempts that no error message is shown.

https://github.com/brimdata/zui/assets/5934157/767269db-3530-4031-a6a9-2edbd12338f0

What led me to repro this manually is that I'd been studying that "flaky test" shown above via a [branch](https://github.com/brimdata/zui/compare/main...bad-data-error-msg) where I [ran just that test repeatedly in CI](https://github.com/brimdata/zui/actions/runs/8073911227) and found it failed 24 times out of 100 in a manner similar to what I showed in my manual repro video.

Based on what's shown here I suspect this may share a root cause with a separate Jest test that's been known to fail occasionally in CI that's been on my list to study:

https://github.com/brimdata/zui/blob/57afe0458ff84182d4e392aada10f1d4608911da/packages/zed-node/src/zq.test.ts#L106-L115

On the same Linux VM where I did the repro/video above I triggered a failure via this loop:

```
$ num=1 ; success=0; fail=0; while true; do yarn nx run zed-node:test --skip-nx-cache; if [ $? -eq 0 ]; then success=$(expr $success + 1); else fail=$(expr $fail + 1); fi ; num=$(expr $num + 1) ; echo "success=$success fail=$fail" ; done
```

To get to a quicker repro, I deleted all the other tests so just this one "zq with a bad zed" test would run, and I also first started a `cat /dev/random > /dev/null &` for every core on my VM since keeping the system under load seems to force the repro quicker (presumably because heavy load aggravates timing issues). With this approach I got 3 repros out of 218 runs, with this being an example of a failure:

```
[...]

PR #3033, opened 2024-03-28
Run pool load e2e tests in CI again

philrz:
The changes in #2972 made it such that our e2e tests now have a reliable way to check query results. In https://github.com/brimdata/zui/pull/2972#issuecomment-1971801505 I did some preliminary verification that this change was having the intended benefits, and that includes some baselining of how bad the problem was before and hence reminded us why we'd been skipping the "pool load" tests in CI. Now that #2972 is merged, in this PR I've banked that improvement by breaking out the e2e pool loads tests that are now reliable into their own set that we can run again in CI, while a separate test that remains unreliable is now broken out separately and is still skipped in CI (addressing that is tracked in #3021).

To verify, I did an Actions run on this branch with [these changes](https://github.com/brimdata/zui/commit/9a9f27dc160661895a7c1a1b2a4ff8a46c515d69) in place (since reverted) to run this set of `pool-load-success` tests in a loop on all platforms. As the results show, it ran successfully through all 1000 runs on both macOS and Windows without a single failure, and on Ubuntu it made it through 735 successful runs with no failures before Actions shot the job for running too long. So I think that shows things are pretty rock solid now!

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -55,18 +55,27 @@ jobs:
       - name: End to end tests
         id: playwright
         run: |
-          if [ "$RUNNER_OS" = "Linux" ]; then
-            # Mock for pcap slice opening to avoid hanging Actions Runner.
-            # See https://github.com/brimdata/zui/pull/2987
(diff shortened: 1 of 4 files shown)


jameskerr (approved):
Hell yeah! Thank you for your thorough testing!

PR #3044, opened 2024-04-09
Fixed the Settings Menu

jameskerr:
Just a few files change. 🤣

I redid the settings modal, and updated all form elements in the app. Hopefully for the last time.

And the bug about the brimcap file is fixed.

CleanShot 2024-04-09 at 16 07 50@2x

CleanShot 2024-04-09 at 16 08 16@2x


philrz:
I'm testing this branch at commit b025b3e using the Dev build from [this Actions run](https://github.com/brimdata/zui/actions/runs/8623622301).

I was testing on Windows just to make sure the `C:\`-style paths for the Brimcap YAML file worked out ok, and that checked out fine. However, while I was testing I happened to notice that the two words in **Brimcap Settings** got split across two lines, whereas on macOS they were in one line. It's not a crisis in any case, but since I assume you do most of your testing on macOS where you develop, I assume you were liking the look on a single and hence might want to make it consistent.

image

philrz:
I'm testing this branch at commit b025b3e using the Dev build from [this Actions run](https://github.com/brimdata/zui/actions/runs/8623622301).

I noticed that whenever I bring up the **Settings** the section for **Brimcap Settings** is always pre-selected by default.

image

It seems like if we wanted to be fancy we could remember which section the user was in last and start out with that pre-selected, since users will sometimes want to change something and see how it affects the app experience, then maybe quickly change it back. I'm not necessarily advocating for fanciness though, but at minimum I'd suggest we use a different selection as the default, specifically the **Application** one since it's at the top of the list, plus I imagine it's a setting users are more likely to twiddle than the Brimcap settings (despite the fact the breakage of the Brimcap setting is what motivated this PR 😄)

philrz (approved):

philrz:
@jameskerr and I had an offline discussion about my first impressions upon reviewing the new design in this PR. At first I was thrown by the lack of an explicit **Save** button like we used to have in the old **Settings** menu. The one thing this loses us is the ability to start making a change but then hit **Esc** or click **Cancel** to back out and leave the prior setting intact. However, after a quick review of some modern tools like Slack and VSCode, I could see that the approach used here where changes are effectively committed immediately is now more in vogue. I've quickly adjusted to the new paradigm and I'm optimistic our users will as well.

Issue #2949, opened 2023-12-26
Using "Brimcap YAML Config File" causes stack trace and breaks Settings


philrz:
Repro is with Zui commit 45bfd7c.

A community user emailed support@brimdata.io with this problem:

> I installed Zui a few days ago and wasn't aware that it was pre-configured with Suricata rules. I tried to set it up with a rules file, using the option in settings to load a YAML file. However, after loading the rules file, I encountered an error when opening the settings. I attempted to resolve the issue by reinstalling Zui, but the error persists. I also tried renaming the rules file to prevent it from loading, but the error persists despite these efforts.

Based on this description, it sounds like the file they were trying to load was not of the specific structure of what's described in the [docs](https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config) but rather a file of actual Suricata rules or some Suricata YAML config file that points at rules. As shown in the video below, if I drag such a file into the app, now each time I try to go into **Settings** I get a stack dump like the user reported:

```
Error: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.
    at HTMLInputElement.set [as value] (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:7721)
    at et (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:9350)
    at file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:99982
    at oS (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:103514)
    at ow (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:96624)
    at file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:96431
    at ok (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:96438)
    at os (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:93209)
    at oi (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:9:91779)
    at E (file:///_next/static/chunks/framework-d2a24e5002b8cad4.js:33:1373)
```

https://github.com/brimdata/zui/assets/5934157/f2f1ae20-c5b6-40a7-a25a-3c1f0a5b3d43

It seems we could stand to increase the robustness here such that if we don't see valid YAML or it's YAML but not of the expected structure we just surface an error saying it can't be used.

philrz:
[...]

PR #3049, opened 2024-04-13
Add Setting for pointing at local Suricata rules

### LINKED ISSUE #3047: Allow usage of local Suricata rules
When i try to add suricata rules or other yaml setting i have always this error, and then i need to reset to default setting because broke the setting window. precisely I would like to import the rules from https://github.com/AssoEchap/stalkerware-indicators

Screenshots
immagine_2024-04-11_152144692

Desktop Info

 - OS:  Windows
 - Zui Version: 1.7.0

philrz:
We've had a couple community issues recently where users mistakenly thought Zui's **Brimcap YAML Config File** setting may have been intended for loading additional local Suricata rules (#3047, #2949). We've never intended to expose UX in Zui to control all the possible customizations in bundled tools like Suricata and instead of tried to maintain a set of sensible defaults to which we've added modest enhancements over time when resources allow. When users ask for customizations that are more ambitious than we can easily offer, our catch-all/fallback answer is to refer them to the [Custom Brimcap Config](https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config) article that guides them through how they can point Zui at their own installs of pcap analyzers like Suricata that they've customized as much as they'd like.

All that said, I recognize that following the Custom Brimcap Config steps could be seen as fairly heavyweight by a user that wants to do something simple such as loading additional Suricata rules. Furthermore, for that specific use case, I already put in most of the hard work via https://github.com/brimdata/build-suricata/pull/63 to make sure that the Suricata rules updater that ships with Brimcap is capable of loading additional rule sets if they're specified on the command line. The user that spawned the inquiry that led to that Brimcap PR only seemed to be interested in standalone Brimcap, so at the time I did not pursue exposing the functionality at the Zui layer. But now that we've established that some Zui users are looking for this functionality, I figure our time is better spent delivering the enhancement rather than repeatedly pointing them at heavyweight docs and hoping for the best.

As described in the [`suricata-update` docs](https://suricata-update.readthedocs.io/en/latest/update.html#cmdoption-local), loading such additional rules is accomplished by appending `--local` and a pathname to a file or directory of rule files to include. This PR leverages this by exposing a new **Local Suricata Rules Folder** setting in Zui that allows the user to select a directory that may contain any number of Suricata rules files. The end result is that the additional rules will be included in the final set used by Zui alongside the default Emerging Threats Open rule set that's always present by default.

The following video taken on macOS shows the feature working as intended using a [Dev build](https://github.com/brimdata/zui/actions/runs/8743245710) based on commit c75e955 from this PR's branch. It uses the attached test data in [example.pcap.gz](https://github.com/brimdata/zui/files/14964797/example.pcap.gz) which is a capture of me pinging one of the IP addresses associated with the "stalkerware" rules requested in #3047.

First I baseline by loading the pcap with the default Emerging Threats Open rules and we see no Suricata alert is generated. Then I use the new setting to point at the folder on my desktop that contains the [additional stalkerware rules](https://github.com/AssoEchap/stalkerware-indicators/blob/master/generated/suricata.rules). Changing the setting triggers the re-run of the Suricata rules updater, and we can see the size of the assembled `suricata.rules` file increase as a result. I then re-import the pcap and we can see the expected alert now appears. Finally, I delete the setting such that the Suricata rules updater re-runs to bring me back to just the default Emerging Threats Open rule set again. Re-processing the pcap one last time, we see the alert is once again gone.

https://github.com/brimdata/zui/assets/5934157/2a72c57b-c028-4fd6-855f-a611eef931e2

I re-ran this same set of steps with the Windows build and saw the same successful outcome there.

If this merges, what I'll likely do right after is write a new article in the [Features](https://zui.brimdata.io/docs/features) section of the Zui docs that covers both this new feature and the existing **Brimcap YAML Config File** setting. Regarding the latter, up to now its "docs" link in **Settings** has pointed directly at the [Custom Brimcap Config](https://github.com/brimdata/brimcap/wiki/Custom-Brimcap-Config) article in the Brimcap wiki. That's been fine, but I feel like a standalone Zui doc that covers pcap processing (maybe reference the plugin system, etc.) in addition to these other topics would be more user-friendly.

Closes #3047

### DIFF
--- a/apps/zui/src/domain/configurations/plugin-api.ts
+++ b/apps/zui/src/domain/configurations/plugin-api.ts
@@ -1,12 +1,7 @@
 import ConfigPropValues from "src/js/state/ConfigPropValues"
 import {Store} from "src/js/state/types"
 
-export type ConfigItemType =
-  | "file"
-  | "string"
-  | "directory"
-  | "boolean"
-  | "char"
+export type ConfigItemType = "file" | "string" | "folder" | "boolean" | "char"
 
 export type ConfigItem = {
   name: string
--- a/apps/zui/src/domain/configurations/plugin-api.ts
+++ b/apps/zui/src/domain/configurations/plugin-api.ts
@@ -1,5 +1,6 @@
 import ConfigPropValues from "src/js/state/ConfigPropValues"
 import {Store} from "src/js/state/types"
+import {onStateChange} from "src/core/on-state-change"
 
 export type ConfigItemType = "file" | "string" | "folder" | "boolean" | "char"
 
@@ -50,5 +51,9 @@ export class ConfigurationsApi {
       if (this.get(config.name, prop) !== undefined) continue
       this.set(config.name, prop, config.properties[prop].defaultValue)
     }
+  }
+
+  watch(namespace: string, name: string, onChange: (val: any) => void) {
+    onStateChange(this.store, ConfigPropValues.get(namespace, name), onChange)
   }
 }
--- a/apps/zui/src/plugins/brimcap/config.ts
+++ b/apps/zui/src/plugins/brimcap/config.ts
@@ -1,2 +1,3 @@
 export const pluginNamespace = "brimcap"
 export const yamlConfigPropName = "yamlConfigPath"
+export const suricataLocalRulesPropName = "suricataLocalRulesPath"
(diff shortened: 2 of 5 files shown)


jameskerr (approved):
This turned out nice.

PR #3053, opened 2024-04-20
Add Setting for folder to store extracted pcap slices

### LINKED ISSUE #1423: Allow user to specify a directory for extracted pcap flows
Right now when the user clicks the Packets button, the flow extracted from the larger pcap gets stored in a file in a tmpdir before being opened in whatever app the OS has tied to that file type. In a group discussion on 2021-02-02, we recognized that it might be helpful to allow the user to specify a specific, non-tmpdir folder. A use case that might be enabled by this would be the saving of several separate extractions to the same folder, such as if data is being accumulated as part of an investigation. This might make sense as an option in the **Preferences** menu, or perhaps something that could be specified via right-click on the Packets button, possibly with memory of recent folder choices to make it easy for the user to keep adding to a particular location.

philrz:
As described in #1423, we've recognized that a user might find it useful to specify a specific directory to hold extracted pcap slices, e.g., if they're gathering them up as part of an investigation. The changes in #3049 made me realize how easy it would be to add this as a **Setting**, so that's what I've done here.

The attached video shows the feature working as intended using the branch from this PR at commit 651e403. As a baseline I show the default app behavior of extracting to `$TMPDIR`. Then I apply the new setting and show the pcap going to the selected directory. Then I clear out the setting and show the pcaps going back to `$TMPDIR` again.

https://github.com/brimdata/zui/assets/5934157/ea524209-183a-4231-8a0b-0e772ad1cd96

Closes #1423

### DIFF
--- a/apps/zui/src/domain/configurations/plugin-api.ts
+++ b/apps/zui/src/domain/configurations/plugin-api.ts
@@ -15,6 +15,7 @@ export type ConfigItem = {
   command?: string
   defaultValue?: string | boolean
   enum?: string[] | [string, string][]
+  placeholder?: string
 }
 
 export type Config = {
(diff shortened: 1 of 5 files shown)


jameskerr (approved):
Hell yeah!

Issue #3056, opened 2024-04-24
Zed lake service launched by Zui is accepting remote connections


philrz:
## tl;dr

In GA Zui `v1.5.0` and older, the Zed lake service was launched by Zui with `zed serve -l localhost:9867` such that client connections would only be permitted from `localhost`. The changes in #2934 resulted in this becoming `zed serve -l :9867` such that in GA Zui `v1.6.0` and newer Zui's local lake service has been open to connections from remote addresses as well. Hopefully default desktop configs are such that firewalls would prevent such incoming connections regardless, but we should likely revert this change in the interest of good security hygiene. Unfortunately, other tool changes (probably the same ones that resulted in the change in #2934) require us to do a bit more than just put back the original code.

## Details

The change from #2934 in question is commit https://github.com/brimdata/zui/pull/2934/commits/e0e3a7d85ebb45e3150a1274463af963a4775dda where this line that formerly said `this.addr(),` changed to:

https://github.com/brimdata/zui/blob/9c98969bb0ce7abc3171d4e26058feeb56db64cc/packages/zed-node/src/lake.ts#L41

If we revert just that change back to `this.addr()` and then try to do something with the lake such as load data, we now get an error:

![image](https://github.com/brimdata/zui/assets/5934157/8eb5016c-3188-4549-a3d5-5cf01f589697)

Web searches regarding this error brings up Node.js issue https://github.com/nodejs/node/issues/40702 wherein [this comment](https://github.com/nodejs/node/issues/40702#issuecomment-1103623246) seems like the best summary I've seen. In brief, it appears that the move to Node v18 in #2972 has made it such that now when the `zed-node` client seeks to connect to `localhost`, new Node behavior has made it such that the connection is now attempted to the IPv6 address `::1` and since the Zed service is listening only on the IPv4 address the connection is refused.

I've been experimenting with some of the fixes/workarounds described in https://github.com/nodejs/node/issues/40702 and am weighing the trade-offs. I'll add a comment next with findings thus far.

philrz:
Between my research and discussions with the team thus far, there's a few ways I see to potentially address this.

1. The Zed lake service could be enhanced to allow for listening for incoming connections on both IPv4 `127.0.0.1` and IPv6 `::1`. However, the Zed Dev team is currently tied up with other priorities, so consensus was to shelve this approach for now.

[...]

PR #3061, opened 2024-05-01
Native Node Fetch

jameskerr:
Update electron dependencies.
Drop node-fetch 🙌
Use node native fetch
It required I add the duplex: "half" option when sending a post body.


philrz (approved):
@jameskerr did some Q&A with me on a brief Zoom and I'm now 👍 on these changes. I've also confirmed that once these changes are in place the fix to #3056 should be a one-liner. In the interest of addressing one thing at a time I'll put up a PR for that once this merges.

PR #3062, opened 2024-05-02
Update Zed with Ast Changes

jameskerr:
- Update Zed
- Adjust Ast Parsing Code

### DIFF
--- a/apps/zui/src/domain/editor/parse.test.ts
+++ b/apps/zui/src/domain/editor/parse.test.ts
@@ -3,12 +3,19 @@
  */
 import {SystemTest} from "src/test/system"
 import {parse} from "./operations"
+import {ZedAst} from "src/app/core/models/zed-ast"
 
 new SystemTest("editor.parse op")
 
 test("editor.parse", async () => {
   const result = await parse("from source | count() by name")
-  expect(result.map((o) => o.kind)).toEqual(["From", "Summarize"])
+  expect(result.map((o) => o.kind)).toEqual(["Pool", "Summarize"])
+})
+
+test("Zed Ast Pool Name", async () => {
+  const result = await parse("from source | count() by name")
+  const ast = new ZedAst(result, null)
+  expect(ast.poolName).toBe("source")
 })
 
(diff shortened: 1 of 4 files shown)


apps/zui/src/domain/editor/parse.test.ts
```diff
@@ -3,12 +3,19 @@
...
 new SystemTest("editor.parse op")
 
 test("editor.parse", async () => {
   const result = await parse("from source | count() by name")
-  expect(result.map((o) => o.kind)).toEqual(["From", "Summarize"])
+  expect(result.map((o) => o.kind)).toEqual(["Pool", "Summarize"])
+})
+
+test("Zed Ast Pool Name", async () => {
+  const result = await parse("from source | count() by name")
```
mattnibs, line 16: Another good test here would be to test that it successfully gets the name from a multi-from:
```
from (
  pool test1
  pool test2
)```

File: apps/zui/src/domain/editor/parse.test.ts [hunk 2]
```diff
@@ -3,12 +3,25 @@
...
+  const ast = new ZedAst(result, null)
+  expect(ast.poolName).toBe("source")
+})
+
+test("Zed Ast Multi Pool", async () => {
+  const result = await parse("from (pool test1 pool test2) | count() by name")
+  const ast = new ZedAst(result, null)
+  expect(ast.poolNames).toEqual(["test1", "test2"])
 })
```
mattnibs, line 25: I'm confused if you have a multiple `from` query and you call poolName you'll get null. Conversely if you call poolNames and you have a single `from` statement you will get null. Shouldn't you just have one method not both?

jameskerr: Many parts of the app are designed around a single pool name. Some features get activated when there is a single pool name. I'm not yet what these features should do if there are multiple pool names in the query.

There is no code in the app that needs to know all of the pool names, I just added that method to test the case you suggested.

mattnibs (approved):

Issue #3063, opened 2024-05-02
e2e pool-load-success test hanging on close


philrz:
Repro is with Zui commit ca047c0.

When working with Zui since the merge of #3061, I noticed one of the e2e tests failing that hadn't been before. How the failure looks:

```
  ✘  2 pool-load-success.spec.ts:27:7 › Pool Loads (successes) › load more data into the pool (2.5s)
  1) pool-load-success.spec.ts:27:7 › Pool Loads (successes) › load more data into the pool ────────
    "afterAll" hook timeout of 30000ms exceeded.
      10 |   });
      11 |
    > 12 |   test.afterAll(async () => {
         |        ^
      13 |     await app.shutdown();
      14 |   });
      15 |
        at /Users/phil/work/zui/packages/zui-player/tests/pool-load-success.spec.ts:12:8
  1 failed
    pool-load-success.spec.ts:27:7 › Pool Loads (successes) › load more data into the pool ─────────
```

I just did a controlled set of repros and it does indeed seem like the failures are correlated with those changes. I used these repro steps:

```
$ SUCCESSES=0; FAILURES=0; while true; do   yarn e2e -g pool-load-success.spec.ts --skip-nx-cache; if [ $? -eq 0 ] ; then SUCCESSES=$(expr $SUCCESSES + 1); else FAILURES=$(expr $FAILURES + 1); fi; echo "Successful test runs=$SUCCESSES, failed test runs=$FAILURES"; done
```

Running at commit 59c4c4a that was right before the changes in #3061 produced 53 successful runs and 0 failures before I stopped it. At commit ca047c0 for the merge of #3061 it produced 46 successful runs and 10 failures.

philrz:
In an attempt to divide & conquer, I made a branch [zui-3063-electron-only](https://github.com/brimdata/zui/compare/main...zui-3063-electron-only) which was started at Zui commit 59c4c4a right before the changes in #3061 merged to `main`. In that branch I made _just_ the changes from #3061 to advance Electron to `30.0.1` and `.node_version` to `20.11.0` but did _not_ make any of the changes to use Node native Fetch in lieu of `node-fetch`. Using the same repro loop described above, with this branch I saw 120 successful runs of `pool-load-success.spec.ts` and 0 failures before I stopped it. That seems to indicate that the problems captured in this issue are somehow specific to the Node native Fetch and/or how we use it.

philrz:
[...]

PR #3066, opened 2024-05-03
Make console logs more accessible during Playwright e2e test runs

philrz:
While starting to debug #3063, I tried adding some `console.log()` debug statements to `packages/zed-node/src/lake.ts`, but the messages weren't visible in the shell window from which I ran `yarn e2e` nor in the `packages/zui-player/run/playwright-itest/Pool Loads (succeeses)/1/logs/main.log`. @jameskerr hacked at this with me a bit over Zoom and came up with these two improvements that make the messages show up in both locations. I'll put comments inline that explain which does what.

### DIFF
--- a/apps/zui/src/electron/run-main/before-boot.ts
+++ b/apps/zui/src/electron/run-main/before-boot.ts
@@ -5,12 +5,16 @@ import {windowsPre25Exists} from "../windows-pre-25"
 import {MainArgs} from "./args"
 import {setLogLevel} from "../set-log-level"
 import {runMigrations} from "./run-migrations"
+import log from "electron-log"
 
 app.disableHardwareAcceleration()
 
 export async function beforeBoot(
   args: Partial<MainArgs>
 ): Promise<string | null> {
+  // Ensure all console logs go through electron-log
+  console.log = log.log
+
   // Disable security warnings
   // process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = "true"
   // Setup app paths, this must be first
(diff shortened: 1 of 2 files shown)


apps/zui/src/electron/run-main/before-boot.ts
```diff
@@ -5,12 +5,16 @@ import {windowsPre25Exists} from "../windows-pre-25"
...
+import log from "electron-log"
 
 app.disableHardwareAcceleration()
 
 export async function beforeBoot(
   args: Partial<MainArgs>
 ): Promise<string | null> {
+  // Ensure all console logs go through electron-log
+  console.log = log.log
+
```
philrz, line 17: This one caused the `console.log()` messages I'd added in `packages/zed-node/src/lake.ts` to show in the `packages/zui-player/run/playwright-itest/Pool Loads (succeeses)/1/logs/main.log`.

packages/zui-player/helpers/test-app.ts
```diff
@@ -45,6 +45,11 @@ export default class TestApp {
     if (bin) launchOpts.executablePath = bin;
     this.zui = await electron.launch(launchOpts);
 
+    // Pipe the stdout from the electron process into the test runner process
+    if (process.env['VERBOSE']) {
+      this.zui.process().stdout.pipe(process.stdout);
+    }
+
```
philrz, line 52: For this one, if I run my tests via `VERBOSE=1 yarn e2e`, this change caused the `console.log()` messages I'd added in `packages/zed-node/src/lake.ts` to also show up in the shell window from which I ran `yarn e2e`.

jameskerr (approved):

Issue #2748, opened 2023-04-10
Infinite scroll of aggregation results


philrz:
At the time of the filing of this issue, Zui is at commit c43dac0.

The changes in #2355 provided infinite scrolling for regular search results, but if an aggregation returned a large number of results the user is shown a message in the bottom status bar "Results: Limited to first 2000". No additional results appear even if they scroll to the bottom, hence they'd need to rework the query themselves if they want to see other results.

Once we have the pagination help from the backend tracked in https://github.com/brimdata/zed/issues/4491, Zui should be able to offer the infinite scroll even on these aggregation results.

PR #3076, opened 2024-05-21
Use zed query/describe endpoint

mattnibs:
This commit uses the Zed service query/describe endpoint to describe to contents of a given Zed query- replacing the use of the compile endpoint and the Zed-ast library for this purpose.


philrz:
I'm currently testing this branch at commit 6088bc5.

I noticed something shown in the attached video: Large aggregation results now allow scrolling past the first 500 rows.

https://github.com/brimdata/zui/assets/5934157/d340e3ed-d171-4a0a-9283-255594c136aa

By comparison, at current tip of `main` (e.g., with Zui Insiders [1.7.1-20](https://github.com/brimdata/zui-insiders/releases/tag/v1.7.1-20) as shown in this next video) it still shows "Limited to 500 rows" and does not load additional data when scrolling to the bottom.

https://github.com/brimdata/zui/assets/5934157/cd353c43-1602-4bd9-a3d1-3b43a7c9b05d

I'm just flagging this difference because my recollection is that we've intentionally held off on allowing infinite scrolling with aggregations because they have the potential to be very expensive to run (#2748).

philrz:
This is effectively a note-to-self but I'll record it here in case anyone else has the same question upon reviewing these changes (cc: @jameskerr). In the ztest in https://github.com/brimdata/zed/pull/5126 I noticed that the `head` in the API call to `/describe` was being populated with the pool source info, but sniffing my API traffic in the Zui branch from this PR showed the `from` pool info going into `query` while `head` was `null`. I checked with @mattnibs and he confirmed that at this point the `head` is something primarily used to make testing easier on the Zed side and so while Zui would have the option to populate it, it's also fine/expected that for now Zui is not populating `head`.

apps/zui/src/js/models/program.test.ts
```diff
@@ -57,21 +58,22 @@ describe("drill down", () => {
     count: 24,
   })
 
-  async function run(value: any, text: string) {
-    const info = await fetchQueryInfo(text)
-    return drillDown(text, value, info.isSummarized, info.groupByKeys)
+  async function run(value: any, text: string, pool?: string) {
+    const d = await fetchQueryInfo(text, pool)
```
nwt, line 62: Nits: Every call to this passes `"test"` as `pool` so no need for a parameter. And `d` is a surprising name for the result of `fetchQueryInfo()`.
```suggestion
  async function run(value: any, text: string) {
    const info = await fetchQueryInfo(text, "test")
```

apps/zui/src/views/session-page/handler.ts
```diff
@@ -66,16 +66,18 @@ export class SessionPageHandler extends ViewHandler {
...
+      if (info.error) {
+        return
+      }
       this.dispatch(QueryInfo.set({isParsed: true, ...info}))
+      const poolName = this.select(QueryInfo.getPoolName)
       this.invoke("updatePluginSessionOp", {poolName, program})
+      const pool = this.select(Pools.getByName(lakeId, poolName))
       if (pool && !pool.hasSpan()) {
         this.dispatch(syncPool(pool.id, lakeId))
       }
```
nwt, line 78: Previously, all this stuff would happen regardless of the value of `info.error`. Why change that?

apps/zui/src/views/results-pane/run-results-query.tsx
```diff
@@ -15,7 +15,7 @@ export const runResultsMain = createHandler(
 
     // See if we can paginate this query
     await waitForSelector(QueryInfo.getIsParsed).toReturn(true)
-    const canPaginate = !select(QueryInfo.getIsSummarized)
+    const canPaginate = select(QueryInfo.hasAggregation)
```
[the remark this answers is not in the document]
philrz: @nwt: I might be able to help field this one. As described in #2748 and https://github.com/brimdata/zed/issues/4491, it's my understanding that Zui's always taken the conservative stance of not trying to do pagination on aggregations because they have the potential to be expensive and hence considered a bad idea to keep re-executing the whole thing just to get single pages of results. By contrast with straight search results the backend has benefit of the the pool key and other potential optimizations, hence the app can ask for page-sized subsets via `head`/`tail` without putting excess burden on the backend.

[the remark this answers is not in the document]
nwt: @philrz: As I understand it, this pagination decision is about correctness rather than performance. The `summarize` operator produces output with nondeterministic order, so you can't paginate that output correctly by re-executing unless you sort it before applying `head` and `tail`.

In fact, now that I've thought about it, I don't think we can paginate `summarize` output correctly unless it's sorted by all of the `summarize` keys.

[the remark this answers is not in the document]
philrz: @nwt: Yes, that all makes sense. But I think both can be true. Without digging up years of history to confirm, I think I remember a time when our aggregations did return results in a deterministic order, and so the decision to not try to paginate aggregation results was at one time based on the performance argument as recorded in those issues (I didn't make that stuff up... I recall it being the outcome of chats with others.) So even if we put in some kind of logic to confirm if the results were sorted, the perf argument might remain.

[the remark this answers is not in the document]
mattnibs: That's what I thought but @philrz noticed that the behavior changed when this was based off of isSorted. I think basing pagination on whether or not the query is sorted is the right way to go but this change should be done on a separate pr and we should try to keep things as close to the way they were in this pr.

apps/zui/src/views/session-page/handler.ts
nwt: Well, that certainly seems like the sort of thing I should have been able to figure out for myself. Maybe next time!

apps/zui/src/views/results-pane/run-results-query.tsx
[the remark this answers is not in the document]
nwt: @mattnibs: That works for me.

philrz, line 18:
```suggestion
    const canPaginate = !select(QueryInfo.hasAggregation)
```

Tabling the broader discussion for the moment, I think the loss of the `!` was a mistake here. I just ran a quick test with this branch as of commit abc397c and as this video shows that had us losing paging for search results but still having it for aggregations. Restoring the `!` seems to bring back what we had before.

https://github.com/brimdata/zui/assets/5934157/930f16d5-0c22-4294-91bf-ce56e02b2644

nwt (approved):
I think this looks good!

apps/zui/src/views/session-page/handler.ts
```diff
@@ -66,16 +66,18 @@ export class SessionPageHandler extends ViewHandler {
     const history = this.select(Current.getHistory)
 
     fetchQueryInfo(program).then((info) => {
-      const {poolName, error} = info
-      const pool = this.select(Pools.getByName(lakeId, poolName))
-
+      if (info.error) {
```
jameskerr, line 69: Fix this.

apps/zui/src/views/session-page/handler.ts
```diff
@@ -66,16 +66,18 @@ export class SessionPageHandler extends ViewHandler {
...
       this.dispatch(QueryInfo.set({isParsed: true, ...info}))
+      const poolName = this.select(QueryInfo.getPoolName)
       this.invoke("updatePluginSessionOp", {poolName, program})
+      const pool = this.select(Pools.getByName(lakeId, poolName))
       if (pool && !pool.hasSpan()) {
         this.dispatch(syncPool(pool.id, lakeId))
       }
 
-      if (!error && history.action === "PUSH") {
+      if (history.action === "PUSH") {
```
jameskerr, line 80: Fix this too

packages/zed-js/src/client/base-client.ts
```diff
@@ -51,15 +51,17 @@ export abstract class BaseClient {
     return new ResultStream(result, abortCtl);
   }
 
-  async compile(
+  async describeQuery(
     query: string,
+    pool?: string,
```
jameskerr, line 56: Move to options

PR #3077, opened 2024-05-22
Apple Silicon macOS test/builds

### LINKED ISSUE #1266: Build/test for Mac M1/Silicon models
Brim has traditionally supported macOS, but most of that time has been in the era of Intel-based Macs. Apple has more recently introduced their Macs based on the [M1](https://www.apple.com/mac/m1/) chip. Smoke tests have confirmed that the Mac releases we currently ship seem to run fine on them. However, when conditions allow we'd surely like to improve our coverage here. Two things come to mind:

1. GitHub Actions will need to provide runners based on M1 Macs (tracked by [virtual-environments/2187](https://github.com/actions/virtual-environments/issues/2187)) so we can run our CI on them and confirm that we don't have M1-specific bugs
2. We'd want to start building macOS artifacts that include the [Electron with M1 support](https://www.electronjs.org/blog/apple-silicon) to deliver the best possible performance on these platforms

philrz:
## tl;dr

Now that GitHub offers free Actions runners based on Apple Silicon, it becomes feasible for us to run CI and create builds for this platform. While getting this working, I managed to simplify the build process a little (was able to drop the separate `electron-builder-notarize` dependency) but also had to add some stuff to work around rough edges in the tooling. Details are below and in-line PR comments.

## Actions Runners

Per [GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), the Runner `macos-14` is on Apple Silicon while the `macos-12` we've been using continues to run on Intel. While I don't anticipate platform-specific issues, I've gone ahead here and added `macos-14` to the OS matrix for our CI so we can continue to test on the same platforms we build for.

## Dropping `electron-builder-notarize`

Early in my research I could tell that there were still some lingering growing pains in the tooling related to Apple Silicon. Rather than burn time bumping into what might be known problems that were already fixed, my first step was therefore to advance to the newest release of `electron-builder`. It's at this point I was reminded that `electron-builder` can now do notarization on its own as an alternative to relying on the separate `electron-builder-notarize` tool as we've done in the past. Indeed, I think I recall @jameskerr having once spotted this and we'd made a mental note to look into it, but I used "if it ain't broke" logic to justify putting it off. In this case I knew I was likely to break and re-fix our build system anyway, so I figured I'd bite the bullet and maybe spare myself from having to work around problems in yet another tool.

This admittedly did send me briefly down a rabbit hole when I saw in the [`electron-builder` Mac docs](https://www.electron.build/configuration/mac) that they recommended a different set of credentials (based on API keys) over the password-based approach we'd been using with `electron-builder-notarize` (explanation in https://github.com/electron-userland/electron-builder/issues/7859). I spent a little time looking at what it would take to start using this alternate approach but ended up backing off because API keys seemed linked to "App Store Connect", hence the assumption we were building something to be published into the App Store, so first steps had Apple asking me to sign new agreements, entering details on payments/taxes related to all the revenue we'd surely make from our App Store sales, etc. Maybe it would have been possible to take these steps only as far as necessary to just get the API keys and not actually publish anything to the App Store. However, just as this was making me nervous, I found that I could extend the password-based approach we've been using all this time in `electron-builder` by making two small changes:

1. Changed one environment variable (`APPLE_ID_PASSWORD` becomes `APPLE_APP_SPECIFIC_PASSWORD`).
2. Added a `teamId` value in the `electron-builder.json` (which is redundant with the `APPLE_TEAM_ID` env var, but needed to work around open issue https://github.com/electron-userland/electron-builder/issues/8103)

## Entitlements, App Sandbox, and Hardened Runtime

My first attempts built successfully but then when @mattnibs would try to run them they'd fail with a crash like:

```
$ /Applications/Zui.app/Contents/MacOS/Zui 

#
# Fatal process out of memory: Failed to reserve virtual memory for CodeRange
#
----- Native stack trace -----

1: 0x1151c002c node::OnFatalError(char const*, char const*) [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
2: 0x1129b3248 _$LT$object..xcoff..Symbol64$u20$as$u20$object..read..xcoff..symbol..Symbol$GT$::n_value::he0fe3918a79f4509 [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
3: 0x10f628db4 _$LT$font_types..tag..Tag$u20$as$u20$core..cmp..PartialEq$LT$$u5b$u8$u3b$$u20$4$u5d$$GT$$GT$::eq::h1cf26754fa042eb7 [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
4: 0x10f628cf8 _$LT$font_types..tag..Tag$u20$as$u20$core..cmp..PartialEq$LT$$u5b$u8$u3b$$u20$4$u5d$$GT$$GT$::eq::h1cf26754fa042eb7 [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
5: 0x10f77f7a0 v8::Unlocker::~Unlocker() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
6: 0x1129b7898 _$LT$object..xcoff..Symbol64$u20$as$u20$object..read..xcoff..symbol..Symbol$GT$::n_value::he0fe3918a79f4509 [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
7: 0x10f77f694 v8::Unlocker::~Unlocker() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
8: 0x10f7f198c v8::CppHeap::wrapper_descriptor() const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
9: 0x10f757b48 v8::Unwinder::PCIsInV8(unsigned long, v8::MemoryRange const*, void*) [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
10: 0x10f7586b0 v8::Unwinder::PCIsInV8(unsigned long, v8::MemoryRange const*, void*) [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
11: 0x10fbea454 v8::internal::TickSample::print() const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
12: 0x10f64cd6c v8::Isolate::Initialize(v8::Isolate*, v8::Isolate::CreateParams const&) [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
13: 0x112b53d70 _$LT$object..xcoff..Symbol64$u20$as$u20$object..read..xcoff..symbol..Symbol$GT$::n_value::he0fe3918a79f4509 [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
14: 0x10e52ea84 v8::Signature::New(v8::Isolate*, v8::Local<v8::FunctionTemplate>) [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
15: 0x10e5158d8 v8::BackingStore::MaxByteLength() const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
16: 0x1108c9dac v8::internal::compiler::CompilationDependencies::FieldTypeDependencyOffTheRecord(v8::internal::compiler::MapRef, v8::internal::compiler::MapRef, v8::internal::InternalIndex, v8::internal::compiler::ObjectRef) const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
17: 0x1108cd700 v8::internal::compiler::CompilationDependencies::FieldTypeDependencyOffTheRecord(v8::internal::compiler::MapRef, v8::internal::compiler::MapRef, v8::internal::InternalIndex, v8::internal::compiler::ObjectRef) const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
18: 0x1108c9680 v8::internal::compiler::CompilationDependencies::FieldTypeDependencyOffTheRecord(v8::internal::compiler::MapRef, v8::internal::compiler::MapRef, v8::internal::InternalIndex, v8::internal::compiler::ObjectRef) const [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
19: 0x10e751a18 v8::CodeEvent::GetScriptLine() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
20: 0x10e752a78 v8::CodeEvent::GetScriptLine() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
21: 0x10e7528f0 v8::CodeEvent::GetScriptLine() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
22: 0x10e751230 v8::CodeEvent::GetScriptLine() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
23: 0x10e7513f4 v8::CodeEvent::GetScriptLine() [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
24: 0x10e42c51c ElectronMain [/Applications/Zui.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework]
25: 0x1828020e0 start [/usr/lib/dyld]
Trace/BPT trap: 5
```

After doings some web searches to see if others were seeing a similar crash, I eventually found my way back to the [`electron-builder` Mac docs](https://www.electron.build/configuration/mac) that include this passing comment:

> Be aware that your app may crash if the right entitlements are not set like `com.apple.security.cs.allow-jit` for example on arm64 builds with Electron 20+.

Sure enough, once I learned how to make that change, the Apple Silicon builds no longer crashed. However, since the concept of "entitlements" was unfamiliar to me, I felt the need to go on a detour to learn a bit about them to understand what they've been set for on our Intel builds all this time, if we should be setting others, etc. The docs were a bit confusing in spots since most seem to be written as if to an audience that have already committed their whole lives to developing in the Apple universe and hence provide limited context. But I'll summarize what I came to understand, with the caveat that someone "more developer than me" could surely explain it better and/or find mistakes in my understanding.

[This post](https://apple.stackexchange.com/questions/52675/how-do-i-find-out-what-entitlements-an-app-has) taught me how to check an app's entitlements. For example, here's what our latest GA Zui Intel build [v1.7.0](https://github.com/brimdata/zui/releases/tag/v1.7.0) shows:

```
$ codesign -d --entitlements - --xml /Applications/Zui.app
Executable=/Applications/Zui.app/Contents/MacOS/Zui
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd"><plist version="1.0"><dict><key>com.apple.security.cs.allow-unsigned-executable-memory</key><true/><key>com.apple.security.cs.disable-library-validation</key><true/></dict></plist>
```

So, picking that apart, we see [`com.apple.security.cs.allow-unsigned-executable-memory`](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-unsigned-executable-memory) and [`com.apple.security.cs.disable-library-validation`](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-library-validation). We never took explicit steps to _ask_ that these be included in our builds, but I found [this file in `electron-builder`](https://github.com/electron-userland/electron-builder/blob/77f977435c99247d5db395895618b150f5006e8f/packages/app-builder-lib/templates/entitlements.mac.plist) that references them, so that seems to explain why they're there by default. Strangely, that also shows an entry for `com.apple.security.cs.allow-jit` and it's been in there since January, 2020 (https://github.com/electron-userland/electron-builder/pull/4491) including in the `electron-builder` version I started using in this PR, so I'm still not 100% sure why I had to ask for it explicitly. But now that I was aware that the Apple docs effectively discourage the use of those two entitlements, I became more interested in knowing if we could live without them going forward.

I did ultimately find evidence that we can leave them behind.

1. [This comment](https://github.com/electron/electron-notarize#prerequisites) linked from that `plist` file explains that the `com.apple.security.cs.allow-unsigned-executable-memory` entitlement is only needed for Electron 11, and that its use in Electron 12+ is discouraged due to the same security exposure cited by Apple.
2. [This issue](https://github.com/electron-userland/electron-builder/issues/3940) also linked from that `plist` file ultimately concludes with [this comment](https://github.com/electron-userland/electron-builder/issues/3940#issuecomment-1421566393) that happens to be a fix I already put in Zui as part of #2917, indicating we can drop this entitlement as well.

Finally, using the config in this PR that dropped those two entitlements and left us with only the JIT one, I did smoke testing with an Intel build doing all the typical user stuff (loading pcaps, opening slices in Wireshark, opening `whois`, etc.) and it all went fine.

It also should be noted that most mentions of entitlements are in the context of discussing [App Sandbox](https://developer.apple.com/documentation/security/app_sandbox/), so I had to learn a bit about that as well. Consider the entitlements shown for a common app like Slack:

```
$ codesign -d --entitlements - --xml /Applications/Slack.app/Executable=/Applications/Slack.app/Contents/MacOS/Slack
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd"><plist version="1.0"><dict><key>com.apple.security.app-sandbox</key><true/><key>com.apple.security.application-groups</key><array><string>BQR82RBBHL.com.tinyspeck.slackmacgap</string><string>BQR82RBBHL.slack</string></array><key>com.apple.security.cs.allow-jit</key><true/><key>com.apple.security.device.audio-input</key><true/><key>com.apple.security.device.camera</key><true/><key>com.apple.security.device.microphone</key><true/><key>com.apple.security.device.usb</key><true/><key>com.apple.security.files.bookmarks.app-scope</key><true/><key>com.apple.security.files.downloads.read-write</key><true/><key>com.apple.security.files.user-selected.read-write</key><true/><key>com.apple.security.network.client</key><true/><key>com.apple.security.network.server</key><true/><key>com.apple.security.print</key><true/></dict></plist>
```

You'll notice entitlements in here like `com.apple.security.network.client`. Zui has acted as a "network client" just fine all these years but we don't list this entitlement, so what's up? Well, what I've learned is how Apple offers this "App Sandbox" wrapper as a way to run apps more safely, and one requirement of this is that the app has to list _all_ of its entitlements for the things it might potentially do, hence things like hitting the network, accessing your microphone, etc. To be distributed through the App Store, an app _must_ run in Sandbox, so that explains why we saw what we did with Slack. Likewise, since Zui isn't distributed through App Store, we've never had to.

If you're curious, you can add a column to Activity Monitor that'll show which apps are running in the Sandbox. While many are (e.g., Slack), many I use every day are not (e.g., Terminal, Google Chrome, Camtasia), so by no means is it a black eye that we're non-Sandbox.

image

To tie this all back to Entitlements, I ultimately recognized that dealing with them is _necessary_ for Sandbox and being in the App Store, hence why I saw these topics mentioned together so often. However, there are also non-Sandbox/non-App-Store contexts when Entitlements manner, such as this JIT setting. But now we know a bit more should the day come that we want to start jumping through hoops to publish Zui in the App Store.

Yet another topic I encountered in my travels was [Hardened Runtime](https://developer.apple.com/documentation/security/hardened_runtime), since that mentions how it disables JIT compilation and why the entitlement is therefore needed in our case (see also: [Porting just-in-time compilers to Apple silicon](https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon) for why it's needed with Apple Silicon but wasn't with Intel). An [EntitlementCheck tool](https://github.com/cedowens/EntitlementCheck) was ultimately useful to confirm that we've been using Hardened Runtime on our Zui builds all along, so once I confirmed we were in line with best practices there and the JIT entitlement got the Apple Silicon builds working, I was satisfied. Below is with the last GA Zui release [v1.7.0](https://github.com/brimdata/zui/releases/tag/v1.7.0):

```
$ ./Hardened_Runtime_Check.py | grep -i zui
[+] Hardened runtime enabled in: /Applications/Zui.app/Contents/MacOS/Zui
```

Per its name, that same repo has a tool that flags undesirable entitlements, and indeed this also flagged the two that we'd had in the past and won't anymore with the new builds that use only the JIT entitlement:

```
$ ./Entitlements_Check.py | grep -i zui
[-] Binary can load arbitrary unsigned plugins/frameworks (has com.apple.security.cs.disable-library-validation entitlement): /Applications/Zui.app/Contents/MacOS/Zui
[-] Binary allows c code patching, NSCreateObjectFileImageFromMemory, or dvdplayback framework (has com.apple.security.cs.allow-unsigned-executable-memory entitlement): /Applications/Zui.app/Contents/MacOS/Zui
```

# Merging `latest-mac.yml`

While the changes above got the Apple Silicon builds working, a problem remained with the auto-updates. The tl;dr is the update process relies on the file `latest-mac.yml` that's generated during the build process. Here's an example of the one from the last GA Zui release [v1.7.0](https://github.com/brimdata/zui/releases/tag/v1.7.0):

```
version: 1.7.0
files:
  - url: Zui-1.7.0-mac.zip
    sha512: 4IO2UaJrfX97593l9oVdq3IyuPoouFyloN8on51HFTZQlRoZF62UKrSQmAj0zzDBVjD4botyw4bInXJsJKOrJw==
    size: 251289997
  - url: Zui-1.7.0.dmg
    sha512: N0csX2QLLWBV9I9Hd0IgVWoeLYb0h63wjq5reVoVE114QzYMmhJwn0imN5mnqMoukXgxywi6ZpVg9dVZif/tZg==
    size: 259715184
path: Zui-1.7.0-mac.zip
sha512: 4IO2UaJrfX97593l9oVdq3IyuPoouFyloN8on51HFTZQlRoZF62UKrSQmAj0zzDBVjD4botyw4bInXJsJKOrJw==
releaseDate: '2024-04-03T15:52:00.673Z'
```

An Apple Silicon build would create a similar file but with entries for `Zui-1.7.4-arm64.dmg` and `Zui-1.7.4-arm64-mac.zip`. When running separate Intel and Apple Silicon builds converged on the same GitHub Releases destination like I'm doing in this PR, what ends up happening is that the build that finishes last pushes its `latest-mac.yml` to GitHub, overwriting the one generated by the first build.

As mentioned in a comment in the `merge-mac-release-files.mjs` script added in this PR, I found an open `electron-builder` issue where other users were struggling with the same problem and so I borrowed the code from [this comment](https://github.com/electron-userland/electron-builder/issues/5592#issuecomment-2004803764) to create a workaround script to do the same for our build process. In brief, the builds for each macOS platform push their own separate `latest` files temporarily to GitHub, then the build that finishes second downloads both files, merges them to a single `latest-mac.yml` that mentions both the Intel and Apple Silicon offerings, then pushes that file back up to GitHub and removes the temporary files. It's a little hacky, but it works.

## Upgrade Path

@mattnibs helped test all this using releases in personal forks I created of Zui and Zui Insiders where I could publish real releases without risk of impacting our existing user base. From what we found, here's what I expect we'll need to communicate to macOS users.

1. Having been running Intel builds all this time, if they continue to rely only on auto-update, they'll continue to be updated to Intel builds even as the Apple Silicon ones become available and referenced in the `latest-mac.yml`.

2. When a user wants to switch to Apple Silicon, they should do a one-time manual download/install of a `Zui-*-arm64.dmg` build. During installation, they can accept the prompt to **Replace** the existing (Intel-based) Zui. All their settings and Zed lake data will remain intact.

3. As newer releases come out, the users that did the one-time manual install of an Apple Silicon build will auto-update to the newer Apple Silicon builds.
 
---

Closes #1266


.github/actions/build-zui/action.yml
```diff
@@ -69,7 +69,7 @@ runs:
       env:
         GH_TOKEN: ${{ inputs.gh_token }}
         APPLE_ID: ${{ inputs.apple_id }}
-        APPLE_ID_PASSWORD: ${{ inputs.apple_id_password }}
+        APPLE_APP_SPECIFIC_PASSWORD: ${{ inputs.apple_id_password }}
```
philrz, line 72: As we transition from the separate `electron-builder-notarize` to leveraging the notarization support that's in `electron-builder`, the name of this environment variable changes but everything else stays the same.

.github/actions/build-zui/action.yml
```diff
@@ -87,5 +87,5 @@ runs:
     - name: Check notorization with gatekeeper
       if: runner.os == 'macOS'
       run: |
-        spctl --assess --type execute --verbose --ignore-cache --no-cache dist/apps/zui/mac/*.app
+        spctl --assess --type execute --verbose --ignore-cache --no-cache dist/apps/zui/mac*/*.app
```
philrz, line 90: The wildcard is needed here because the Apple Silicon builds end up in a subdirectory `mac-arm64/` whereas the Intel ones were in `mac/`.

.github/actions/setup-zui/action.yml
```diff
@@ -8,10 +8,13 @@ runs:
...
         go-version: '1.21'
 
+    # Caching is disabled because it resulted in getting x86_64 Zed binaries
+    # on arm64 builds. See:
+    # - https://github.com/actions/setup-node/issues/1008
     - name: Install Node
       uses: actions/setup-node@v3
       with:
-        cache: yarn
+        # cache: yarn
```
philrz, line 17: In an early test iteration, @mattnibs spotted that the `zed` binaries bundled with Zui were still showing up as Intel. I ultimately traced it to a side effect of this caching, then found open issue https://github.com/actions/setup-node/issues/1008 that seems to indicate this is a known problem. I briefly went down the path of trying to find another way to keep caching in play via some other config, but I also noticed that disabling caching wasn't slowing down the build process much at all. In the interest of simplicity I figured we could just live without it for now. I'm Subscribed to that issue so if/when it gets addressed I could revisit this. In the meantime I've included this comment so someone doesn't turn it on again without recognizing the effect it'll have on the macOS builds.

.github/actions/upload-build-artifacts/action.yml
```diff
@@ -1,14 +1,25 @@
...
 runs:
   using: 'composite'
   steps:
     - uses: actions/upload-artifact@v3
       with:
-        name: Mac Artifact
+        name: Mac Artifact (${{ runner.arch }})
         path: dist/apps/zui/*.dmg
 
+    - name: Merge latest-mac.yml Mac release files for Intel/ARM64
```
mattnibs, line 15: Would go with either `Intel/Apple` or `amd64/arm64`.

.github/workflows/build-insiders.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     needs: check_latest
     strategy:
       matrix:
-        platform: [windows-2019, macos-12, ubuntu-20.04]
+        platform: [windows-2019, macos-12, macos-14, ubuntu-20.04]
```
mattnibs, line 32: Maybe have a comment here that says `macos-14` is for Apple Silicon builds since this isn't immediately obvious looking at the platform names.

.github/workflows/build.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   release:
     strategy:
       matrix:
-        platform: [macos-12, ubuntu-20.04, windows-2019]
+        platform: [macos-12, macos-14, ubuntu-20.04, windows-2019]
```
mattnibs, line 13: Consider adding a platform comment as suggested above.

apps/zui/electron-builder.json
```diff
@@ -10,11 +10,11 @@
     "sign": "./scripts/sign.js"
   },
   "linux": {"target": ["deb", "rpm"]},
+  "mac": {"entitlements": "darwin.plist", "notarize": {"teamId": "2DBXHXV7KJ"}},
```
philrz, line 13: The `teamId` value here is redundant with the value in the `APPLE_TEAM_ID` environment variable, but it's needed here due to open issue https://github.com/electron-userland/electron-builder/issues/8103. I'm Subscribed to that issue so I can revisit this if/when it gets fixed. Also, [found consensus](https://stackoverflow.com/questions/44417171/should-i-keep-the-apple-developer-team-id-secret) validated my guess that this team ID does not seem to be sensitive info (e.g., similar to a company/domain name), so I've gone ahead and put the value here rather than adding yet more code to populate it from a Secret.

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,227 @@
...
+
+const client = new Octokit({
+  auth: token
+});
+
+// These are derived from settings in package.json so the script will work on
+// both regular Zui and Zui Insiders.
+const OWNER = pkg.repository.split('/')[3];
+const REPO = pkg.repository.split('/')[4];
+const PRODUCT_NAME= pkg.productName.replaceAll(' ', '-');
```
philrz, line 20: This `replaceAll` was needed because the build filenames that start out with space characters in them (e.g., `Zui - Insiders-1.7.1-19.dmg`) have those spaces turned into hyphens by the time they're uploaded to GItHub. I briefly explored using this as an opportunity drop those space characters entirely, but post-update that ends up putting the new app state & saved data in a fresh, new directory `~/Library/Application Support/Zui-Insiders` hence losing the prior state/data from the traditional `~/Library/Application Support/Zui - Insiders`, hence that would require a new migration logic for users to have a seamless update. Yuck! Hence the `replaceAll` to just keep going with what we've got.

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,227 @@
...
+    // avoids moving the sha512 checksum into its own line
+    lineWidth: -1,
+  };
+
+  const mergedString = yaml.dump(mergedObject, dumpOptions);
+  return mergedString;
+}
+
+const getPlatformFromLatestMacYml = (content) => {
+  const intelRe = `${PRODUCT_NAME}-${VERSION}.dmg`
```
mattnibs, line 51: Is there anyway that we could have the intel version be labled `*-amd64.dmg` or `*-intel64.dmg` (vlc calls its intel64/arm64)?

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,227 @@
...
+
+// These are derived from settings in package.json so the script will work on
+// both regular Zui and Zui Insiders.
+const OWNER = pkg.repository.split('/')[3];
+const REPO = pkg.repository.split('/')[4];
+const PRODUCT_NAME= pkg.productName.replaceAll(' ', '-');
+const REPO = 'zui';
+const URL = `/repos/${OWNER}/${REPO}/releases`;
+const VERSION = pkg.version;
+const RELEASE_NAME = (PRODUCT_NAME == 'Zui') ? 'v' + VERSION : VERSION;
```
philrz, line 24: Here's another coping mechanism. For all the time they've existed, Zui Insiders builds have names _without_ the leading `v` (e.g., [1.7.1-19](https://github.com/brimdata/zui-insiders/releases/tag/v1.7.1-19)) while regular Zui have it (e.g., [v1.7.0](https://github.com/brimdata/zui/releases/tag/v1.7.0)). Here I rationalized leaving things as they are because GitHub already does some annoying things with how it sorts releases, so I didn't want to make things even worse by also changing the naming convention.

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,227 @@
...
+const client = new Octokit({
+  auth: token
+});
+
+// These are derived from settings in package.json so the script will work on
+// both regular Zui and Zui Insiders.
+const OWNER = pkg.repository.split('/')[3];
+const REPO = pkg.repository.split('/')[4];
+const PRODUCT_NAME= pkg.productName.replaceAll(' ', '-');
+const REPO = 'zui';
```
mattnibs, line 21: You are redefining the REPO variable- I'm not sure how the compiler is allow you to do this.

philrz: @mattnibs: Thanks for catching! That was indeed a mistake, specifically when bringing changes from my personal fork back to the regular Zui code base. The potential for these kinds of mistakes is why I've been planning on subjecting @nwt to a final round of test builds made from yet another set of fresh personal fork repos using what's now in this PR, so this error would have been caught at that point. But glad to have been able to fix it now!

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
+// Based on https://github.com/electron-userland/electron-builder/issues/5592#issuecomment-2004803764
+import { Octokit } from "octokit";
```
mattnibs, line 2: Since nowhere else in the directory uses semicolons at the end of statements I think we should get rid of them in this file.

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  const currentRelease = allReleases.data.find(release => {
+    return release.name === RELEASE_NAME;
+  })
+
+  if (!currentRelease) {
+    console.log('No release found. Skipping merge')
+    return;
+  } else {
+    console.log('Release found')
+  }
```
mattnibs, line 73:
```suggestion
  }
  console.log('Release found')
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  const localLatestMacYmlExists = fs.existsSync(LOCAL_FILE_PATH);
+
+  if (!localLatestMacYmlExists) {
+    console.log(`[local] could not find ${FILE_NAME}. Skipping merge`);
+    return;
+  } else {
+    console.log(`[local] ${FILE_NAME} found`);
+  }
+
+  const localLatestMacYmlContent = fs.readFileSync(LOCAL_FILE_PATH, { encoding: 'utf8' });
```
mattnibs, line 84:
```suggestion
  if (!fs.existsSync(LOCAL_FILE_PATH)) {
    console.log(`[local] could not find ${FILE_NAME}. Skipping merge`);
    return;
  }
  console.log(`[local] ${FILE_NAME} found`);
  
  const localLatestMacYmlContent = fs.readFileSync(LOCAL_FILE_PATH, { encoding: 'utf8' });
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  const localLatestMacYmlExists = fs.existsSync(LOCAL_FILE_PATH);
+
+  if (!localLatestMacYmlExists) {
+    console.log(`[local] could not find ${FILE_NAME}. Skipping merge`);
+    return;
+  } else {
+    console.log(`[local] ${FILE_NAME} found`);
+  }
```
mattnibs, line 82:
```suggestion
  if (!fs.existsSync(LOCAL_FILE_PATH)) {
    console.log(`[local] could not find ${FILE_NAME}. Skipping merge`);
    return;
  }
  console.log(`[local] ${FILE_NAME} found`);
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  const localLatestMacYmlContent = fs.readFileSync(LOCAL_FILE_PATH, { encoding: 'utf8' });
+
+  const localPlatform = getPlatformFromLatestMacYml(localLatestMacYmlContent);
+
+  if (localPlatform === 'none' || localPlatform === 'both') {
+    console.log(`[local] ${FILE_NAME} invalid. Platform: ${localPlatform}. Skipping merge`)
+    return;
+  } else {
+    console.log(`[local] ${FILE_NAME} valid: Platform: ${localPlatform}`)
+  }
```
mattnibs, line 93:
```suggestion
  if (localPlatform === 'none' || localPlatform === 'both') {
    console.log(`[local] ${FILE_NAME} invalid. Platform: ${localPlatform}. Skipping merge`)
    return;
  }
  console.log(`[local] ${FILE_NAME} valid: Platform: ${localPlatform}`)
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+    const originalAsset = currentRelease.assets.find(asset => {
+      return asset.name === FILE_NAME;
+    })
+
+    if (!originalAsset) {
+      console.log(`[remote] ${FILE_NAME} not found. Skipping merge`);
+      return;
+    } else {
+      console.log(`[remote] ${FILE_NAME} found`)
+    }
```
mattnibs, line 172: I would move this stuff out of the try block since that it was you are doing elsewhere in this file. Also you can drop the else and just console log.

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  try {
+    await client.rest.repos.uploadReleaseAsset({
+      url: uploadUrl,
+      headers: {
+        'content-type': 'application/octet-stream',
+        'content-length': Buffer.byteLength(mergedContent),
+      },
+      name: FILE_NAME,
+      data: assetStream,
+    })
```
mattnibs, line 197:
```suggestion
  try {
    await client.rest.repos.uploadReleaseAsset({
      url: uploadUrl,
      headers: {
        'content-type': 'application/octet-stream',
        'content-length': Buffer.byteLength(mergedContent),
      },
      name: FILE_NAME,
      data: Readable.from(mergedContent),
    })
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  try {
+    await client.rest.repos.uploadReleaseAsset({
+      url: uploadUrl,
+      headers: {
+        'content-type': 'application/octet-stream',
+        'content-length': Buffer.byteLength(localLatestMacYmlContent),
+      },
+      name: `latest-mac-${localPlatform}.yml`,
+      data: localAssetStream,
+    })
```
mattnibs, line 124:
```suggestion
  try {
    await client.rest.repos.uploadReleaseAsset({
      url: uploadUrl,
      headers: {
        'content-type': 'application/octet-stream',
        'content-length': Buffer.byteLength(localLatestMacYmlContent),
      },
      name: `latest-mac-${localPlatform}.yml`,
      data: Readable.from(localLatestMacYmlContent),
    })
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  })
+
+  if (!remotePlatformFileExists) {
+    console.log(`[remote] latest-mac-${remotePlatform}.yml does not exist. Skipping merge`)
+    return;
+  } else {
+    console.log(`[remote] latest-mac-${remotePlatform}.yml found`)
+  }
+
+  let remotePlatformFile = null;
```
mattnibs, line 145:
```suggestion
  let remotePlatformFile;
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+      }
+    });
+    console.log(`[remote] latest-mac-${remotePlatform}.yml downloaded`)
+  } catch(e) {
+    console.log(`[remote] error downloading latest-mac-${remotePlatform}.yml. Skipping merge`);
+    console.log(e);
+    return;
+  }
+
+  const remoteLatestMacYmlContent = new TextDecoder().decode(remotePlatformFile.data);
```
mattnibs, line 160: Pretty sure remotePlatformFile.data is a Buffer so you should be able to do the following and not need the TextDecoder import:
```suggestion
  const remoteLatestMacYmlContent =remotePlatformFile.data.toString();
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+
+  try {
+    const originalAsset = currentRelease.assets.find(asset => {
+      return asset.name === FILE_NAME;
+    })
```
mattnibs, line 165: nit: I like this better
```suggestion
    const originalAsset = currentRelease.assets.find(asset => asset.name === FILE_NAME)
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+    console.log(`[remote] error uploading merged ${FILE_NAME}. Skipping merge`)
+    console.log(e);
+    return;
+  }
+
+  // cleanup
+  const updatedRelease = await client.request(`GET ${URL}`);
+  const updatedCurrentRelease = updatedRelease.data.find(release => {
+    return release.name === RELEASE_NAME;
+  })
```
mattnibs, line 209: nit:
```suggestion
  const updatedCurrentRelease = updatedRelease.data.find(release => release.name === RELEASE_NAME)
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+    console.log(`[remote] error uploading latest-mac-${localPlatform}.yml. Skipping merge`);
+    console.log(e);
+    return;
+  }
+
+  const remotePlatform = localPlatform === 'intel' ? 'arm' : 'intel';
+
+  const remotePlatformFileExists = currentRelease.assets.find(asset => {
+    return asset.name === `latest-mac-${remotePlatform}.yml`;
+  })
```
mattnibs, line 136: nit:
```suggestion
  const remotePlatformFileExists = currentRelease.assets.find(asset => asset.name === `latest-mac-${remotePlatform}.yml`)
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+
+  mergedObject.files = [
+    ...intelObject.files,
+    ...armObject.files,
+  ];
+
+  const dumpOptions = {
+    // avoids moving the sha512 checksum into its own line
+    lineWidth: -1,
+  };
```
mattnibs, line 43: nit:
```suggestion
  // avoids moving the sha512 checksum into its own line
  const opt = { lineWidth: -1 };
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+  const armObject = yaml.load(arm);
+
+  const mergedObject = {
+    ...intelObject,
+  };
+
+  mergedObject.files = [
+    ...intelObject.files,
+    ...armObject.files,
+  ];
```
mattnibs, line 38:
```suggestion
  const mergedObject = {
    ...intelObject,
    files: [...intelObject.files, ...armObject.files]
  };
```

apps/zui/scripts/merge-mac-release-files.mjs
```diff
@@ -0,0 +1,226 @@
...
+
+  const dumpOptions = {
+    // avoids moving the sha512 checksum into its own line
+    lineWidth: -1,
+  };
+
+  const mergedString = yaml.dump(mergedObject, dumpOptions);
+  return mergedString;
```
mattnibs, line 46:
```suggestion
  return yaml.dump(mergedObject, dumpOptions);
```

apps/zui/scripts/merge-mac-release-files.mjs
mattnibs, line 38:
```suggestion
  const mergedObject = {
    ...intelObject,
    files: [...intelObject.files, ...armObject.files]
  };
```

.github/actions/upload-build-artifacts/action.yml
philrz: @mattnibs: As long as I'm changing it, how would you feel about `x86/arm64`? I ask because that matches what `electron-builder` shows for the `arch` value in respective builds. E.g., Intel:

```
  • packaging       platform=darwin arch=x64 electron=30.0.1 appOutDir=/Users/runner/work/zui/zui/dist/apps/zui/mac
```

Apple Silicon:

```
  • packaging       platform=darwin arch=arm64 electron=30.0.1 appOutDir=/Users/runner/work/zui/zui/dist/apps/zui/mac-arm64
```

apps/zui/scripts/merge-mac-release-files.mjs
philrz: @mattnibs: I looked into it and did find a way, so I've just pushed a commit that makes the change. I just tested it out in my personal fork and you can see the improvement in https://github.com/philrz/zui/releases/tag/v1.7.5. I just tested it out in my Intel Mac by starting from the previous release https://github.com/philrz/zui/releases/tag/v1.7.4 when the filename was just `Zui-1.7.4.dmg` and it auto-updated to the newer release with filename `Zui-1.7.5-x64.dmg` and restarted without complaint. If you'd like to repeat that test with the arm64 builds on your Apple Silicon Mac that's be great, but I'm planning on a final test after I've resolved all PR feedback, so, whenever. I see no reason why it wouldn't work since this change should have been a no-op for those builds, but I've seen some wacky stuff on this journey, so I've become quite paranoid. 😛

apps/zui/electron-builder.json
```diff
@@ -10,11 +10,15 @@
     "sign": "./scripts/sign.js"
   },
   "linux": {"target": ["deb", "rpm"]},
+  "mac": {
+    "entitlements": "darwin.plist",
+    "notarize": {"teamId": "2DBXHXV7KJ"},
+    "artifactName": "${productName}-${version}-${arch}.${ext}"
```
philrz, line 16: At first when using `electron-builder` at its defaults, we'd get a a DMG for Intel like `Zui-1.7.4.dmg` and for Apple Silicon like `Zui-1.7.4-arm64.dmg`. This use of `artifactName` gives us symmetry by having the Intel one now have the arch included as well and hence `Zui-1.7.4-x64.dmg`.

apps/zui/scripts/merge-mac-release-files.mjs
philrz: @mattnibs: I tested out the proposed change and it didn't work as is. I looked closer and found that the `.data` in the GitHub response is an `ArrayBuffer`, not a `Buffer`. I'm not super familiar with this but I did some reading/searching about best practices for converting ArrayBuffer to strings and the [first StackOverflow post I found](https://stackoverflow.com/questions/6965107/converting-between-strings-and-arraybuffers) has a top-voted answer of folks recommending `TextDecoder`. Therefore I'm inclined to leave this as is for now.

nwt (approved):

Issue #1588, opened 2021-04-27
build-id conflict during RPM install


philrz:
A community user reported on Slack:

zmiimz:
I have the same case during rpm package installation

> sudo rpm -ihv --nodeps zui-1.4.1.x86_64.rpm 
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
        file /usr/lib/.build-id/51/74642779eb830026a665bff7d1869207a574ac from install of zui-1.4.1-1.x86_64 conflicts with file from package xmind-vana-23.11.4336-202311170221.x86_64
        file /usr/lib/.build-id/7b/bde14ca6192338b4c5d44af06e90ca06276940 from install of zui-1.4.1-1.x86_64 conflicts with file from package xmind-vana-23.11.4336-202311170221.x86_64
        file /usr/lib/.build-id/8b/060bea4cd4260d05f778be1856691d20377fcb from install of zui-1.4.1-1.x86_64 conflicts with file from package xmind-vana-23.11.4336-202311170221.x86_64
        file /usr/lib/.build-id/90/9d0b5007ba7bf20d9e4205235b3fe5a9f3309e from install of zui-1.4.1-1.x86_64 conflicts with file from package xmind-vana-23.11.4336-202311170221.x86_64

philrz:
Thanks for letting us know @zmiimz. We unfortunately have not been able to fix this. Please do let us know if you were able to workaround the problem with the `--replacefiles` approach described above or any other ways.

philrz:
Per https://github.com/brimdata/zui/pull/3084#issuecomment-2159007093, I've noticed that this can also occur when attempting to install a Zui RPM when a Zui Insiders release is already installed.

PR #3084, opened 2024-06-07
New electron-updater

### LINKED ISSUE #3082: Linux update bugs
## tl;dr

The heavy testing I did related to the merge of #3077 led to the discovery of some Zui update issues on Linux. I've concluded that moving to the latest [electron-updater](https://yarnpkg.com/package?name=electron-updater) version along with a modest set of Zui code changes will address these issues and also put us in position to hopefully start doing auto-updates on Linux soon (#871).

## Background

On macOS and Windows our [mac-win-updater.ts](https://github.com/brimdata/zui/blob/main/apps/zui/src/domain/updates/mac-win-updater.ts) uses [electron-update's autoUpdater](https://www.electron.build/auto-update.html). Because that autoUpdater traditionally did not support Linux, we rolled our own [linux-updater.ts](https://github.com/brimdata/zui/blob/main/apps/zui/src/domain/updates/linux-updater.ts) to act as an "update helper" of sorts. It asks about the latest available release on GitHub by polling an endpoint of the [update.electronjs.org](https://github.com/electron/update.electronjs.org) service and, if the service responds indicating a newer release is available, the Zui user is shown a pop-up notification. If the user then clicks the **Install** button in the pop-up, the [brimdata.io Download page](https://www.brimdata.io/download/) page is opened in their browser (or the [Zui Insiders "latest release" page on GitHub](https://github.com/brimdata/zui-insiders/releases/latest) if they're running Insiders).

This has always been a little hacky because our code actually checks for a later _non-Linux_ release (specifically, macOS) on the assumption that a newer Linux artifact of that same release would also be present. This isn't 100% bulletproof, since it's possible for an Actions build job to fail part way through such that it produces usable artifacts for some platforms but not others. But it's generally been seen as better than just requiring Linux users to always manually check for newer releases.

The two problems with this approach described below are both related to this polling of the Electron update server:

https://github.com/brimdata/zui/blob/619f5d890edd8dbe8cf1e2c90ae3ebdcbd9b2b9d/apps/zui/src/domain/updates/linux-updater.ts#L35

## Problem 1 - The new artifact naming convention makes releases look ineligible as update targets

This was a detail of which I was previously unaware, but the way our Linux update helper polls the Electron update server endpoint relies on the existence of an artifact of the form `*(mac|darwin|osx).*(-arm).*.zip` to consider a release as an eligible update target. This can be seen if we manually construct a URL similar to the one our Linux update helper uses, but point at a repo that has no artifacts at all.

```
$ curl -w "%{http_code}" https://update.electronjs.org/brimdata/zui-docs-site/darwin-x64/1.2.3
No updates found (needs asset matching *(mac|darwin|osx).*(-arm).*.zip in public repository)404
```

Before the merge of #3077 (so, Zui Insiders release [v1.7.1-19](https://github.com/brimdata/zui-insiders/releases/tag/v1.7.1-19) and older) the generated artifacts always included a file of this format (e.g., `Zui---Insiders-1.7.1-19-mac.zip`) But since the merge of #3077 (so, Zui Insiders release [v1.7.1-20](https://github.com/brimdata/zui-insiders/releases/tag/v1.7.1-20) and newer) the release automation now creates separate ZIP files for each macOS platform (e.g., `Zui---Insiders-1.7.1-23-x64.zip` and `Zui---Insiders-1.7.1-23-arm64.zip`). As you can see, these lack the word `mac` in their names, so without manual intervention or additional change to the build automation, our Linux update helper would not find recent Zui Insiders releases to be eligible update targets.

In my testing of this in personal forks, I've confirmed two additional things:

1. I can work around the problem by manually downloading the `*-x64.zip` artifact for a given release, renaming it to `*-mac.zip`, then uploading it back so it becomes available as an additional release artifact. I plan to do this with the next GA Zui release as a stopgap measure.

2. If we wanted to maintain our existing Linux update helper and avoid dependency on these ZIP filenames, we could get equivalent functionality by polling an Electron update URL for platform `win32-x64` instead of `darwin-x64`, since Windows updates don't rely on these ZIP files.

All that said, I determined that trying to fix the existing Linux update helper is probably futile due to the problem described next.

## Problem 2 - Our approach for polling the Electron update server appears to have always been broken with certain Zui Insiders build numbers

This only affects Zui Insiders releases because I've seen the problem only occurs with "prerelease" version strings (i.e., `-` followed by digits, such as in `v1.7.1-5`). My tests show that version strings that lack the hyphen (such as we use for regular Zui) don't have this problem.

We can observe the problem if we manually construct a valid URL that an older Zui Insiders release like [v1.7.1-5](https://github.com/brimdata/zui-insiders/releases/tag/v1.7.1-5) would construct upon launch to check if there's a newer release available.

```
$ curl -w "%{http_code}" https://update.electronjs.org/brimdata/zui-insiders/darwin-x64/1.7.1-5
{"name":"1.7.1-8","notes":null,"url":"https://github.com/brimdata/zui-insiders/releases/download/v1.7.1-8/Zui---Insiders-1.7.1-8-mac.zip"}200
```

Notice how this shows that the update server believes `1.7.1-8`(*) is the newest available update target, despite the fact that there's many releases (`-10`, `-11`, …) newer than that. In my testing I've confirmed that this appears to be due to the update server treating the post-hyphen version numbers alphabetically rather than numerically, e.g., `-8` looks "higher" than `-10`. This is an annoying pattern I've seen in other contexts (#2852). It definitely looks like a bug since the [semver spec](https://semver.org/) says quite clearly RE: pre-release versions "Identifiers consisting of only digits are compared numerically". We could raise this as a update.electronjs.org issue and/or try to fix it ourselves and submit a PR. But since the update server is ultimately out of our control, it would probably take a while to force that change through and be able to take advantage, and that would all be in pursuit of maintaining a semi-hacky approach. Therefore it seems worth checking if there's another way.

(*) - Note: If there'd been a `1.7.1-9` build that would have been returned as an eligible update target here, but that one doesn't happen to exist due to a failed Actions build run.

## Conclusion

Our electron-updater dependency has been pointing at `^4.3.8` and the latest available electron-updater is now `6.2.1`. Hoping it could be a quick fix, I tried pointing at this latest electron-updater and making no other Zui code changes to see either/both of these problems went away. Not only did they not go away, I found enough other things had changed between those electron-updater releases that break update logic for _non-Linux_ Insiders releases as well! (I'll cover details about that in an upcoming PR.) However, at that point I'd spent enough time hacking and looking at debug logs that I can see a set of changes that would allow us to jump to the newer electron-updater, drop the buggy parts of our homegrown Linux update helper, and have updates working correctly on all platforms.

A further advantage to doing this is that in recent releases the electron-update autoUpdater has introduced beta support for auto-updates on Linux (https://github.com/electron-userland/electron-builder/pull/7060). My first attempt at using it in a personal fork test didn't work out so maybe it's too early to rely on 100%, but getting us on to a newer electron-update will get us one step closer.

philrz:
Fixes #3082, which has detailed background on what's being fixed here and what else we expect to gain by moving to a newer `electron-updater`.

Starting from the code on this branch, I've done an exhaustive set of successful upgrade tests in personal fork repos of Zui and Zui Insiders on both macOS and Ubuntu (`.deb`). If this PR should get approved, after having updated the branch with any suggestions from reviewers, I'll catch it up with `main` and repeat the tests one last time including both Windows and Rocky Linux (`.rpm`) as well for completeness. I happened to skip those on the first pass since in the past I've found update behavior on Windows closely tracks what we see on macOS and likewise between the different Linux installers, so I figured I'd save a little time on the first pass since I assumed I'd want to re-test again after merging suggestions anyway.

Here's a complete run down of all the tests I performed and the results.

1. Make personal forks of both Zui and Zui Insiders repos per instructions [here](https://lookylabs.atlassian.net/wiki/spaces/PROD/pages/1804730369/Making+a+Personal+Fork+of+the+App+Repo)
   * I manually staged an initial Zui Insiders release based on `v1.6.1-33` from the "official" Insiders repo so that way there'd be an existing release to increment from. I then deleted the release and its tag after I made the first proper Insiders release in my fork repo.
   * In the Zui fork, I made the repo name changes from `brimdata` to `philrz` in both the `main` branch and the `release/v1.7.0` branch

2. In the Zui fork, delete the `v1.7.0` tag and create a new `v1.7.0` release based off the `release/v1.7.0` branch
   * This will serve as a baseline similar to our production users running GA Zui `v1.7.0` today, which has the old electron-updater

3. In the Zui Insiders fork, create a new release based off the `release/v1.7.0` branch
   * This will serve as a baseline similar to our production users running GA Zui Insiders builds that pre-date the changes in #3077
   * Test result: Zui Insiders release `v1.7.0` was created

4. In the Zui Insiders fork, create a new release in my personal repo based off the tip of `main`
   * This will serve as a baseline similar to our production users running GA Zui Insiders builds in the time since the changes from #3077 have merged
   * Test result: Zui Insiders release `v1.7.1-0` was created
   * Test result: On my Mac, I installed the Insiders `v1.7.0` and it found/auto-updated to `v1.7.1-0`
   * Test result: On my Linux VM, I installed the Insiders `v1.7.0` and it did **not** initially see the `v1.7.1-0` as an eligible update target, but this is due to the known "Problem 1" described in #3082. Therefore I manually downloaded the `Zui---Insiders-1.7.1-0-x64.zip` artifact, renamed it to `Zui---Insiders-1.7.1-0-mac.zip`, and uploaded that to the Release. After waiting about 15 minutes, now the app _did_ find `v1.7.1-0` as an eligible update target.

5. In the Zui fork, advance the version string in `package.json` to `1.8.0` and merge the `new-electron-updater` PR
   * This is effectively the same step we'll do before prepping the next GA Zui release.

6. In the Zui fork, create a new `v1.8.0` release
   * Since the older `v1.7.0` Linux release still has the "Problem 1" from #3082, I manually added the `Zui-1.8.0-mac.zip` to its Release page so Linux users would see this release as an eligible update target
   * Test result: On my Mac, I installed the `v1.7.0` release and it auto-updated to `v1.8.0`
   * Test result: On my Linux VM, I installed the `v1.7.0` release and it saw `v1.8.0` as an eligible update target

7. In the Zui Insiders fork, create a new Zui Insiders release
   * Test result: Zui Insiders release `v1.8.0` was created. Since the older Insiders releases still have the "Problem 1" from #3082, I manually added the `Zui---Insiders-1.8.0-mac.zip` to its Release page so Linux users would see this release as an eligible update target.
   * Test result: On my Mac, I installed the `v1.7.0` release and it auto-updated to the `v1.8.0` release
   * Test result: On my Mac, I installed the `v1.7.1-0` release and it auto-updated to the `v1.8.0` release
   * Test result: On my Linux VM, I installed the `v1.7.0` release and it saw `v1.8.0` as an eligible update target
   * Test result: On my Linux VM, I installed the `v1.7.1-0` release and it saw the `v1.8.0` release as an eligible update target.

8. In the Zui Insiders fork, trigger creation of releases `-insiders.1`, `-insiders.12`, ... , `-insiders.10`
   * Test result: On my Mac, I installed `v1.7.0` and auto-updated to `v1.8.1-insiders.10`
   * Test result: On my Mac, I installed `v1.7.1-0` and auto-updated to `v1.8.1-insiders.10`
   * Test result: On my Mac, I installed `v1.8.0` and auto-updated to `v1.8.1-insiders.10`
   * Test result: On my Mac, I installed `v1.8.1-insiders.9` it auto-updated to `v1.8.1-insiders.10`
   * Test: On my Linux VM, I installed `v1.7.0` and it found `v1.8.0` as the latest available, but of course clicking the **Install** button brought up the `/releases/latest` page on GitHub such that I was shown `v1.8.1-insiders.10` as the update target
   * Test result: On my Linux VM, I installed `v1.7.0` and it found `v1.8.0` as the latest available, but of course clicking the Install button brought up the `/releases/latest` page on GitHub such that I was shown `v1.8.1-insiders.10` as the update target
   * Test result: On my Linux VM, I installed `v1.7.1-0` and it found `v1.8.0` as the latest available, but of course clicking the **Install** button brought up the `/releases/latest` page on GitHub such that I was shown `v1.8.1-insiders.10` as the update target
   * Test result: On my Linux VM, I installed `v1.8.0` and it found `v1.8.1-insiders.10` as the latest available (and I'd not had to manually add any "mac ZIP" file for the release)
   * Test result: On my Linux VM, I installed  `v1.8.1-insiders.9` and it found `v1.8.1-insiders.10 `as the latest available (and I'd not had to manually add any "mac ZIP" file for the release)

9. In the Zui fork, create a new `v1.9.0` release
   * Test result: On my Mac, I installed the `v1.7.0` release and it auto-updated to the `v1.9.0` release
   * Test result: On my Mac, I installed the `v1.8.0` release and it auto-updated to the `v1.9.0` release
   * Test result: On my Linux VM, I installed `v1.7.0` and it found `v1.8.0` as the latest available, but of course clicking the Install button brought up the brimdata.io **Downloads** that would offer `v1.9.0` as the update target
   * Test result: On my Linux VM, I installed `v1.8.0` and it found `v1.9.0` as the latest available (and I'd not had to manually add any "mac ZIP" file for the release)

Conclusion: Once users get on to one of the builds running the newer `electron-updater`, they'll find the proper "latest" release for both Zui and Zui Insiders without any manual intervention on our part. The test results indicate that auto-update users (macOS) will make this transition without any manual intervention on our part. On Linux, we'll need to manually provide the "mac ZIP" file as part of the first releases that have the newer `electron-updater`, but once they make that leap, their apps will keep finding "latest" releases based on just the presence of the `latest-linux.yml` that's created automatically by our build process. And since our Linux "helper" just lands them on a "manually download the latest" web page whenever a newer release is detected, they complete this hop easily.

### DIFF
--- a/apps/zui/package.json
+++ b/apps/zui/package.json
@@ -91,7 +91,7 @@
     "electron-localshortcut": "^3.2.1",
     "electron-log": "5.0.0-beta.6",
     "electron-mock-ipc": "^0.3.12",
-    "electron-updater": "^4.3.8",
+    "electron-updater": "6.2.1",
     "esbuild": "^0.18.14",
     "eslint": "^8.11.0",
     "eslint-config-prettier": "^8.5.0",
--- a/apps/zui/src/domain/updates/linux-updater.ts
+++ b/apps/zui/src/domain/updates/linux-updater.ts
@@ -13,7 +13,6 @@ autoUpdater.forceDevUpdateConfig = true
 export class LinuxUpdater implements Updater {
   async check() {
     const {updateInfo} = await autoUpdater.checkForUpdates()
-    console.log(updateInfo)
     const latest = updateInfo.version
     const current = app.getVersion()
     if (semver.lt(current, latest)) {
--- a/apps/zui/src/domain/updates/linux-updater.ts
+++ b/apps/zui/src/domain/updates/linux-updater.ts
@@ -13,6 +13,7 @@ autoUpdater.forceDevUpdateConfig = true
 export class LinuxUpdater implements Updater {
   async check() {
     const {updateInfo} = await autoUpdater.checkForUpdates()
+    console.log(updateInfo)
     const latest = updateInfo.version
     const current = app.getVersion()
     if (semver.lt(current, latest)) {
--- a/apps/zui/src/domain/updates/linux-updater.ts
+++ b/apps/zui/src/domain/updates/linux-updater.ts
@@ -1,15 +1,19 @@
-import {app, shell} from "electron"
-import fetch from "node-fetch"
+import {autoUpdater} from "electron-updater"
+import {Updater} from "./types"
 import semver from "semver"
+import {app, shell} from "electron"
 import env from "src/app/core/env"
 import links from "src/app/core/links"
 import pkg from "src/electron/pkg"
-import {Updater} from "./types"
-import {getMainObject} from "src/core/main"
+
+autoUpdater.autoDownload = false
+autoUpdater.autoInstallOnAppQuit = false
+autoUpdater.forceDevUpdateConfig = true
 
 export class LinuxUpdater implements Updater {
   async check() {
-    const latest = await this.latest()
+    const {updateInfo} = await autoUpdater.checkForUpdates()
+    const latest = updateInfo.version
     const current = app.getVersion()
     if (semver.lt(current, latest)) {
       return latest
@@ -20,19 +24,6 @@ export class LinuxUpdater implements Updater {
 
   async install() {
     shell.openExternal(this.downloadUrl())
-  }
-
-  private async latest() {
-    const resp = await fetch(this.latestUrl())
-    if (resp.status === 204) return app.getVersion()
-    const data = await resp.json()
-    return data.name
-  }
-
-  private latestUrl() {
-    const repo = getMainObject().appMeta.repo
-    const platform = "darwin-x64" // If the mac version exists, the linux does too
-    return `https://update.electronjs.org/${repo}/${platform}/${app.getVersion()}`
   }
 
   private downloadUrl() {
--- a/apps/zui/src/domain/updates/mac-win-updater.ts
+++ b/apps/zui/src/domain/updates/mac-win-updater.ts
@@ -11,7 +11,6 @@ autoUpdater.forceDevUpdateConfig = true
 export class MacWinUpdater implements Updater {
   async check() {
     const {updateInfo} = await autoUpdater.checkForUpdates()
-    console.log(updateInfo)
     const latest = updateInfo.version
     const current = app.getVersion()
     if (semver.lt(current, latest)) {
--- a/apps/zui/src/domain/updates/mac-win-updater.ts
+++ b/apps/zui/src/domain/updates/mac-win-updater.ts
@@ -6,6 +6,7 @@ import {getMainObject} from "src/core/main"
 
 autoUpdater.autoDownload = false
 autoUpdater.autoInstallOnAppQuit = false
+autoUpdater.forceDevUpdateConfig = true
 
 export class MacWinUpdater implements Updater {
   async check() {
--- a/apps/zui/src/domain/updates/mac-win-updater.ts
+++ b/apps/zui/src/domain/updates/mac-win-updater.ts
@@ -11,6 +11,7 @@ autoUpdater.forceDevUpdateConfig = true
 export class MacWinUpdater implements Updater {
   async check() {
     const {updateInfo} = await autoUpdater.checkForUpdates()
+    console.log(updateInfo)
     const latest = updateInfo.version
     const current = app.getVersion()
     if (semver.lt(current, latest)) {
--- a/apps/insiders/src/insiders-packager.ts
+++ b/apps/insiders/src/insiders-packager.ts
@@ -40,7 +40,7 @@ export class InsidersPackager {
     const version =
       this.strategy === 'match-stable'
         ? this.stableVersion
-        : semver.inc(this.lastVersion, 'prerelease');
+        : semver.inc(this.lastVersion, 'prerelease', 'insiders');
 
     return version;
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5579,7 +5579,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/semver@npm:^7.3.3, @types/semver@npm:^7.3.4":
+"@types/semver@npm:^7.3.3":
   version: 7.3.4
   resolution: "@types/semver@npm:7.3.4"
   checksum: 577dc7f607fbf46cf185f61ba726e4cfecea41c4b17e55069e70a64b6cd92658f8286aa075a607303455658e87ed045d3d14c32a513c0eff381d64aecd7b72a7
@@ -7296,16 +7296,6 @@ __metadata:
     base64-js: ^1.3.1
     ieee754: ^1.1.13
   checksum: e2cf8429e1c4c7b8cbd30834ac09bd61da46ce35f5c22a78e6c2f04497d6d25541b16881e30a019c6fd3154150650ccee27a308eff3e26229d788bbdeb08ab84
-  languageName: node
-  linkType: hard
-
-"builder-util-runtime@npm:8.7.3":
-  version: 8.7.3
-  resolution: "builder-util-runtime@npm:8.7.3"
-  dependencies:
-    debug: ^4.3.2
-    sax: ^1.2.4
-  checksum: eaf55ffcbde90a65639d7e102626ea819d0c0e4b296984755dab8d9df2bd85c56fb586e536a175b9229477d7e122c8f553aa232827aa7e275b43274b87a6e3e5
   languageName: node
   linkType: hard
 
@@ -9219,18 +9209,19 @@ __metadata:
   languageName: node
   linkType: hard
 
-"electron-updater@npm:^4.3.8":
-  version: 4.3.8
-  resolution: "electron-updater@npm:4.3.8"
+"electron-updater@npm:6.2.1":
+  version: 6.2.1
+  resolution: "electron-updater@npm:6.2.1"
   dependencies:
-    "@types/semver": ^7.3.4
-    builder-util-runtime: 8.7.3
-    fs-extra: ^9.1.0
-    js-yaml: ^4.0.0
-    lazy-val: ^1.0.4
+    builder-util-runtime: 9.2.4
+    fs-extra: ^10.1.0
+    js-yaml: ^4.1.0
+    lazy-val: ^1.0.5
+    lodash.escaperegexp: ^4.1.2
     lodash.isequal: ^4.5.0
-    semver: ^7.3.4
-  checksum: 1814fe7c99eb6c1cb2251fc24eb62f51df716e44c891d5433e8e036a77ff2beec241249f7ba26b33a60cf134158c2160a74117dff7751b4d2e59c57e688ae17a
+    semver: ^7.3.8
+    tiny-typed-emitter: ^2.1.0
+  checksum: 92a064610a3c9df747dce9c3eccd69c48adb2c8b37b9d7c13d1c39f7e2a9ffaef4e909ab50e6973d557566bde6de7ec9c64e2cc3a2cdef18b3220c5d91333e1c
   languageName: node
   linkType: hard
 
@@ -13287,7 +13278,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"js-yaml@npm:4.1.0, js-yaml@npm:^4.0.0, js-yaml@npm:^4.1.0":
+"js-yaml@npm:4.1.0, js-yaml@npm:^4.1.0":
   version: 4.1.0
   resolution: "js-yaml@npm:4.1.0"
   dependencies:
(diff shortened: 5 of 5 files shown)


apps/zui/package.json
```diff
@@ -91,7 +91,7 @@
     "electron-localshortcut": "^3.2.1",
     "electron-log": "5.0.0-beta.6",
     "electron-mock-ipc": "^0.3.12",
-    "electron-updater": "^4.3.8",
+    "electron-updater": "6.2.1",
```
philrz, line 94: Before we were using the `^` to automatically update to minor/patch `electron-updater` releases as they became available. I've now spent enough time testing through the release history of `electron-updater` that I don't wholly trust the stewardship of the project to maintain stability between major releases. We also don't have automated test coverage in this area, which is more risk. Therefore since I've done so many days of testing on this `6.2.1` release I'm keen to lock us in place until we have a specific reason to move to something newer and make the appropriate investment in another round of testing at that time.

apps/zui/src/domain/updates/linux-updater.ts
```diff
@@ -1,15 +1,19 @@
...
+
+autoUpdater.autoDownload = false
+autoUpdater.autoInstallOnAppQuit = false
+autoUpdater.forceDevUpdateConfig = true
 
 export class LinuxUpdater implements Updater {
   async check() {
-    const latest = await this.latest()
+    const {updateInfo} = await autoUpdater.checkForUpdates()
+    const latest = updateInfo.version
```
philrz, line 16: This is the big payoff. What I'm adding here plus the code removal in this same file effectively drops our old "Linux update helper" (the one that tried to look for a newer macOS release as a sign that a newer Linux release is surely available as well) and allows us to use the autoUpdater to check for the latest Linux release. This approach became possible when `electron-updater` added the "beta" support for Linux autoUpdate in https://github.com/electron-userland/electron-builder/pull/7060 (so, it became part of `electron-updater` v6). This includes the creation of the `latest-linux.yml` build metadata which is why we can now just call `autoUpdater.checkForUpdates()` and get a correct answer for "What's the latest Linux release available relative to the one I'm running?" (i.e., it fixes the "Problem 2" described in #3082).

FWIW, I did test out the "beta" support for true Linux autoUpdate and it didn't work for me, complaining of sha512 checksum mismatch between the build itself and its metadata. Since the scope of my testing/changes was already significant enough, plus it's a "beta" feature after all, I chose to not dig further into that for now and instead maintain feature parity with what we already provide on Linux, i.e., check for a newer release and pop up a notification, but clicking **Install** just brings up the page in the user's browser where they can download the latest.

I recognize that I'm effectively duplicating some code from `mac-win-updater.ts` here, but since I'm nervous about my JavaScript skills I wanted to minimize the degree to which I was modifying existing code. @jameskerr if you'd like to push a commit to de-duplicate any of what I've done between these two updater files that's fine with me, since I'm planning on doing a final round of testing after getting approval on the PR anyway. Or if you're ok living with some duplication, once I figure out what it'd take to get Linux autoUpdate working (something I'd definitely want to do soon) we should be able to collapse down to a single file of updater code at that time anyway.

apps/zui/src/domain/updates/mac-win-updater.ts
```diff
@@ -6,6 +6,7 @@ import {getMainObject} from "src/core/main"
 
 autoUpdater.autoDownload = false
 autoUpdater.autoInstallOnAppQuit = false
+autoUpdater.forceDevUpdateConfig = true
```
philrz, line 9: I previously had not been using it since I always opt to test in personal fork repos, but during my test work I became aware of the `apps/zui/dev-app-update.yml` file. This allows for quick & dirty update testing in Dev mode since it allows us to fiddle with the current `version` string in `package.json` and the `owner` & `repo` settings in this YAML file and immediately see how the app would react if it had been bundled/released and was checking for an update at the configured GitHub Releases page. Well, per https://github.com/electron-userland/electron-builder/issues/6863 and the fix in https://github.com/electron-userland/electron-builder/pull/7117, within the jump period between our prior `electron-updater` version and the latest v6.2.1, they effectively broke it and now require setting this `forceDevUpdateConfig` in order for the file to be leveraged in Dev mode.

jameskerr (approved):

philrz:
I'll also document for the record that during some of the macOS auto-updates, I've seen crashes similar to the one in the screenshot below, which happened during the auto-update from the Zui Insiders `v1.7.0` to `v1.7.1-0` in my test fork repo. I just ignored it and the auto-update seemed to complete successfully and the app relaunched automatically into `v1.7.1-0`. I'm not too concerned because I recall having seen crashes like this during other macOS auto-updates in the past and I'm guessing it's not something we can easily debug or control. However, I've attached the full [crash_report.txt](https://github.com/user-attachments/files/15776697/crash_report.txt) if anyone feels like taking a closer look.

image

philrz:
Here's another "full disclosure" finding: In my testing on Rocky Linux, we find that I can't seem to install the RPM for regular Zui while Zui Insiders is already installed. It was on RPM-based Linux that I first found and filed open issue #1588 where the conflict surfaced during an install of a newer Zui when an older Zui was already installed, which led to the guidance in the [Zui installation docs](https://zui.brimdata.io/docs/Installation#linux-installation) to explicitly uninstall first to avoid this problem, so I'm not altogether surprised that the same symptom has surfaced between Zui and Zui Insiders installs. This is unfortunate because we'd like users to be able to have both installed side-by-side, but there may be problems related to the installers that are out of our control. In any case, I'll make note of this in #1588 and perhaps at some point we can find a decent fix or workaround.

image

philrz:
During a final round of upgrade testing to include Windows and Rocky Linux (RPM installer), I ended up bumping into unrelated bug #3089. Once its fix in #3090 was merged and I caught this branch up with `main` to have the benefit of that fix, I restarted a final testing round from the beginning one more time, and this time it all checked out without hitting any new bugs, so I'm going to merge.

I did have to employ one final trick in the Windows testing. The `release/v1.7.0` branch was old enough that it was still dependent on an expired code-signing certificate, so I wasn't able to make proper `v1.7.0` releases on Windows in my fork repo from which I could check the ability to see newer Windows releases as eligible for update. Thankfully, however, the `dev-app-update.yml` approach proved useful here. To simulate running a `v1.7.0` Zui release, I checked out the `release/v1.7.0` branch and set the `dev-app-update.yml` to:

```
owner: philrz
repo: zui
provider: github
updaterCacheDirName: zui-dev-updater
```

Likewise for Zui Insiders:

```
owner: philrz
repo: zui-insiders
provider: github
updaterCacheDirName: zui-dev-updater
```

After doing a `yarn && yarn start` the app launched and did its check for newer updates and popped up a notification just like the real thing.

PR #3085, opened 2024-06-07
Parse Error UI Improvement

### LINKED ISSUE #3081: Opaque error in stacked bar area when there's no data to plot
Repro is with Zui commit bfbb55e.

When a user clicks the **+** and selects **New Query Session** it starts a new tab without a pool pin and hence there's no initial Zed query or results. As shown in the attached video, at this point the stacked bar chart area shows the error message **No date range found with 'ts'.**

https://github.com/brimdata/zui/assets/5934157/5ff00e6f-7cd5-4e26-97eb-5712a90e8f32

As detailed in #3079, this message used to be **undefined: pool not found**, which was also not ideal.

Since this state came about through an innocent set of clicks in the app, it seems we should catch this case and surface something a bit clearer and phrased less like an error.

jameskerr:
fixes #3081 by not showing the histogram if there is a parse error

- Handle Error Case Differently
- Handle missing properties in query info
- Do not show the histogram if there is a parse error
- Loosen missing pool error detection

Here are the screenshots of the various "pool missing" stats. I also improved the error message of the histogram when the pool has no data in it.

## With no pools

CleanShot 2024-06-07 at 15 49 08@2x

## With a pool

CleanShot 2024-06-07 at 15 48 57@2x

## When there is no data in the pool

CleanShot 2024-06-07 at 16 06 15@2x

### DIFF
--- a/apps/zui/src/js/state/QueryInfo/selectors.ts
+++ b/apps/zui/src/js/state/QueryInfo/selectors.ts
@@ -9,12 +9,17 @@ export const get = activeTabSelect((tab) => {
 export const getParseError = createSelector(get, (info) => info.error)
 export const getIsParsed = createSelector(get, (info) => info.isParsed)
 export const getPoolName = createSelector(get, (info) => {
-  let source = find(info.sources, {kind: "Pool"})
+  let source = find(info.sources || [], {kind: "Pool"})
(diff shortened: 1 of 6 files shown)


philrz (approved):
Tests out well for me! 👍

PR #3087, opened 2024-06-10
Editor Error Markers

jameskerr:
Markers in the editor when the query contains errors.


philrz:
I'm currently testing the branch at commit 581fad3.

Its definitely seems to be doing its thing, such as in this example where I've taken some Zed from the `join` tutorial and tried to run it despite the referenced pools not existing. The errors in the query output area were already a great improvement, but it's even better to be able to hover my mouse over the red-lined text and see the error in context.

image

Probably not a big deal, but comparing that to this parse error, I see that the error in the query output is pointing at the semicolon which indeed does seem like the right call. Meanwhile the red line in the editor is pointing a couple characters further along. Should those be matching up?

image

mattnibs:
>Probably not a big deal, but comparing that to this parse error, I see that the error in the query output is pointing at the semicolon which indeed does seem like the right call. Meanwhile the red line in the editor is pointing a couple characters further along. Should those be matching up?

@philrz yes not sure what is going on here.

mattnibs:
I think we should get rid of this error area when the errors are markers.
338760960-5becd398-6d28-4dd1-a3a3-90a72e0e6659

jameskerr:
Hi all, I've fixed the error in the source set code in the last commit. I was adding the pipe at the wrong moment. 

@mattnibs You're right that there is some duplication. The error markers are shown as the user types. Not when they press enter to submit the query. During the time they've modified the query but not yet submitted it, they'll only see the red markers. Once they press enter, they will see the error in the results view. I think this is better than displaying nothing in the results error.

So I think we should leave this as is for now. Here's a video showing the current behavior:

https://github.com/brimdata/zui/assets/3460638/b4a5bbcc-32c9-48f0-b0bd-ce787063bd3c

mattnibs (approved):

philrz (approved):

PR #3094, opened 2024-06-12
Only Render Histogram if you Can

### LINKED ISSUE #3093: Meta-query returns "SyntaxError: Unexpected end of JSON input"
Repro is with Zui commit 21abbf6.

Steps shown in the attached video:

1. Load some sample data into Zui
2. Click **+** and **New Query Session**
3. Run the [meta-query](https://zed.brimdata.io/docs/commands/zed#meta-queries) `from :branches`

At Zui commit 21abbf6 (associated with the changes in #3085) the app shows the error "SyntaxError: Unexpected end of JSON input". At prior commit, ad5657d the meta-query result was successfully shown.

https://github.com/brimdata/zui/assets/5934157/73257cd6-1ba9-4892-b6fa-3985f4a69db4

I poked through the changes in #3085 and I think I see what's going on, which is also visible in the attached packet captures 
[ad5657d.pcapng.gz](https://github.com/user-attachments/files/15794191/ad5657d.pcapng.gz) and [21abbf6.pcapng.gz](https://github.com/user-attachments/files/15794193/21abbf6.pcapng.gz) that I took when performing this operation at the respective commits. If we **Follow TCP Stream** in Wireshark on each, for the ad5657d case, we see that the API activity halts after the app gets a response to its query `from :branches\n | count()\n  | { i: count(), v: this}\n  | i > 0\n  | head 500\n  | yield v`.

image

Whereas at commit 21abbf6 that same first query result comes back much the same, but then the app marches on and issues a query `from undefined | min(ts), max(ts)` which comes back with an `HTTP 400 Bad Request`.

image

And based on my added console logging, it looks like this is getting handled as a `DefaultError` at this line:

https://github.com/brimdata/zui/blob/8cec81d3e64e73fdcf84273642eb31d348f5830e/apps/zui/src/views/results-pane/error.tsx#L11

jameskerr:
Hello! 

These changes will improve a few things.

1. If there is no pool name present in the query, this histogram will not render.
2. If you press Cmd+L (on macOS) or Ctrl+L (on Linux/Windows), the editor will receive focus.
3. If you switch tabs to a query session, the editor will receive focus automatically.

fixes #3093

### DIFF
--- a/apps/zui/src/components/zed-editor-handler.ts
+++ b/apps/zui/src/components/zed-editor-handler.ts
@@ -1,5 +1,14 @@
 export class ZedEditorHandler {
-  constructor(public monaco, public editor) {}
+  public monaco
+  public editor
+  public props
+
+  onMount(editor, monaco) {
+    this.editor = editor
+    this.monaco = monaco
+    if (this.props.autoFocus) editor.focus()
+    this.setErrors(this.props.markers)
+  }
 
   focus() {
     if (!this.mounted) return
--- a/apps/zui/src/components/zed-editor.tsx
+++ b/apps/zui/src/components/zed-editor.tsx
@@ -1,5 +1,5 @@
-import {Editor, useMonaco} from "@monaco-editor/react"
-import {useEffect, useMemo, useRef} from "react"
+import {Editor} from "@monaco-editor/react"
+import {useEffect, useRef} from "react"
 import {useSelector} from "react-redux"
 import {cmdOrCtrl} from "src/app/core/utils/keyboard"
 import Config from "src/js/state/Config"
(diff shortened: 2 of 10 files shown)


philrz (approved):
I tested out the branch and it's a functional 👍 for me.

As would be expected, the Cmd+L behavior on macOS worked as Ctrl+L on Linux, so I updated the PR text to reflect that's expected. Also, while the "focus automatically" sounds like what we want, I banged on some older releases and in everything I tried the cursor always landed in the editor as soon as I selected a tab with a Query Session anyway. But if these changes make it extra certain that'll happen a la "belt & suspenders", I'm all about it. 😄

PR #3096, opened 2024-06-14
Refactor Lake Model

jameskerr:
- Change the Lake type to LakeAttrs
- Rename LakeModel to just Lake
- Convert to Domain Model
- Update lake model usage
- Move lake.ts to src/models


apps/zui/src/models/lake.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  get port() {
+    return this.attrs.port
+  }
+
+  get version() {
+    return this.attrs.version
+  }
+}
```
jameskerr, line 42: This is the I file I changed and moved. Everything else is just updating references and slight usage.

apps/zui/src/js/models/lake.ts
```diff
@@ -1,23 +0,0 @@
...
-        id: this.id,
-        name: this.name,
-        port: this.port,
-        version: this.version,
-        authType: this.authType,
-        authData: this.authData,
-      }
-    },
-  }
-}
```
jameskerr, line 23: This is what the lake.ts model used to look like. No it looks like the code in the next comment.

philrz (approved):

PR #3099, opened 2024-06-17
Handle unions in toZedScript

### LINKED ISSUE #3098: Can't "Pivot to Values" after applying fuse
Repro is with Zui commit 942b10f.

A common workflow for our pcap users is to do a `count() by _path` to see categories of Zeek events and then right-click **Pivot to Values** to see the individual records for a particular category, e.g., `http`.

As shown in the attached video, this workflow can be foiled when Suricata are also generated from the pcap. The Suricata records lack a `_path` field and therefore end up with `error(missing)` in the output of `count() by _path`. If the user does **Pivot to Values** on the original output, everything still works fine. However, if they follow the app's guidance and apply `fuse` to first turn the aggregation output into a tidy table, the **Pivot to Values** silently fails.

https://github.com/brimdata/zui/assets/5934157/186aa0ce-7de6-4d49-bce7-6bc55b224c9d

As shown in the video, the error in DevTools console looks like:

```
Uncaught Error: Can't convert object to Zed script:

{
  "type": {
    "types": [
      {
        "kind": "primitive",
        "name": "string"
      },
      {
        "type": {
          "kind": "primitive",
          "name": "string"
        },
        "kind": "error"
      }
    ],
    "kind": "union"
  },
  "innerType": {
    "kind": "primitive",
    "name": "string"
  },
  "index": 0,
  "value": {
    "value": "http",
    "type": {
      "kind": "primitive",
      "name": "string"
    }
  }
}
```

Peeking at the code, it appears that the app is just not prepared to deal with the Zed "union" values that result from applying `fuse`. I think I see a possible fix, so I'll put up a PR.

philrz:
Fixes #3098, which has more background.

The video below shows the change having the intended benefit on the original repro shown in #3098, and also with this simpler input file `foo.ndjson`:

```
{"_path": "http", "foo": "bar"}
{"_path": 1, "foo": "baz"}
{"_path": true, "foo": "bonk"}
```

https://github.com/brimdata/zui/assets/5934157/651cffc8-13e4-48ed-9582-c0d16e1ca24f

### DIFF
--- a/apps/zui/src/js/zed-script/toZedScript.ts
+++ b/apps/zui/src/js/zed-script/toZedScript.ts
@@ -25,6 +25,7 @@ export const toFieldPath = (arg: string | string[] | zed.Field) => {
 export function toZedScript(object: unknown): string {
   if (object instanceof zed.Field) return toFieldPath(object)
   if (object instanceof zed.Primitive) return toZedScriptPrimitive(object)
+  if (object instanceof zed.Union) return toZedScript(object.value)
   if (isString(object)) return toZedScriptString(object)
   if (object instanceof Date) return toZedScriptDate(object)
   if (typeof object === "boolean") return toZedScriptBool(object)


jameskerr (approved):
Yup, that looks right to me.

PR #3103, opened 2024-06-18
Output union values in Detail panel/window

### LINKED ISSUE #2938: Zed union values rendered blank in Detail window
Display the field as empty

While the desired fields have data, but when clicking to open and display the fields, sometimes the fields are displayed empty.

Mostly there are numbers in the fields inside them, this problem is seen more often

image

philrz:
Since my approach in #3099 passed muster, that made me recall another open, union-centric issue #2938. Sure enough, as shown in the attached video, it seems a similar small change allows these union values to start being output where they weren't before.

https://github.com/brimdata/zui/assets/5934157/9e6ea525-51d7-46b0-bf68-1e30a0a37f23

Fixes #2938


jameskerr (approved):
Works for me. I wonder if the logic to should be moved inside the "field" object. Something to consider. But I'll approve this.

Issue #3071, opened 2024-05-10
Use Native Node fetch (when/if possible)


philrz:
At first there was celebration when the changes in #3061 allowed us to drop the `node-fetch` dependency. Unfortunately we learned after the fact that the change caused intermittent CI failures as described in #3063. The fix for _that_ problem came in #3069 where we switched over to using Electron's `net.fetch` in the app and going back to `node-fetch` elsewhere.

As described in https://github.com/brimdata/zui/issues/3063#issuecomment-2099167661, evidence suggests that https://github.com/nodejs/undici/issues/2026 might have been the root cause of the symptoms described in #3063, but the fix is only available in Node v21 and newer, and Electron's current latest release is still using Node v20, and based on prior history it looks like we're probably several months away from a release that will include a Node release new enough to include the fix.

This issue serves as a reminder for us to revisit the topic when a new Electron release is available to test with. The rough summary of the steps would be to create a branch where we:

1. Back out the changes in #3069
2. Put back the changes in #3061
3. Do a looping test using the repro steps in #3063 to confirm still we don't have hangs
4. If that's all successful, merge that branch to `main`, which will once again achieve the goal of dropping `node-fetch`

philrz:
It turns out we were able to drop `node-fetch` dependency by the time we merged #3069 and no longer rely on the Fetch implementation that has the https://github.com/nodejs/undici/issues/2026 bug. See https://github.com/brimdata/zui/pull/3069#issuecomment-2153536647 for details.

Closing this issue.

philrz:
We ended up reopening this one after all. The perf problem cited in #3105 led us to bring `node-fetch` back into the picture again via the changes in #3106.

That brings us back to the previous plan of record: Once there's an [Electron release](https://www.electronjs.org/docs/latest/tutorial/electron-timelines) that's bundled with a Node v21 or newer, I can check if our CI runs reliably when using the `fetch` implementation in _that_ Node, and if that's true and it also performs as well as what we're getting right now with `node-fetch`, we could once again drop the `node-fetch` dependency.

PR #3106, opened 2024-06-20
Node Fetch Is Back

### LINKED ISSUE #3105: Data loads seem much slower with Net Fetch
## tl;dr

Unfortunately, we failed to notice before the merge of #3069 that those changes there are making data loads much slower For instance, a load that previously took 5 seconds on my Intel Macbook now takes 3+ minutes.

## Details

Repro is with Zui commit cc740ad.

The first video shows how things were performing better at commit 64f50eb right before the changes in #3069 merged. Here I'm doing a Quick Load of the [ZNG-format zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zng). As it shows, the data load takes about 5 seconds.

https://github.com/brimdata/zui/assets/5934157/70e85ba8-db09-4a27-badd-3b106c3a38d6

The second video is at commit cc740ad which is associated with the changes in #3069 that has the app using Net Fetch. As it shows, the same data load is now taking 3+ minutes.

https://github.com/brimdata/zui/assets/5934157/7ae24219-138d-4bbb-a94c-f2009d594985

jameskerr:
Maybe Fixes #3105

Give this a spin @philrz


philrz (approved):
LGTM! 👍

To test, I first confirmed that these changes bring the data load from #3105 back to the ~5 second range on my Macbook. I also reconfirmed that the Zed lake service is still only listening for connections from localhost (#3056). Finally, I started up a loop of the e2e test studied in #3063 and got through 100 successes and 0 failures before I stopped it.

Finally, in anticipation of this merging, I've reopened #3071 because per that issue there's still a possible future where we could dump the `node-fetch` dependency _and_ have the reliable CI. I've put #3071 in my backlog as a reminder to keep an eye on the Electron releases and when one comes out that ships with a new enough Node version I can re-test then.

PR #3114, opened 2024-06-25
Show an error message when pcap slices can't be opened

### LINKED ISSUE #1379: Provide guidance on macOS/Linux when no app is tied to opening of pcaps
A community user recently [reported a problem](https://brimsec.slack.com/archives/C010DR0HHMF/p1611371926002900) with Wireshark not opening when the Packets button was clicked in Brim. This inspired me to revisit the user experience when they're running Brim on an OS that lacks any installed/configured app tied to the opening of pcaps (e.g. Wireshark not installed). Repros are with GA Brim tagged `v0.22.0`.

On Windows (repro was on Server 2019) we seem pretty well covered. As shown in the attached video, when the Packets button is hit, a familiar "Windows can't open this type of file (.pcap)" message pops up.

https://user-images.githubusercontent.com/5934157/105641843-0252e300-5e3b-11eb-86c6-59884300384b.mov

On macOS (repro was on Big Sur) and Linux (repro was on Debian 10) it's not great. As shown in the videos below, the user sees the "Download Complete" message, but no app appears (expected) and no message is shown to explain why (that's bad!) I've kept the Dev Tools console window open during these repros to show that there doesn't seem to be an underlying message tied to this.

https://user-images.githubusercontent.com/5934157/105641898-4c3bc900-5e3b-11eb-83ac-d1e449dfad6a.mov

https://user-images.githubusercontent.com/5934157/105641903-5067e680-5e3b-11eb-8366-4bc22790b0ea.mov

If we can recognize this occurrence and provide guidance, that would be a great improvement.

After the community user's experience, they also offered the opinion:

> I think it might be helpful if the "download complete" message showed the directory that the files get saved into. I found them in my /tmp directory by accident.

Since these are supposed to be just temporary flow extractions, I'm not sure if this is something we'd want to add since it might give the user the impression that they're responsible for hunting down the files in that location, but for the "happy path" they should simply open in the appropriate app without issue. But this is good feedback, so I'm including it here for consideration when revisiting this UX.

philrz:
## tl;dr

Per #1379, we've had problems on some platforms when a user clicks **Download Packets** and there's no application installed to open the extracted slice, e.g., Wireshark. Though we've merged some changes over time that have helped us improve in this area, this change covers us for Linux where the problem is still significant.

Fixes #1379

## Details

As of the most recent update in https://github.com/brimdata/zui/issues/1379#issuecomment-2187902520, Zui actually gets decent help here from the OS on both macOS and Windows since in those cases the OS pops up a notification about the lack of a helper app for pcaps. However, on Linux the user would still see just the **Packets extracted. Opening...** message from Zui, no launched pcap app, and no other guidance, so there was still room for improvement there.

I pursued an angle advocated in https://github.com/brimdata/zui/issues/1379#issuecomment-769232791 of researching what's returned by the function call to open the pcap helper app, which is currently Electron's [`shell.openPath()`](https://www.electronjs.org/docs/latest/api/shell#shellopenpathpath). What I found is that it _does_ sometimes return an error message in the absence of a helper app, though it's not the most helpful. Specifically:

|**OS**|**Message**|
|-|-|
|macOS|"Failed to open path"|
|Linux|"The action failed"|
|Windows|(no message returned)|

Because these messages are pretty vague, in the branch for this PR I'm instead opting to display the more explicit error message **Could not open extracted pcap. Is Wireshark installed?** Granted, in theory a user could install a different pcap helper app instead of Wireshark, but it seems like 99% of our users think of Wireshark here, and if a user is advanced enough to be seeking alternate helper app I imagine they'll get the essence of what the message is implying and install the app of their choosing.

The fact there's no message returned at all on Windows is annoying, but since we have the OS pop-up to cover us there, I'm not inclined to get obsessed.

## Tests

Using a [Dev build](https://github.com/brimdata/zui/actions/runs/9656295221) generated from this branch, here's video on each OS showing evidence of the failure case (no Wireshark installed) and success case (just to confirm we've not disturbed what was already working ok).

### Linux

**Failure**

This is the main benefit we're getting here, since the user now sees our error message instead of the silence from before.

https://github.com/brimdata/zui/assets/5934157/6a136d4d-6b24-4c38-bb6a-1bce88121607

**Success**

https://github.com/brimdata/zui/assets/5934157/4dbc97dd-b54e-4d59-a931-95f4c5cffe3d

### Windows

**Failure**

Because `shell.openPath()` returns no error message here, we rely on the OS to inform the user.

https://github.com/brimdata/zui/assets/5934157/ec6eb498-ea86-453b-bc9b-7e94c8438ce8

**Success**

https://github.com/brimdata/zui/assets/5934157/5af61f13-a985-4c1c-b7f0-6fe43442b231

### macOS

**Failure**

Here we get the "belt & suspenders" benefit of both the error message from Zui and the pop-up from the OS.

https://github.com/brimdata/zui/assets/5934157/26b00c36-d2e3-49a4-8ecc-4f9be271295a

**Success**

https://github.com/brimdata/zui/assets/5934157/82ee7b38-903f-48f5-a1d0-cef7410baba0

### DIFF
--- a/apps/zui/src/plugins/brimcap/packets/download.ts
+++ b/apps/zui/src/plugins/brimcap/packets/download.ts
@@ -43,8 +43,14 @@ export async function downloadPackets(root: string, pool: string, uid: string) {
       const msg = JSON.parse(err)?.error || `brimcap search failed: ${err}`
       window.showErrorMessage(msg)
     } else {
-      shell.openPath(dest)
-      window.showSuccessMessage("Packets extracted. Opening...")
+      const msg = await shell.openPath(dest)
+      if (msg) {
+        window.showErrorMessage(
+          "Could not open extracted pcap. Is Wireshark installed?"
+        )
+      } else {
+        window.showSuccessMessage("Packets extracted. Opening...")
+      }
     }
   }
 }


jameskerr (commented):
Makes good sense to me.

jameskerr (approved):

PR #3115, opened 2024-06-25
Surface Export Errors

### LINKED ISSUE #2751: Query runtime errors are not being surfaced
The problem will occur when searching for several Pools that are imported with a different type of csv, line
Because one line does not have a csv structure, the results output is equivalent to 0 KB

I will put the sample file

[imdb.txt](https://github.com/brimdata/zui/files/11294675/imdb.txt)
[imdb.csv](https://github.com/brimdata/zui/files/11294674/imdb.csv)

I think you should add line option to export

jameskerr:
A toast error will now appear if there was a problem with the export.

CleanShot 2024-06-25 at 16 44 12@2x

Fixes #2751

### DIFF
--- a/apps/zui/src/core/invoke.ts
+++ b/apps/zui/src/core/invoke.ts
@@ -4,5 +4,12 @@ export function invoke<K extends OperationName>(
   name: K,
   ...args: Parameters<Operations[K]>
 ): Promise<Awaited<ReturnType<Operations[K]>>> {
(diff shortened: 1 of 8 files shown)


philrz (approved):
LGTM! 👍

In addition to witnessing the fix with CSV data, I also revived my Parquet changes referenced in #2591 and saw it addressed that too. So once this merges I'll be putting up a PR to support a new Export format. 😉

PR #3118, opened 2024-06-26
Handle zed.Union types in format.ts

### LINKED ISSUE #3117: "TypeError: Cannot read properties of null (reading 'isUnset')" when opening Detail view
Repro is with GA Zui tagged `v1.8.0` and at commit 52b4b58.

A user first reported the issue in a [community Slack thread](https://brimdata.slack.com/archives/C010DR0HHMF/p1719413291888149). In their own words:

> Can't open any packet after the recent update, I keep getting this error!
> Same issue on both Windows and Linux deb
>
>  image
> 
> Will try that but everything works fine except when I double click a packet. I created a new pool then loaded the pcap file then query after that when I double click on a packet, I get that error

I seem to have reproduced the issue on macOS as well, though I'd word the repro a little differently than the user. The problem doesn't seem to be "packet" specific, but rather, I was able to trigger it by applying `fuse` and then opening a record in the **Detail** view. Here's a repro in GA Zui tagged `v1.8.0` using the attached [ifconfig.pcapng.gz](https://github.com/user-attachments/files/15993903/ifconfig.pcapng.gz) test data (after uncompressing).

https://github.com/brimdata/zui/assets/5934157/332ca0e5-2de7-4e31-a30d-ba2a9c0ce8c6

I then did a binary search and confirmed the problem started happening at commit 52b4b58, which is associated with the changes in #3103. This next video is a repro at commit 52b4b58 using the attached [results.zng.gz](https://github.com/user-attachments/files/15993929/results.zng.gz) test data, which is just the records generated by loading that `ifconfig.pcapng`.

https://github.com/brimdata/zui/assets/5934157/0e92053f-e07f-468b-8237-42b14be0fd8c

Since this shows a more detailed error message I can start digging in and figure out what needs to be fixed. However, a secondary question is why the stack dump shown to the user in the GA release is so opaque. In this case I was lucky to have been able to make a lucky guess as to how to trigger the problem (the many `null` values in the background of the user's screenshot made me suspicious that `fuse` was involved) but it would have been much more helpful if the stack trace could have referenced the original lines of code.

philrz:
The changes in #3103 made it possible to smuggle a `null` union value into the Detail output formatter that was not prepared to handle them. After walking the code with @jameskerr, his suggestion was to back out the change from #3103 and instead add code in the Detail output formatter to handle unions in a way that's tolerant of `null` values. The attached video shows the fix working as intended with this PR's branch at commit 7807deb.

https://github.com/brimdata/zui/assets/5934157/a3353339-552c-4036-aa7d-f54843624667

As it turns out, @jameskerr already had work underway to sunset this particular code in the Detail view in favor of what we do with Inspector, and he's already confirmed that new code is immune to this problem. However, since we took the time to fully understand what's wrong here, I figure it doesn't hurt to fix it so we're leaving the code in better shape as we lay it to rest, plus it'll let the #3117 community user work off a Zui Insiders release with the fix in the days leading up to the next GA Zui release.

Fixes #3117

### DIFF
--- a/apps/zui/src/components/format.ts
+++ b/apps/zui/src/components/format.ts
@@ -65,6 +65,9 @@ export function formatValue(
   if (data instanceof zed.Map) {
     return `|{...${data.value.size}}|`
   }
+  if (data instanceof zed.Union) {
+    return formatValue(data.value, config)
+  }
   return null
 }
 
--- a/apps/zui/src/views/detail-pane/Fields.tsx
+++ b/apps/zui/src/views/detail-pane/Fields.tsx
@@ -26,9 +26,7 @@ const DataPanel = React.memo<DTProps>(function DataTable({
         <Data key={index}>
           <Name>{field.path.join(" ‣ ")}</Name>
           <Value className={zedTypeClassName(field.data)}>
-            {field.data instanceof zed.Union
-              ? format(field.data.value as zed.Primitive)
-              : format(field.data as zed.Primitive)}
+            {format(field.data as zed.Primitive)}
           </Value>
         </Data>
       ))}


jameskerr (approved):
Very good.

PR #3119, opened 2024-07-01
Tiny Duration Fix

### LINKED ISSUE #3101: Tiny duration shown as "Not available" in correlations view
Repro is with Zui commit b4ed51d.

The attached test data  [conn.log.gz](https://github.com/user-attachments/files/15890341/conn.log.gz)  contains two Zeek `conn` records. As shown in the attached video, when the `duration` value is large, the "Duration" line printed below the correlations view shows an accurate approximation of the the value (specifically, `21 minutes 9 seconds` whereas the full-precision original value was `21m9.512465s`). However, for the other record where the original value is `1us`, the value shown below the correlations view is `Not available`, which is incorrect.

https://github.com/brimdata/zed/assets/5934157/6db713ab-4dfc-46f3-8d7d-81bd21320a4f

I poked through the code and I see the math that's triggering the problem, but I'm not certain yet of the best fix.

jameskerr:
Fixes #3101 

With this fix:

If the duration is less than 1 second, it will display "xxxmilliseconds"
If it is less than 1 millisecond, it will display "less than one millisecond"

CleanShot 2024-07-01 at 11 34 24@2x

### DIFF
--- a/apps/zui/src/ppl/detail/models/ZeekEvent.ts
+++ b/apps/zui/src/ppl/detail/models/ZeekEvent.ts
@@ -1,5 +1,6 @@
 import * as zed from "@brimdata/zed-js"
 import {SecurityEventInterface} from "./security-event"
+import {isNumber} from "lodash"
 
 export class ZeekEvent implements SecurityEventInterface {
   constructor(private r: zed.Record) {}
@@ -14,10 +15,10 @@ export class ZeekEvent implements SecurityEventInterface {
 
   getEndTime() {
     if (this.r.get("_path").toString() !== "conn") return null
-    const dur = this.r.get<zed.Duration>("duration").asSeconds()
-    if (!dur) return
+    const dur = this.r.get<zed.Duration>("duration").asMs()
+    if (!isNumber(dur)) return
     const ts = this.r.get<zed.Time>("ts").toDate()
-    return new Date(ts.getTime() + dur * 1000)
+    return new Date(ts.getTime() + dur)
   }
 
   getType() {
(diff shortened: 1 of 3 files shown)


apps/zui/src/ppl/detail/models/ZeekEvent.ts
```diff
@@ -14,10 +15,10 @@ export class ZeekEvent implements SecurityEventInterface {
 
   getEndTime() {
     if (this.r.get("_path").toString() !== "conn") return null
-    const dur = this.r.get<zed.Duration>("duration").asSeconds()
-    if (!dur) return
+    const dur = this.r.get<zed.Duration>("duration").asMs()
```
jameskerr, line 18: Use milliseconds here instead of seconds, to get the most precision we can.

apps/zui/src/ppl/detail/util/formatDur.ts
```diff
@@ -14,7 +14,11 @@ export default function formatDir(start, end) {
     parts.push(formatPart(obj[unit], unit))
   }
   if (allZero) {
-    parts.push(formatPart((end.getTime() - start.getTime()) / 1000, "seconds"))
+    if (end.getTime() === start.getTime()) {
```
jameskerr, line 17: Check if the two dates are the same to determine if the duration is less than 1ms.

philrz (approved):
The fix seems effective. I saw one of the tests was broken so I pushed a commit to fix it.

jameskerr:
Thanks Phil!

PR #3122, opened 2024-07-03
Parquet export

### LINKED ISSUE #2591: Add Parquet as an Export format

philrz:
I've been holding off on adding Parquet as an Export format in Zui until #2751 was fixed, since there's so many Parquet limitations that can prevent the export from succeeding (e.g., lack of support for multiple types, lack of duration support, poor union support, etc.) But now that #2751 has been fixed and users will see the errors when exports fail, here I'm adding support for Parquet as an Export format.

Due to those same Parquet limitations, I could not just extend the existing loop-through-all-formats approach in the existing Export e2e test because an attempted export of that same test data file would fail. I saw a few ways this could be addressed:

1. I could rework the existing test to use a different, Parquet-eligible test data file for export in all formats and change all the expected byte counts to match. I chose not to pursue this since it didn't seem appropriate to "dumb down" to the lowest common denominator.

2. I could add an additional test just for Parquet that imports its own Parquet-eligible test data file. I took this approach in the first commit on this branch. However, I had to duplicate some code from the existing test.

3. Same as the prior, but I could break out the common code to a helper function that could be called by both the prior loop-through-all-formats approach as well as the new test for just Parquet. I took this approach in the second commit on this branch.

Of course, there's surely other variations, so I'm open to other adjustments.

Closes #2591

### DIFF
--- a/apps/zui/src/domain/results/handlers/export.ts
+++ b/apps/zui/src/domain/results/handlers/export.ts
@@ -69,7 +69,7 @@ export const exportToClipboard = createHandler(
 )
 
 export const getExportQuery = createHandler((ctx, format: ResponseFormat) => {
-  const formatNeedsFuse = ["csv", "tsv", "arrows"]
+  const formatNeedsFuse = ["arrows", "csv", "parquet", "tsv"]
   const query = ctx.select(Results.getQuery(RESULTS_QUERY))
   const isTable = ctx.select(Layout.getEffectiveResultsView) == "TABLE"
   const hiddenColCount = ctx.select(Table.getHiddenColumnCount)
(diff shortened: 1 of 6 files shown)


jameskerr (approved):
This is great!

PR #3123, opened 2024-07-05
Split Correlations and Detail Panes

jameskerr:
This PR is preparing the grounds for swapping out the fields detail view with and inspector.

A summary of changes:

1. A new correlation pane arrived on the detail view.
2. The look of the header in the detail pane is slightly different.
3. The look of the toolbar in the "columns" pane is slightly different.
4. The "TCP connection diagram" will now wrap if the detail pane gets too small, rather than overflow.
5. The button to collapse the right pane has been moved into the right pane when it is open, and back outside when it is closed.
6. The detail window now only shows the fields from the record.

https://github.com/brimdata/zui/assets/3460638/a73b8868-ba01-48f6-b919-dd6238689b22

### DIFF
--- a/apps/zui/pages/detail.tsx
+++ b/apps/zui/pages/detail.tsx
@@ -8,7 +8,7 @@ import initialize from "src/js/initializers/initialize"
 import TabHistories from "src/js/state/TabHistories"
 import Tabs from "src/js/state/Tabs"
 import {getPersistedWindowState} from "src/js/state/stores/get-persistable"
-import {DetailsWindow} from "src/views/details-window"
+import DetailPane from "src/views/detail-pane/Pane"
 
 export default function DetailPage() {
   const [app, setApp] = useState(null)
@@ -34,7 +34,7 @@ export default function DetailPage() {
   return (
     <AppProvider store={app.store} api={app.api}>
       <AppWindowRouter>
-        <DetailsWindow />
+        <DetailPane />
         <Modals />
         <Tooltip />
       </AppWindowRouter>
--- a/apps/zui/pages/detail.tsx
+++ b/apps/zui/pages/detail.tsx
@@ -34,7 +34,9 @@ export default function DetailPage() {
   return (
     <AppProvider store={app.store} api={app.api}>
       <AppWindowRouter>
-        <DetailPane />
+        <div className="scroll-y">
+          <DetailPane />
+        </div>
         <Modals />
         <Tooltip />
       </AppWindowRouter>
--- a/apps/zui/src/app/routes/app-wrapper/app-grid.tsx
+++ b/apps/zui/src/app/routes/app-wrapper/app-grid.tsx
@@ -26,7 +26,7 @@ export function AppGrid({children}) {
   const width = sidebarIsOpen ? sidebarWidth : 0
   const width2 = secondarySidebarIsOpen ? secondarySidebarWidth : 0
   const rows = ["40px", "1fr"]
-  const columns = [`min(${width}px, 70vw)`, "1fr", `min(${width2}px, 30vw)`]
+  const columns = [`min(${width}px, 50vw)`, "1fr", `min(${width2}px, 50vw)`]
   const style = {
     gridTemplateAreas: areas,
     gridTemplateRows: rows.join(" "),
--- a/apps/zui/src/components/data.tsx
+++ b/apps/zui/src/components/data.tsx
@@ -8,7 +8,7 @@ export const Data = styled.dl`
   position: relative;
   margin: 0;
   cursor: default;
-  min-height: 26px;
+  line-height: 1.7;
 `
 
 export const Name = styled.dt`
--- a/apps/zui/src/components/section-tabs.tsx
+++ b/apps/zui/src/components/section-tabs.tsx
@@ -1,70 +1,7 @@
 import React, {useLayoutEffect, useRef, useState} from "react"
 import {MenuItem} from "src/core/menu"
-import styled from "styled-components"
 import {MoreItemsButton} from "./more-items-button"
 import {useResponsiveMenu} from "src/js/components/hooks/use-responsive-menu"
-
-const BG = styled.div`
-  display: flex;
-  min-width: 0;
-  align-items: center;
-  height: 100%;
-  position: relative;
-  overflow: hidden;
-`
-
-const Nav = styled.nav`
-  display: flex;
-  align-items: center;
-  min-width: 0;
-  flex: 1;
-
-  button {
-    background: none;
-    border: none;
-
-    display: flex;
-    align-items: center;
-
-    border-radius: 6px;
-    padding: 6px 6px;
-
-    text-transform: uppercase;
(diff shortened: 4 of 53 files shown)


philrz:
I'm currently testing this branch at commit 791c64d.

In the attached screenshot, I've got a fresh Zui window open at the defaults on my Macbook. As it shows, only the **History** and **Detail** tabs are visible in the right pane and so the **>>** needs to be clicked to see the other two. I respect that as the app reacts to the user's different resizing preferences there's going to be configurations where certain things need to wrap and controls might need to be semi-hidden out of sight and that's very much the spirit of this PR. But I'm slightly concerned that something like **Correlations** (that's been one of our more beloved features) is now semi-hidden in this default state. Do you think it would be worth shrinking fonts and/or increasing the default panel width to bring **Correlations** and maybe **Columns** back into view by default? I respect that if we can imagine continuing to pile on more tabs in that panel we might not be able to keep this up indefinitely, so if you've already given it a thought and are comfortable with just **History** and **Detail** being out in the open by default on the assumption they're the ones that would be most frequently referenced and that we need to break the seal on users noticing and making use of the **>>** stuff, I could come around on that.

image

philrz:
I'm currently testing this branch at commit 791c64d.

I can see that vertical scrolling is working as expected in:
1. **Detail** (both when I click the tab in the right panel and when I open up a separate window)
2. **History**
3. **Columns**

However, it's not currently working in **Correlations**. For instance, when I shrink the window small enough that they start wrapping like you showed in your video, and then I make the window vertically shorter, they can become out-of-view below the bottom of the window and no vertical scrollbar activates to let me get at them.

philrz:
I'm currently testing this branch at commit 791c64d.

I happened to try running the e2e tests and noticed it's failing on a couple.

```
$ yarn e2e -g right-click-menus.spec.ts

> nx run zui-player:test -g right-click-menus.spec.ts

Running 9 tests using 1 worker
  ✓  1 right-click-menus.spec.ts:12:3 › right-click-menus › Inspector node menu (719ms)
  ✓  2 right-click-menus.spec.ts:23:3 › right-click-menus › table cell menu (794ms)
  ✓  3 right-click-menus.spec.ts:31:3 › right-click-menus › sort asc (694ms)
  ✓  4 right-click-menus.spec.ts:39:3 › right-click-menus › sort desc (809ms)
  ✓  5 right-click-menus.spec.ts:48:3 › right-click-menus › hide table column (668ms)
  ✓  6 right-click-menus.spec.ts:56:3 › right-click-menus › who is (2.3s)
  ✓  7 right-click-menus.spec.ts:66:3 › right-click-menus › virus total (944ms)
  ✘  8 right-click-menus.spec.ts:74:3 › right-click-menus › inspector show details (30.0s)
  ✘  9 right-click-menus.spec.ts:84:3 › right-click-menus › table show details (30.0s)
  1) right-click-menus.spec.ts:74:3 › right-click-menus › inspector show details ───────────────────
    Test timeout of 30000ms exceeded.
    Error: locator.waitFor: Application exited
    Call log:
      - waiting for getByRole('heading', { name: 'Fields', exact: true }).first() to be visible
       at ../helpers/test-app.ts:199
      197 |
      198 |   async attached(role: Role | RegExp, name?: string) {
    > 199 |     return this.locate(role, name).waitFor();
          |                                    ^
      200 |   }
      201 |
      202 |   async detached(role: Role | RegExp, name?: string) {
        at TestApp.attached (/Users/phil/work/zui/packages/zui-player/helpers/test-app.ts:199:36)
        at /Users/phil/work/zui/packages/zui-player/tests/right-click-menus.spec.ts:80:15
  2) right-click-menus.spec.ts:84:3 › right-click-menus › table show details ───────────────────────
    Test timeout of 30000ms exceeded.
    Error: locator.waitFor: Application exited
    Call log:
      - waiting for getByRole('heading', { name: 'Fields', exact: true }).first() to be visible
       at ../helpers/test-app.ts:199
      197 |
      198 |   async attached(role: Role | RegExp, name?: string) {
    > 199 |     return this.locate(role, name).waitFor();
          |                                    ^
      200 |   }
      201 |
      202 |   async detached(role: Role | RegExp, name?: string) {
        at TestApp.attached (/Users/phil/work/zui/packages/zui-player/helpers/test-app.ts:199:36)
        at /Users/phil/work/zui/packages/zui-player/tests/right-click-menus.spec.ts:89:15
  2 failed
    right-click-menus.spec.ts:74:3 › right-click-menus › inspector show details ────────────────────
    right-click-menus.spec.ts:84:3 › right-click-menus › table show details ────────────────────────
  7 passed (1.3m)

 ——————————————————————————————————————————————————————————————————————————————

 >  NX   Ran target test for project zui-player (1m)
 
         With additional flags:
           --g=right-click-menus.spec.ts
 
    ✖    1/1 failed
    ✔    0/1 succeeded [0 read from cache]
```

I can tell it's surely an easy fix since it just waiting on the "Fields" header and it looks like it doesn't have a "Name" handle like it did in the past.

philrz:
I'm currently testing this branch at commit 0641f2b.

I can see what I've found here is not altogether new, but it became more obvious because the correlations seem to be rendered a bit bigger in this branch.

In the attached video taken at this PR's branch at commit 0641f2b, I load sample data [conn.log.gz](https://github.com/user-attachments/files/16135159/conn.log.gz) and then click to the new **Correlations** tab with the window and panels all at their default sizes. We can see that the correlations view happens to "wrap", which is kind of unfortunate for this state diagram because in order to interpret it the arrows need to line up between the entires on the left and right sides. My first instinct was to try to widen the right panel by clicking and dragging leftward, but there was only a tiny bit of "give". I then dragged it its right edge rightward, and _that_ did do the trick.

As I fiddle with it more after having figured that out, I find that the main results area acts a little "greedy" in this regard, i.e., even if I collapse the left panel and make the main results area super wide, I seem to be locked into a max width of the right side panel. That's not a crisis as long as it can be made wide enough to show its contents and would offer horizontal scrolling even if that became impossible in the available real estate, but I found myself struggling with it a bit for the first time ever.

https://github.com/brimdata/zui/assets/5934157/9994cf57-7408-447a-9f07-c7cfedc1ccfd

By comparison, here's how things look with the same repro steps on Zui Insiders `1.8.1-insiders.6`. In its default size, the correlation when still inside the **Detail** panel is too small to fit and so is chopped off on the right side, which is not great and I know we're happy to see addressed with these changes. However, it just so happens that the click and drag leftward _does_ happen to bring the whole state diagram into view in this case, so I got to my happy state a little quicker. The other "greedy" behaviors with the main area I noted above still appeared much the same then as they do in this PR's branch, though.

https://github.com/brimdata/zui/assets/5934157/2c495a72-58ce-4a23-957d-495026d263b9

In conclusion, I'm not suggesting some big overhaul of behaviors here, but wanted to point it out in case you see any small adjustments that might be worth making here, even if just something like rendering the correlation a little smaller like it was before so it wraps less often.

philrz (approved):

philrz (approved):

PR #3131, opened 2024-07-17
Update zed and brimcap to v1.17.0 and prep Zui v1.17.0

philrz:
https://github.com/brimdata/zui/compare/v1.8.0...256ecfd


jameskerr (approved):
Solid

PR #3133, opened 2024-07-29
Use scrolltop and height to determine if we should load more

### LINKED ISSUE #2742: Pulling scrollbar to bottom not triggering load of more data
Repro is with Zui commit 8444565.

I've noticed that often if I grab the vertical scrollbar control and pull it all the way to the bottom, it doesn't automatically trigger the loading of more data. In the attached video I got more data on the first "pull", but not on the subsequent ones. Counterintuitively, once in that state, if I scroll _downward_ with my mouse roller (can't easily be shown in the video, but I'm doing it) it doesn't trigger the loading of more data, which gives me the impression there's no more available, but then if I scroll _upward_ with the roller it loads more.

https://user-images.githubusercontent.com/5934157/228941352-3cf81388-7b4b-4c3f-8c86-1d4dc48b2416.mp4

See also the issue #2636. It seems we have a few things we could improve in terms of visual cues about how much data there is to view. The "Results: First ..." in the bottom of the window is doing some of this for us, but I find it to be fairly subtle.

jameskerr:
Fixes #2742

The previous version of the code was relying on the rendered visible index, which didn't get updated fast enough. So now the code relies on the scrollTop and the height to determine if we are near the bottom of the scrollable area.

### DIFF
--- a/apps/zui/src/views/results-pane/inspector.tsx
+++ b/apps/zui/src/views/results-pane/inspector.tsx
@@ -26,7 +26,7 @@ export function Inspector(props: {height?: number}) {
 
   function onScroll({top, left}) {
     dispatch(Slice.setScrollPosition({top, left}))
-    if (list.current?.nearBottom(30)) loadMore()
+    if (list.current?.nearBottom(30, top)) loadMore()
   }
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


philrz (approved):
Tests out nicely! 👍

PR #3142, opened 2024-08-21
Add Features doc for new Time settings

philrz:


apps/zui/docs/features/Packet-Captures.md
```diff
@@ -1,7 +1,3 @@
----
-sidebar_position: 2
```
philrz, line 2: I dropped these `sidebar_position` directives in all these Features articles since that way they're automatically sorted alphabetically.

apps/zui/docs/features/Time-Display.md
```diff
@@ -0,0 +1,35 @@
+# Time Display
+
+By default, Zui displays Zed `time` values in the Table, Inspector,
+and Detail views using [ZSON format](https://zed.brimdata.io/docs/formats/zson#23-primitive-values).
+This is an [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)
+date/time string in nanosecond precision ending in `Z` to indicate
+[UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), e.g.,
+`2024-08-14T19:12:51.123456789Z`.
+
+Starting with Zui release v1.18.0, two different options in Zui **Settings**
```
philrz, line 10: I put in the forward-looking reference to the next Zui release so that way I can merge this right away and ping people on community Slack and/or the person that opened https://github.com/brimdata/brimcap/issues/352 to get early feedback.

apps/zui/src/electron/run-main/run-configurations.ts
```diff
@@ -50,7 +50,7 @@ export function runConfigurations() {
         placeholder: "%Y-%m-%dT%H:%M:%S.%L%:z",
         helpLink: {
           label: "docs",
-          url: "https://github.com/samsonjs/strftime?tab=readme-ov-file#supported-specifiers",
+          url: "https://zui.brimdata.io/docs/features/Time-Display",
```
philrz, line 53: Chicken/egg. This link will be broken until the PR merges. C'est la vie!

apps/zui/docs/features/Time-Display.md
```diff
@@ -0,0 +1,35 @@
...
+   represented using [strftime directives](https://github.com/samsonjs/strftime?tab=readme-ov-file#supported-specifiers)
+   as `%Y-%m-%dT%H:%M:%S.%L%:z`. Note that this format is acceptable as a
+   literal `time` value in a Zed program, e.g., assuming data containing a field
+   `ts` of the `time` type, the [search expression](https://zed.brimdata.io/docs/language/search-expressions)
+   `ts > 2024-08-14T12:12:51.123-07:00` is syntactically valid.
+
+2. If the **Time Format** setting is changed from its empty default, any
+   [strftime directives](https://github.com/samsonjs/strftime?tab=readme-ov-file#supported-specifiers)
+   in the setting are used to format time values. For instance the setting
+   `%m/%d/%Y %Z` would produce the displayed value `08/14/2024 Pacific Daylight Time`.
```
jameskerr, line 28: I don't think we should use the capital %Z in this example since it's broken in the strftime library.

philrz: @jameskerr: Yeah, you'd mentioned offline how you perceived `%Z` to be broken and I'd not spotted a problem before putting up this PR. And trying it again now it seems to be working as I'd expect.

For instance, I start by importing this file `t.zson`:

```
{ts:2024-08-14T19:12:51.123456789Z}
```

Here's how it looks with the default ZSON formatting:

image

Here's how it changes if I change just the **Time Zone** to "US/Pacific" and let the the default `strftime` formatting show the modified base time with the offset. So, hour `19` changes to hour `12`.

image

Now I go back into **Settings** and also change **Time Format** to `%Y-%m-%dT%H:%M:%S.%L %Z`. The base time still shows hour `12` so that seems correct, and the written out time zone communicates what the numeric offset had been doing before.

image

So, that all seems correct to me, but maybe there's something else I'm missing. 😄  If it'd be easier to hash this out interactively on a Zoom just ping me.

jameskerr (approved):
Maybe that one example can be updated, but then looks good.

apps/zui/docs/features/Time-Display.md
philrz: @jameskerr and I spoke offline and I understand what's broken now. Here's some detail for the record.

I wasn't triggering the problem in my last comment above because I picked "US/Pacific" which happens to also be my local time zone. However, if I pick a different time zone like "US/Eastern", we see:

image

That shows the hour has been correctly adjusted to `15` but the time zone shown is still my local one. In other words, the library behaves as if `%Z` is to "tell me about the time zone I'm in" as opposed to "tell me about the time zone I asked to format in". This seems like a bug. However, I'm guessing that most users that reach for these settings will probably be seeking to display in their local time zone anyway, so they may not even notice.

For now I'll rework the example to avoid showing `%Z` to avoid drawing unnecessary attention.

PR #3143, opened 2024-08-28
New Tab Bar

### LINKED ISSUE #3140: Empty tab slot left behind after deleting Session
Repro is with Zui commit d3a0402. This was first spotted [when testing the #3132 branch](https://github.com/brimdata/zui/pull/3132#issuecomment-2276743784) but we decided to merge what was there and fix this after.

Repro steps shown in the attached video:

1. Load some data
2. Click to the Sessions list in the left panel
3. Close the pool page for the data just added
4. Start one Session and run a `count()` on the pool
5. Start a second session and run a search (e.g., `google`) on the pool
6. Right-click and delete the Session with `count()` from the list

An empty slot is left behind for the deleted tab. This empty slot is filled in as soon as I click another tab.

https://github.com/user-attachments/assets/1f68a8fb-93fa-4f26-a461-2675fbc15a06

jameskerr:
Things to address:

- [x] Hover state remains after dropping
- [x] If user drags to far to the left or right keep the spots open.
- [x] Clicking a tab needs to activate it.
- [x] If you close a tab that has no queries in it, just delete that session.
- [x] Check if it's a preview tab and make the text italic

Fixes #3140


philrz:
I'm currently testing this branch at commit 123a16a.

It seems the app has lots its ability to initiate a data load by dragging a file icon into the window.

https://github.com/user-attachments/assets/4bfec16b-5d85-481a-a5ae-74bbad4f351a

philrz:
I'm currently testing this branch at commit 123a16a.

This is just cosmetic stuff regarding the text in the tabs. It seems things have been made a bit "tighter" such that there's less blank space in the tab to the right of the text like "Welcome to Zui" or "Query Session". This seems mostly fine, but two things did stand out to me as shown in the attached video:

1. When starting from the initial app default of the "Welcome to Zui" being the only open tab, as soon as I open my first Query Session tab, the "i" in "Zui" gets clipped out of view.
2. When tab text does get clipped out of view (which is inevitable as enough tabs are opened, of course), we no longer show "..." in the right side of the text to indicate something's out of view.

https://github.com/user-attachments/assets/ec3b33fd-dd11-4076-a587-eccfd28a06fc

Here's a baseline for comparison with Zui Insiders `1.17.1-insiders.20`. Here we need to open four Query Session tabs before any tab text starts being clipped out of view, and the "..." appear when that happens.

https://github.com/user-attachments/assets/ee0d9bba-4708-4b8d-98d1-9bcb1c8b5c46

Some of this surely comes down to personal preference, of course. I'm not wedded to "..." so if it was dropped intentionally I'm fine with that. I do think the way the "i" in "Zui" was clipped so easily is probably worth addressing, though, just because it's a "new user" experience thing.

jameskerr:
Thank Phil for your review. I've just pushed some commits that address both the dropping of data, and the clipping of the text when there are a small number of tabs. 

I did drop the ellipsis because it ended up taking more space to show the dots. Without the dots, you can see more of the text, even though it abruptly ends.

philrz (approved):

PR #3145, opened 2024-09-03
Don't Wrap Values in Records

jameskerr:
Steve pointed out the confusion that comes when the app wraps a non-record in a record to display it in a table format. Now we just display in the inspector view. 

The meaning of the "table view" button is now "table view preferred when possible". 

**Before**

CleanShot 2024-09-03 at 12 22 51@2x

**After**

CleanShot 2024-09-03 at 12 22 38@2x

### DIFF
--- a/apps/zui/src/js/flows/viewLogDetail.ts
+++ b/apps/zui/src/js/flows/viewLogDetail.ts
@@ -8,7 +8,7 @@ import Current from "../state/Current"
 import {invoke} from "src/core/invoke"
 
 export const viewLogDetail =
-  (record: zed.Record): Thunk =>
+  (record: zed.Value): Thunk =>
   (dispatch, getState) => {
     const current = LogDetails.build(getState())
     if (record && !isEqual(record, current)) {
--- a/apps/zui/src/js/state/LogDetails/actions.ts
+++ b/apps/zui/src/js/state/LogDetails/actions.ts
@@ -9,7 +9,7 @@ import {
 } from "./types"
 
 export default {
-  push: (record: zed.Record): LOG_DETAIL_PUSH => ({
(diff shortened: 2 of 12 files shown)


philrz:
I'm currently testing this branch at commit 8a1294d.

As shown in the attached video, I can trigger a stack dump if I load the attached [primitive.zson.gz](https://github.com/user-attachments/files/16854262/primitive.zson.gz), click the option to `fuse` it, and then click one of the values in an attempt to see it rendered in the **Detail** panel.

https://github.com/user-attachments/assets/ab03f28f-1c19-4abd-87da-b7d4dbcc3712

The dump text:

```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'rootRecord')

Source
src/views/results-pane/inspector.tsx (64:34) @ rootRecord

  62 | valueProps={{
  63 |   onClick: (e, value, field) => {
> 64 |     const rootValue = field.rootRecord
     |                            ^
  65 |     dispatch(Selection.set({value, field, rootValue}))
  66 |     if (field && field instanceof zed.Field) {
  67 |       dispatch(viewLogDetail(field.rootRecord))
```

philrz:
Here's a bonus thought about what I showed in the last comment, and maybe @mccanne would have some input here.

The app's guidance "...or `fuse` results to view as a table" is arguably incorrect since with a sequence of primitive values like this `fuse` _can't_ turn it into a table, and by applying `fuse` each value will instead just take on a union type. The way the branch is working now, clicking that `fuse` link just gives the unlabeled sequence of values.

image

If the app had picked up on the presence of any primitive values in the query output, perhaps it could have changed the guidance to more accurately reflect that it's impossible to make into a table and say as much, and perhaps drop the `fuse` link in that case?

philrz (approved):

PR #3146, opened 2024-09-03
Make preview and load draggable

jameskerr:
Previously, it was not possible to drag zui when the preview and load modal was open. This fixes that.

Fix Demo

https://github.com/user-attachments/assets/ae8d0381-ab05-4052-84cf-b3c2d325670a


philrz (approved):
Tests out nicely for me! 👍

PR #3181, opened 2025-03-07
Convert to ESM

jameskerr:
I've made the zui-player repo a "ESModule" instead of a "commonjs" module. So we need to use this module syntax when importing files.


philrz (approved):
Thanks @jameskerr! I tested this had the intended effect by witnessing `yarn e2e:ci` running successfully whereas previously it failed. I also snuck in an extra commit to the GitHub Actions workflow that runs `yarn e2e:ci` when the super dependency changes.

PR #3183, opened 2025-03-11
Adjust packets test and fix standalone e2e Actions Workflow

philrz:
Since reviving the backend dependency updating via the changes in #3180, the e2e test portion of the Actions Workflow now known as `advance-super.yml` has been consistently failing in CI though it runs ok locally. The specific failure such as observed in [this recent job](https://github.com/brimdata/zui/actions/runs/13796048677/job/38587841092) reflects a timeout in the `packets.spec.ts` test.

I tried using the separate Actions Workflow `e2e.yml` that runs the same test to study the problem and found it needed some updates to run at all, so I've made those necessary changes here. Once I got it to run, I used Playwright's video capture to get a closer look at it failing. [This job](https://github.com/brimdata/zui/actions/runs/13795660794) resulted in the following screenshot that shows the failure message appearing in the **Preview** panel with the headline **Shaper Error**.

image

By comparison, when I watch the same test pass locally or run the same steps locally by hand, the error is presented like this with the headline **Error Reading Data**:

image

Since the test was waiting on the **Error Reading Data** message, that explains why it was passing locally. I asked @kerr if he had a guess why it showed up differently in CI, and it sounds like it may come down to timing. In his own words:

>I think I run the original query and the shaper query at the same time. It's probably the shaper finishing first in the CI case?

This test was recently morphed into this much simpler "confirm pcap doesn't work" test since we recently dropped pcap support in the app, so it's probably not worth obsessively fiddling with the app code or the test to make this more deterministic. Instead I'm just morphing the test here to instead look for the "wrong number of fields" message since that appears in both contexts. I've confirmed this now runs successfully in [this job](https://github.com/brimdata/zui/actions/runs/13798460167), so once this merges the "advance super" Workflow should similarly work.

### DIFF
--- a/packages/app-player/tests/packets.spec.ts
+++ b/packages/app-player/tests/packets.spec.ts
@@ -1,11 +1,9 @@
 import { play } from 'app-player';
 import { getPath } from '@brimdata/sample-data';
 
-// Timeouts are increased due to observed long pcap load times in CI.
-// See https://github.com/brimdata/zui/pull/2978
 play('packets.spec', (app, test) => {
   test('dropping a pcap throw an error now', async () => {
     await app.dropFile(getPath('sample.pcap'));
-    await app.attached(/Error Reading Data/);
+    await app.attached(/wrong number of fields/);
   });
 });
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -56,9 +56,6 @@ jobs:
         id: playwright
         run: |
           if [ "$RUNNER_OS" = "Linux" ]; then
-            # Mock for pcap slice opening to avoid hanging Actions Runner.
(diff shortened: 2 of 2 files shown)


packages/app-player/tests/packets.spec.ts
```diff
@@ -1,11 +1,9 @@
 import { play } from 'app-player';
 import { getPath } from '@brimdata/sample-data';
 
-// Timeouts are increased due to observed long pcap load times in CI.
-// See https://github.com/brimdata/zui/pull/2978
```
philrz, line 5: This comment is related to timeout increases that were indeed added back in #2978 but the timeouts were dropped when the test was recently simplified, so the comment should go too.

jameskerr (approved):

Issue #3059, opened 2024-04-26
CI hang during "yarn --inline-builds"


philrz:
In the time after #2972 merged, I've started to notice a repeating intermittent failure in CI during the "Setup Zui" Actions Workflow in the `yarn --inline-builds` step. For example, here's how it looked in the [most recent failure](https://github.com/brimdata/zui/actions/runs/8842643153) which was during a "Create Insiders Release" run on macOS:

```
Link step
  ➤ YN0007: │ @swc/core@npm:1.3.41 must be built because it never has been before or the last one failed
  ➤ YN0007: │ esbuild@npm:0.17.12 must be built because it never has been before or the last one failed
  ➤ YN0007: │ esbuild@npm:0.18.14 must be built because it never has been before or the last one failed
  ➤ YN0007: │ @parcel/watcher@npm:2.0.4 must be built because it never has been before or the last one failed
  ➤ YN0007: │ brimcap@https://github.com/brimdata/brimcap.git#commit=bf7fb4996738767bb4f27eee939ec67dd21aab52 must be built because it never has been before or the last one failed
  ➤ YN0007: │ electron@npm:28.0.0 must be built because it never has been before or the last one failed
  ➤ YN0007: │ keytar@npm:7.7.0 must be built because it never has been before or the last one failed
  ➤ YN0007: │ msw@npm:0.36.8 must be built because it never has been before or the last one failed
  ➤ YN0007: │ styled-components@npm:5.3.5 [cb7c7] must be built because it never has been before or the last one failed
  ➤ YN0007: │ playwright-chromium@npm:1.41.1 must be built because it never has been before or the last one failed
  ➤ YN0000: │ playwright-chromium@npm:1.41.1 STDOUT Downloading Chromium 121.0.6[167](https://github.com/brimdata/zui/actions/runs/8842643153/job/24281616674#step:3:177).57 (playwright build v1097) from https://playwright.azureedge.net/builds/chromium/1097/chromium-mac.zip
  ➤ YN0000: │ playwright-chromium@npm:1.41.1 STDOUT |                                                                                |   0% of 139.7 MiB
  ➤ YN0000: │ playwright-chromium@npm:1.41.1 STDOUT |■■■■■■■■                                                                        |  10% of 139.7 MiB
  ➤ YN0000: │ playwright-chromium@npm:1.41.1 STDOUT |■■■■■■■■■■■■■■■■                                                                |  20% of 139.7 MiB
[...]

PR #3186, opened 2025-03-25
Add debug info and disable shaper run

jameskerr:


philrz:
As I recall, this branch had been an attempt to further debug the intermittent failure of the e2e "packets" test. In brief, after re-enabling backend dependencies for the SuperDB era, at first that test was failing consistently, then the changes in #3183 made it able to pass but it still failed intermittently so the test was disabled entirely in #3184. The theory was that the error message was sometimes not surfacing due to a timing problem similar to what's studied in #3059. With the debug changes in this branch we were trying to better isolate the root cause of #3059 but sadly never figured it out.

PR #3190, opened 2025-04-07
Advance Node client's super dependency automatically (regular dep)

philrz:
## What's Changing

This change enables the automatic advancing of the Node client's `super` dependency via Actions workflow, just like we do for the app's `super` dependency.

## Why

https://github.com/brimdata/zui/pull/3187 illustrated how we'll sometimes need to advance the Node client's dependency, and so I ended up doing a one-shot advance of the dependency in that PR with intent to follow up and make it automatic in a separate PR, and that's what this PR is.

## Details

When I set out to make the change, I had some struggles. At first I tried to just repeat the same `yarn workspace ... add ...` approach that's already used for the app's dependency, but this caused some errors related to how superdb-node-client had `super` listed as both a `devDependency` and a `peerDependency`. The relevant docs online didn't 100% make sense to me so after some false starts in #3188 and #3189 I flagged down @jameskerr who advised:

>  I think it's more preferable to have superdb-node-client be it's own self-contained thing. So it depends on super directly. As a normal dependency. Not peer, not dev.
>
> In a project like superdb-desktop which depends on both, yarn will see that they depend on the same commit and it will only download that one for both.

That's the approach I'm using in these changes here. Both `yarn test` and `yarn e2e` pass locally.

### DIFF
--- a/.github/workflows/advance-super.yml
+++ b/.github/workflows/advance-super.yml
@@ -38,6 +38,7 @@ jobs:
       - uses: ./.github/actions/setup-app
       - run: sudo apt-get -y install whois
       - run: yarn workspace superdb-desktop add super@brimdata/super#${{ env.super_ref }}
+      - run: yarn workspace superdb-node-client add super@brimdata/super#${{ env.super_ref }}
       - run: yarn lint
       - run: yarn test
       - run: yarn build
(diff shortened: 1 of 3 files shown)


jameskerr (approved):
Looks solid to me!

philrz:
Thanks very much @jameskerr!

nwt (approved):

Issue #3218, opened 2025-12-26
drillDown test failures correlated with advance to super commit a12d23c


philrz:
The app has had its super dependency stuck at commit `b0dc059` because once it advances to `a12d23c` (associated with the changes in https://github.com/brimdata/super/pull/6453) or newer, `yarn test` has the following failures:

```
 FAIL  src/js/models/program.test.ts (6.03 s)
  ● drill down › when there is no leading filter

    "i d" no such field at line 1, column 12:
    count() by this["i d"]["orig h"]
               ~~~~

      61 |   async function run(value: any, text: string) {
      62 |     const info = await fetchQueryInfo(text, "test")
    > 63 |     if (info.error) throw new Error(info.error.error)
         |                           ^
      64 |     const hasAggs = !!info.channels[0].aggregations_keys
      65 |     return drillDown(text, value, hasAggs, info.channels[0].aggregation_keys)
      66 |   }

      at run (src/js/models/program.test.ts:63:27)
      at Object.<anonymous> (src/js/models/program.test.ts:69:20)

  ● drill down › when there is a sort on there

    "name" no such field at line 1, column 1:
    name=="james" | count() by proto | sort -r count
    ~~~~
    "proto" no such field at line 1, column 28:
    name=="james" | count() by proto | sort -r count
                               ~~~~~

      61 |   async function run(value: any, text: string) {
      62 |     const info = await fetchQueryInfo(text, "test")
    > 63 |     if (info.error) throw new Error(info.error.error)
         |                           ^
      at run (src/js/models/program.test.ts:63:27)
      at Object.<anonymous> (src/js/models/program.test.ts:94:20)

  ● drill down › easy peasy

    "name" no such field at line 1, column 1:
    name=="james" | count() by proto
    ~~~~
    "proto" no such field at line 1, column 28:
    name=="james" | count() by proto
                               ~~~~~

      61 |   async function run(value: any, text: string) {
      62 |     const info = await fetchQueryInfo(text, "test")
    > 63 |     if (info.error) throw new Error(info.error.error)
         |                           ^
      64 |     const hasAggs = !!info.channels[0].aggregations_keys
      65 |     return drillDown(text, value, hasAggs, info.channels[0].aggregation_keys)
      66 |   }

      at run (src/js/models/program.test.ts:63:27)
      at Object.<anonymous> (src/js/models/program.test.ts:105:20)

  ● drill down › count by and filter the same

    "md5" no such field at line 1, column 1:
[...]

PR #3219, opened 2025-12-26
Use an explicit source in program.test.ts

philrz:
In contemplating #3218, @nwt remarked:

> This ultimately traces back to the change we made to the HTTP API to ignore the pool name zui sends along side a query. That pool was treated as a default source if the query didn't include an explicit source. Without it, you get the null source. And with type checking against the null source instead of a pool, you get errors.

Therefore in this branch I've added explicit leading `from test` to the queries, which gets the tests to pass again with the newer SuperDB dependency.

### DIFF
--- a/apps/superdb-desktop/package.json
+++ b/apps/superdb-desktop/package.json
@@ -142,7 +142,7 @@
     "set-tz": "^0.2.0",
     "sprintf-js": "^1.1.2",
     "styled-components": "^5.3.5",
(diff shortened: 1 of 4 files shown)


philrz:
@nwt has confirmed that my proposed change here is indeed just papering over an underlying problem. The test failure is believed to be an accureate reflection of features in the app (probably related to pins) that no longer work due to an API change, so at some point someone will do the work to fix the app. In the meantime, closing this PR without merging.

